PR #12, opened 2017-05-22
Fix #11

### LINKED ISSUE #11: OpenCL Headers Not Found
Hi, the build process for tests would complain that OpenCL headers could not be found.As long as the OpenCL SDK is not installed on regular system path.
It turns out that the FindComputeCpp.cmake doesn't add `OpenCL_INCLUDE_DIR` variable into include directories. 
I'll create a pull request later.

mshockwave:
Add OpenCL_INCLUDE_DIR into include path in FindComputeCpp.cmake


Ruyk:
Good catch! Will merge in

Issue #14, opened 2017-05-30
Fixing Arg/Min/Max


mehdi-goli:
It is possible to remove code duplication by separationg component tasks into different individual tasks.
```
// ReducAssignNewOp3 can be eliminated completely by introducing (In blas1_tree.hpp)
//STEP1:Adding Tuple OP node
//.....
template<typename RHS>
struct TupleOP {
using value_type = IndVal<size_t, typename RHS::value_type>;
RHS r;
UnaryOp(RHS& _r) : r(_r){};
size_t getSize() { return r.getSize(); }
inline value_type eval(size_t i) { return value_type(i, r.eval(i); }
inline value_type eval(cl::sycl::nd_item<1> ndItem) {
return eval(ndItem.get_global(0));
}/....
//STEP2// By adding that it is possible to use the ReducAssignNewOp2 to do the imax, imean.
// Therefore for constructing the _imax, we wrap the input node with <TupleOP<INPUTTYPE>(input_node);
// This is in blas1_interface_sycl.hpp
///....
auto vec =std::vector<IndVal<T> >(nWG)
auto buf_val=cl::sycl::buffer<IndVal<T>, 1>(vec.data(), cl::sycl::range{nWG});
auto val1 = BufferVectorView<IndVal<T>>>(valT1, 0, 1, nWG);
auto val_tuple = TupleOP<vector_view<IndVal<T>, std::vector<IndVal<T>>>>(my_vx);
auto assignOp1 =
make_maxIndReducAssignNewOp2(val1, val_tuple, localSize, nWG * localSize);
ex.execute(assignOp1);
std::vector<IndVal<T>> valT2(1);
auto val2 = vector_view<IndVal<T>, std::vector<IndVal<T>>>(valT1, 0, 1, 1);
auto assignOp2 = make_maxIndReducAssignNewOp2(val2, val1, localSize, nWG);
//.....
// next step is to modify the blas_operators for Imax and Imean
//...
SYCLBLAS_DEFINE_TERNARY_OPERATOR(maxIndOp3_struct, const_val::min, > )
SYCLBLAS_DEFINE_TERNARY_OPERATOR(minIndOp3_struct, const_val::max, < )
// Should change to
SYCLBLAS_DEFINE_BINARY_OPERATOR(maxIndOp3_struct, const_val::min,
((l.getVal() > r.getVal()) ? l : r))
[...]

PR #15, opened 2017-05-31
fixing imin/imax

mehdi-goli:
#14

### DIFF
--- a/include/executors/blas1_tree_executor.hpp
+++ b/include/executors/blas1_tree_executor.hpp
@@ -183,19 +183,19 @@ struct Evaluate<BinaryOp<Operator, LHS, RHS>> {
   }
 };
 
-/*! Evaluate<ReducAssignNewOp2<Operator, LHS, RHS>>
+/*! Evaluate<AssignReduction<Operator, LHS, RHS>>
  * @brief See Evaluate.
  */
 template <typename Operator, typename LHS, typename RHS>
-struct Evaluate<ReducAssignNewOp2<Operator, LHS, RHS>> {
+struct Evaluate<AssignReduction<Operator, LHS, RHS>> {
   using value_type = typename LHS::value_type;
   using oper_type = Operator;
   using LHS_type = LHS;
   using cont_type = typename LHS::ContainerT;
   using lhs_type = typename Evaluate<LHS>::type;
   using rhs_type = typename Evaluate<RHS>::type;
-  using input_type = ReducAssignNewOp2<Operator, LHS, RHS>;
-  using type = ReducAssignNewOp2<Operator, lhs_type, rhs_type>;
+  using input_type = AssignReduction<Operator, LHS, RHS>;
+  using type = AssignReduction<Operator, lhs_type, rhs_type>;
 
   static type convert_to(input_type v, cl::sycl::handler &h) {
     auto lhs = Evaluate<LHS>::convert_to(v.l, h);
--- a/include/executors/blas1_tree_executor.hpp
+++ b/include/executors/blas1_tree_executor.hpp
@@ -30,11 +30,11 @@
 
 #include <CL/sycl.hpp>
 
-#include <views/view_sycl.hpp>
 #include <executors/executor_base.hpp>
 #include <operations/blas1_trees.hpp>
 #include <operations/blas2_trees.hpp>
 #include <operations/blas3_trees.hpp>
+#include <views/view_sycl.hpp>
 
 namespace blas {
 
@@ -55,7 +55,7 @@ struct Evaluate {
   /** convert_to.
    * @brief .
    */
-  static type convert_to(input_type v, cl::sycl::handler& h) { return v; }
+  static type convert_to(input_type v, cl::sycl::handler &h) { return v; }
 };
 
 /*! Evaluate <Join<LHS, RHS>>.
@@ -69,7 +69,7 @@ struct Evaluate<Join<LHS, RHS>> {
   using input_type = Join<LHS, RHS>;
   using type = Join<lhs_type, rhs_type>;
 
-  static type convert_to(input_type v, cl::sycl::handler& h) {
+  static type convert_to(input_type v, cl::sycl::handler &h) {
     auto lhs = Evaluate<LHS>::convert_to(v.l, h);
     auto rhs = Evaluate<RHS>::convert_to(v.r, h);
     return type(lhs, rhs);
@@ -87,7 +87,7 @@ struct Evaluate<Assign<LHS, RHS>> {
   using input_type = Assign<LHS, RHS>;
   using type = Assign<lhs_type, rhs_type>;
 
-  static type convert_to(input_type v, cl::sycl::handler& h) {
+  static type convert_to(input_type v, cl::sycl::handler &h) {
     auto lhs = Evaluate<LHS>::convert_to(v.l, h);
     auto rhs = Evaluate<RHS>::convert_to(v.r, h);
     return type(lhs, rhs);
@@ -106,7 +106,7 @@ struct Evaluate<DobleAssign<LHS1, LHS2, RHS1, RHS2>> {
   using input_type = DobleAssign<LHS1, LHS2, RHS1, RHS2>;
   using type = DobleAssign<lhs1_type, lhs2_type, rhs1_type, rhs2_type>;
 
-  static type convert_to(input_type v, cl::sycl::handler& h) {
+  static type convert_to(input_type v, cl::sycl::handler &h) {
     auto lhs1 = Evaluate<LHS1>::convert_to(v.l1, h);
     auto lhs2 = Evaluate<LHS2>::convert_to(v.l2, h);
     auto rhs1 = Evaluate<RHS1>::convert_to(v.r1, h);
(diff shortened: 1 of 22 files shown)


include/executors/blas1_tree_executor.hpp
```diff
@@ -126,13 +126,29 @@ struct Evaluate<ScalarOp<Operator, SCL, RHS>> {
...
+  using rhs_type = typename Evaluate<RHS>::type;
+  using input_type = TupleOp<RHS>;
+  using type = TupleOp<rhs_type>;
+
+  static type convert_to(input_type v, cl::sycl::handler &h) {
+    auto rhs = Evaluate<RHS>::convert_to(v.r, h);
+    return type(rhs);
+  }
+};
+
```
Ruyk: Why is there a generic Evaluate on the blas1 interface? cannot this be part of the generic tree operations so it is easily re-used?

include/interface/blas1_interface_sycl.hpp
```diff
@@ -338,67 +338,108 @@ void _asum(Executor<ExecutorType> ex, int _N, vector_view<T, ContainerT> _vx,
...
-  auto val1 = vector_view<IndVal<T>, std::vector<IndVal<T>>>(valT1, 0, 1, nWG);
+  size_t localSize = 256, nWG = 512;
+  auto tupOp = TupleOp<vector_view<T, ContainerT>>(my_vx);
+  std::vector<IndVal<T>> valT1(nWG,
+                               IndVal<T>(std::numeric_limits<size_t>::max(),
+                                         std::numeric_limits<T>::min()));
+  cl::sycl::buffer<IndVal<T>, 1> bvalT1(valT1.data(), cl::sycl::range<1>{nWG});
+  BufferVectorView<IndVal<T>> val1(bvalT1, 0, 1, nWG);
+  /* vector_view<IndVal<T>,BufferVectorView<IndVal<T>>> val1(valT1, 0, 1, nWG);
+   */
```
Ruyk: why commented code?

include/interface/blas1_interface_sycl.hpp
```diff
@@ -338,67 +338,108 @@ void _asum(Executor<ExecutorType> ex, int _N, vector_view<T, ContainerT> _vx,
...
-  auto assignOp2 = make_maxIndReducAssignNewOp2(val2, val1, localSize, nWG);
-  ex.execute(assignOp2);
+      make_maxIndReducAssignNewOp2(val1, tupOp, localSize, localSize * nWG);
+  ex.reduce(assignOp1);
+  auto assignOp2 = make_maxIndReducAssignNewOp2(my_rs, val1, localSize, nWG);
+  ex.reduce(assignOp2);
 #ifdef VERBOSE
   std::cout << "ind = " << val1.eval(0).getInd() << std::endl;
 #endif  //  VERBOSE
-  return val1.eval(0).getInd();
```
Ruyk: There is not return value in this function, is that expected?

include/operations/blas1_trees.hpp
```diff
@@ -248,6 +248,25 @@ struct BinaryOp {
...
+
+  size_t getSize() { return r.getSize(); }
+
+  value_type eval(size_t i) { return value_type(i, r.eval(i)); }
+
+  value_type eval(cl::sycl::nd_item<1> ndItem) {
+    return eval(ndItem.get_global(0));
+  }
+};
+
```
Ruyk: Why is a generic TupleOp on the specific blas1_tree header? Should not it be on the generic tree for all blas levels?

include/operations/blas1_trees.hpp
```diff
@@ -275,9 +294,9 @@ struct ReducAssignNewOp2 {
       value_type local_val = Operator::init(r);
       for (size_t k = j; k < vecS; k += 2 * grdS) {
         local_val = Operator::eval(local_val, r.eval(k));
-        local_val = ((k + blqS) < vecS)
-                        ? Operator::eval(local_val, r.eval(k + blqS))
-                        : local_val;
+        if (k + blqS < vecS) {
+          local_val = Operator::eval(local_val, r.eval(k + blqS));
+        }
```
Ruyk: This is now simpler to vectorize, thanks

include/operations/blas2_trees.hpp
```diff
@@ -318,7 +319,7 @@ struct RedRowMatVct {
     }
     return valWG;
   }
-#endif // ORIGINAL_CODE
+#endif  // ORIGINAL_CODE
```
Ruyk: Do we still need this if/def for the original code?

include/operations/blas3_trees.hpp
```diff
@@ -26,12 +26,12 @@
 #ifndef BLAS3_TREES_HPP
 #define BLAS3_TREES_HPP
 
-#include <iostream>
 #include <stdexcept>
 #include <vector>
+#include <iostream>
```
Ruyk: We shouldn't include iostream on the tree files, is it that we have some debug cout?

include/operations/blas_constants.hpp
```diff
@@ -39,12 +39,35 @@
 #define SYCLBLAS_DEFINE_CONSTANT(type, indicator, val) \
   template <>                                          \
   struct constant<type, indicator> {                   \
-    static constexpr type value = val;                 \
-  };
+    static const type value;                           \
+  };                                                   \
+  const type constant<type, indicator>::value =        \
+      val;  // temporary work around for duplicator issue will be fixed int the
+            // next release
```
Ruyk: "Temporary work around for a compiler issue on ComputeCpp 0.2.0, bug reported internally"

include/operations/blas_constants.hpp
```diff
@@ -39,12 +39,35 @@
...
+struct IndVal {
+  using value_type = ScalarT;
+  size_t ind;
+  value_type val;
+
+  // NEEDED FOR VECTOR CONSTRUCTION
+  /* IndVal(): */
+  /*   ind(std::numeric_limits<size_t>::max()), */
+  /*   val(0) */
+  /* {}; */
```
Ruyk: why is this commented?

include/operations/blas_operators.hpp
```diff
@@ -71,17 +55,17 @@ struct IndVal {
...
-    }                                                          \
-                                                               \
-    template <typename R>                                      \
-    static typename R::value_type init(const R& r) {           \
-      return constant<typename R::value_type, initial>::value; \
-    }                                                          \
+#define SYCLBLAS_DEFINE_BINARY_OPERATOR(name, initial, expr)          \
+  struct name {                                                       \
+    template <typename L, typename R>                                 \
+    static typename strip_asp<R>::type eval(const L &l, const R &r) { \
```
Ruyk: is this a workaround for a Comp[uteCpp issue?

include/operations/blas3_trees.hpp
theoden8: It was there

theoden8:
https://github.com/codeplaysoftware/sycl-blas/blob/master/.clang-format
The clang-format file I used.

include/interface/blas1_interface_sycl.hpp
theoden8: So, there is a version of _iamax which returns nothing, but writes the index-value pair into the _rs parameter, and there is another function which uses it and returns size_t.

Ruyk: Makes sense, thanks.

include/operations/blas3_trees.hpp
Ruyk: Can you remove it?

Ruyk:
Seems some of the previous commits were formatted without that file, so your formatting is correct.

include/operations/blas3_trees.hpp
theoden8: Yes in this case, not in case of the interface files though.

include/operations/blas2_trees.hpp
theoden8: This is for blas2, and we have not really gone into that yet. I'd better leave as it is and deal with it in another pull request.

Ruyk: Makes sense.

include/operations/blas1_trees.hpp
theoden8: At least for the time being, it is only being used for amax/amin, so it belongs to blas1_tree header.

include/operations/blas_operators.hpp
theoden8: No, it is a workaround for a compiler issue.

theoden8: Yes

include/executors/blas1_tree_executor.hpp
theoden8: Evaluate is in all three files, and Evaluate<TupleOP> is an evaluator specific to blas1 interface.

include/operations/blas_operators.hpp
Ruyk: Can you move it to a separate new header file where we can put all the ComputeCpp-specific macros?

include/operations/blas1_trees.hpp
Ruyk: It is easier if you move it now to a generic header so that you always have it available in all the other places where you will use it on the future

include/operations/blas_operators.hpp
theoden8: I think @mehdi-goli has almost fixed that and this workaround will disappear shortly, so it makes sense not to touch it, i think.

Ruyk: Ok.

PR #19, opened 2017-07-12
Current work at restructuring the interface

theoden8:
Please, do not accept this pull request, this is solely to keep track of the progress.

(1) Split Operation and Evaluate trees into Expression, Evaluate<expr> and Convert<eval>.
(2) Added a concept of the device with blas_packet_traits and blas_packet_traits_sycl.
(3) Added functor_traits to change device-specific blas_operations; merged with `address_space` and `static_cast` removal.
(4) Tried to change the executor_sycl to use `execute`, `reduce` with expressions, but was blocked by (a) vector_view/matrix_view being an expression, had to make an evaluator (b) reduce which reduces evaluator, needs to reduce an expression like in the interface (c) localsize/nWG which are specified in the interface differently depending on the context, but need to be a part of expression execution.
(5) Adding a workaround with `bool eval_subexpr_if_needed(cont_type *data)` to execute a subtree when a node runs in multiple kernels as an overhead of the main fused kernel. As a workaround for 4(b) `reduce` moved queue into device instantiation and using a device instance as a parameter to both `execute` and `reduce` so `evaluator<reduction>::eval_subexpr_if_needed` does not need to know what is an `Executor<sycl>` and we can declare `reduce<evaluatorT, device>` function in advance.
(6) Changing the interface accordingly and trying to compile the whole thing.

Improvements suggestions are warmly welcome.


include/evaluators/blas2_tree_evaluator.hpp
```diff
@@ -0,0 +1,482 @@
...
+  Evaluator<RHS1, Device> r1;
+  Evaluator<RHS2, Device> r2;
+
+  size_t nThr = 0;
+
+  Evaluator(Expression &expr)
+      : l(Evaluator<LHS, Device>(expr.l)),
+        r1(Evaluator<RHS1, Device>(expr.r1)),
+        r2(Evaluator<RHS2, Device>(expr.r2)) {
+    nThr = 2;
```
Ruyk: use the constructor list for this initialization
```
, nThr(2) { 
```

include/evaluators/blas2_tree_evaluator.hpp
```diff
@@ -0,0 +1,482 @@
...
+  Evaluator<RHS1, Device> r1;
+  Evaluator<RHS2, Device> r2;
+
+  Evaluator(Expression &expr)
+      : scl(expr.scl),
+        l(Evaluator<LHS, Device>(expr.l)),
+        r1(Evaluator<RHS1, Device>(expr.r1)),
+        r2(Evaluator<RHS2, Device>(expr.r2)) {}
+
+  size_t getSize() { return r1.getSizeR(); }
```
Ruyk: const

include/evaluators/blas2_tree_evaluator.hpp
```diff
@@ -0,0 +1,482 @@
...
+      // The result is stored in lhs
+      if ((rowid < dimR) && (colid == 0)) {
+        l.eval(rowid, blqidC) = scl * scratch[localid] + r3.eval(rowid);
+      }
+      rowid += rowSz;
+    }
+#endif  // BLAS_EXPERIMENTAL
+    return val;
+  }
+#endif  // BLAS_EXPERIMENTAL
```
Ruyk: what is this macro for?

include/evaluators/blas2_tree_evaluator.hpp
```diff
@@ -0,0 +1,482 @@
...
+  Evaluator<RHS1, Device> r1;
+  Evaluator<RHS2, Device> r2;
+  Evaluator<RHS3, Device> r3;
+
+  Evaluator(Expression &expr)
+      : r1(Evaluator<RHS1, Device>(expr.r1)),
+        r2(Evaluator<RHS2, Device>(expr.r2)),
+        r3(Evaluator<RHS3, Device>(expr.r3)) {}
+
+  size_t getSize() { return r1.getSizeR(); }
```
Ruyk: getSize can be const

include/evaluators/blas3_tree_evaluator.hpp
```diff
@@ -0,0 +1,89 @@
...
+  static constexpr bool supported = RHS1::supported && RHS2::supported;
+
+  Evaluator<RHS1, Device> r1;
+  Evaluator<RHS2, Device> r2;
+
+  Evaluator(Expression &expr)
+      : r1(Evaluator<RHS1, Device>(expr.r1)),
+        r2(Evaluator<RHS2, Device>(expr.r2)) {}
+
+  size_t getSize() { return r1.getSize(); }
```
Ruyk: getSize can be const method, apply it everywhere

include/evaluators/blas_functor_traits.hpp
```diff
@@ -0,0 +1,144 @@
...
+ *  limitations under the License.
+ *
+ *  SYCL-BLAS: BLAS implementation using SYCL
+ *
+ *  @filename blas_functor_traits.hpp
+ *
+ **************************************************************************/
+
+#ifndef BLAS_FUNCTOR_TRAITS_HPP_7FBGTUQ8
+#define BLAS_FUNCTOR_TRAITS_HPP_7FBGTUQ8
```
Ruyk: why the 7FBGTUQ8 ?

include/evaluators/blas_functor_traits.hpp
```diff
@@ -0,0 +1,144 @@
...
+struct functor_traits<absOp1_struct, ScalarT, Device> : absOp1_struct {
+  static constexpr bool supported = Packet_traits<ScalarT, Device>::has_abs;
+  template <typename R = ScalarT>
+  static R eval(R r) {
+    return cl::sycl::abs(r);
+  }
+};
+template <typename ScalarT, typename Device>
+struct functor_traits<addOp2_struct, ScalarT, Device> : addOp2_struct {
+  static constexpr bool supported = Packet_traits<ScalarT, Device>::has_add;
```
Ruyk: Packet_traits is a new thing on this merge request too? Can it be done on a separate one? 
Also Packet_traits is an inconsistent naming scheme, use either packet_traits or PacketTraits (I believe the first one is better)

include/evaluators/blas_tree_evaluator_base.hpp
```diff
@@ -0,0 +1,48 @@
+#ifndef BLAS_TREE_EVALUATOR_BASE_HPP_L8QC4HLH
+#define BLAS_TREE_EVALUATOR_BASE_HPP_L8QC4HLH
```
Ruyk: why L8QC4HLH ?

Ruyk: The license header is missing in this file

include/evaluators/blas_tree_evaluator_base.hpp
```diff
@@ -0,0 +1,48 @@
...
+ * When the value is not an integral basic type,
+ * it is assumed to be a vector and the first value
+ * is used.
+ */
+template <typename T,
+          bool isscal = std::is_same<T, int>::value ||
+                        std::is_same<T, float>::value ||
+                        std::is_same<T, double>::value ||
+                        std::is_same<T, std::complex<float>>::value ||
+                        std::is_same<T, std::complex<double>>::value>
```
Ruyk: why don't simply use is_arithmetic?
http://en.cppreference.com/w/cpp/types/is_arithmetic

include/executors/blas_packet_traits.hpp
```diff
@@ -0,0 +1,65 @@
...
+ *  limitations under the License.
+ *
+ *  SYCL-BLAS: BLAS implementation using SYCL
+ *
+ *  @filename blas_packet_traits.hpp
+ *
+ **************************************************************************/
+
+#ifndef BLAS_PACKET_TRAITS_HPP_9LPTVQ25
+#define BLAS_PACKET_TRAITS_HPP_9LPTVQ25
```
Ruyk: why the 9LPTVQ25

include/executors/blas_packet_traits.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+template <typename T, typename Device>
+struct Packet_traits : default_packet_traits<T, Device> {};
+
+template <typename T, typename Device>
+using packet_type = typename Packet_traits<T, Device>::packet_type;
+
+}  // blas
+
+#endif
```
Ruyk: Do we need this as part of this merge request?

include/executors/blas_packet_traits_sycl.hpp
```diff
@@ -0,0 +1,124 @@
...
+   * @brief SYCL queue for execution of trees.
+   */
+  cl::sycl::device m_device;
+  /*!
+   * @brief SYCL device.
+   */
+  cl::sycl::queue m_queue;
+public:
+  SYCLDevice():
+    m_queue(cl::sycl::queue([&](cl::sycl::exception_list l) {
```
Ruyk: move the lambda out so the code is clear

include/executors/blas_packet_traits_sycl.hpp
```diff
@@ -0,0 +1,124 @@
...
+        try {
+          if(e) {
+            std::rethrow_exception(e);
+          }
+        } catch(cl::sycl::exception e) {
+          std::cerr << e.what() << std::endl;
+        }
+      }
+    })),
+    m_device(m_queue.get_device())
```
Ruyk: m_device should be initialized first

include/executors/blas_packet_traits_sycl.hpp
```diff
@@ -0,0 +1,124 @@
...
+  {}
+
+  static void parallel_for_setup(size_t &localSize, size_t &nWG, size_t globalSize = UINT_MAX) {
+    localSize = 256;
+    nWG = 512;
+    if(globalSize != UINT_MAX) {
+      nWG = (nWG + (2 * localSize) - 1) / (2 * localSize);
+    }
+  }
```
Ruyk: why is this a static method? what is the purpose of this?

include/executors/blas_packet_traits_sycl.hpp
```diff
@@ -0,0 +1,124 @@
...
+
+  cl::sycl::queue &sycl_queue() {
```
Ruyk: if this is a get method, it should be named get_sycl_queue()
Note also that this is returning a reference to an internal class object, what is the reason for it? SYCL objects are reference counted so there is no need for this

include/executors/reduction.hpp
```diff
@@ -0,0 +1,132 @@
+#ifndef REDUCTION_HPP
```
Ruyk: missing license header

include/executors/reduction.hpp
```diff
@@ -0,0 +1,132 @@
...
+#define REDUCTION_HPP
```
Ruyk: name should be SYCLBLAS_REDUCTION_HPP

include/executors/reduction.hpp
```diff
@@ -0,0 +1,132 @@
...
+/*               ((nWG == 1) ? lhs : (even ? opShMem2 : opShMem1)), */
+/*               (even ? opShMem1 : opShMem2), localSize, globalSize); */
+/*       execute_subtree<Device, using_shared_mem::enabled>(dev.sycl_queue(), localExpr, localSize, globalSize, sharedSize); */
+/*     } */
+/*     _N = nWG; */
+/*     nWG = (_N + (2 * localSize) - 1) / (2 * localSize); */
+/*     frst = false; */
+/*     even = !even; */
+/*   } while (_N > 1); */
+/* } */
```
Ruyk: why is this code commented? (is ok if it is just 'cos its in progress)

include/operations/blas1_trees.hpp
```diff
@@ -23,8 +23,8 @@
  *
  **************************************************************************/
 
-#ifndef BLAS1_TREES_HPP
-#define BLAS1_TREES_HPP
+#ifndef BLAS1_TREE_EXPR_HPP_TEJQWMOC
+#define BLAS1_TREE_EXPR_HPP_TEJQWMOC
```
Ruyk: why the TEJQWMOC

include/operations/blas1_trees.hpp
```diff
@@ -33,402 +33,111 @@
...
-@tparam operatorT Operator type of the operation node.
-@tparam operandT0 Operand type of the first by value operand of the operation
-node.
-@tparam operandsTN Operand types of the subsequent reference operands of the
-operation node.
-@param operand0 First by value operand of the operation node.
-@param operands Subsequent reference operands of the oepration node.
-@return Constructed operation node.
+@brief Template function for constructing expression nodes based on input
+tempalte and function arguments. Specialised case for an expression, a single by
```
Ruyk: typo, tempalte -> template

include/operations/blas2_trees.hpp
```diff
@@ -23,13 +23,12 @@
  *
  **************************************************************************/
 
-#ifndef BLAS2_TREES_HPP
-#define BLAS2_TREES_HPP
+#ifndef BLAS2_TREE_EXPR_HPP_VE71NQCB
+#define BLAS2_TREE_EXPR_HPP_VE71NQCB
```
Ruyk: why the VE71NQCB

include/operations/blas3_trees.hpp
```diff
@@ -23,13 +23,13 @@
  *
  **************************************************************************/
 
-#ifndef BLAS3_TREES_HPP
-#define BLAS3_TREES_HPP
+#ifndef BLAS3_TREE_EXPR_HPP_X30IEJA7
+#define BLAS3_TREE_EXPR_HPP_X30IEJA7
```
Ruyk: why the X30IEJA7

include/operations/blas_operators.hpp
```diff
@@ -55,78 +55,31 @@ namespace blas {
...
-#endif  // __SYCL_DEVICE_ONLY__  && __COMPUTECPP__
-
-/**
- * syclblas_abs.
- *
- * SYCL 1.2 defines different functions for abs for floating point
- * and integer numbers, following the OpenCL convention.
- * To choose the appropriate one we use this template specialization
- * that is enabled for floating point to use fabs, and abs for everything else.
+/** wang random generator
```
Ruyk: attribute proper references using links or explain where the code is coming from

include/views/view_sycl.hpp
```diff
@@ -59,8 +60,9 @@ struct get_size_struct<bufferT<ScalarT>> {
  */
 template <typename ScalarT>
 struct vector_view<ScalarT, bufferT<ScalarT>> {
+  static constexpr bool supported = Packet_traits<ScalarT, SYCLDevice>::Supported;
   using ContainerT = bufferT<ScalarT>;
-  ContainerT &data_;
+  ContainerT data_;
```
Ruyk: This is a fundamental change of the design, is it intentional? @mehdi-goli  ?

include/views/view_sycl.hpp
```diff
@@ -241,6 +239,26 @@ struct vector_view<ScalarT, bufferT<ScalarT>> {
...
+      throw std::out_of_range("invalid packet start index");
+    packet_type<ScalarT, SYCLDevice> retVec;
+    /* { */
+    /*   auto hostPtr = data_.template
+     * get_acceess<cl::sycl::access::mode::read_write,
+     * cl::sycl::access::target::host_buffer>(); */
+    /*   retVec.load(ind, hostPtr); */
+    /* } */
+    return retVec;
+  }
```
Ruyk: evalPacket seems that can be done on a separate merge request

lukeiwanski:
Can you please split this to smaller PR's?

include/evaluators/blas1_tree_evaluator.hpp
```diff
@@ -0,0 +1,152 @@
...
+ *  limitations under the License.
+ *
+ *  SYCL-BLAS: BLAS implementation using SYCL
+ *
+ *  @filename blas1_tree_evaluator.hpp
+ *
+ **************************************************************************/
+
+#ifndef BLAS1_TREE_EVALUATOR_HPP_CDFFQDGR
+#define BLAS1_TREE_EVALUATOR_HPP_CDFFQDGR
```
lukeiwanski: How about `PATH_TO_FILE_FROM_PROJECT_ROOT_FILE_NAME_FILE_EXTENSION` here?

include/evaluators/blas1_tree_evaluator.hpp
```diff
@@ -0,0 +1,152 @@
...
+ *  with y a scalar and x a subexpression tree.
+ */
+template <typename Functor, class LHS, class RHS>
+struct Evaluator<ReductionExpr<Functor, LHS, RHS>, SYCLDevice> {
+  using Expression = ReductionExpr<Functor, LHS, RHS>;
+  using Device = SYCLDevice;
+  using value_type = typename Expression::value_type;
+  using dev_functor = functor_traits<Functor, value_type, Device>;
+  using cont_type = typename Evaluator<LHS, Device>::cont_type;
+  auto &self = *this;
```
theoden8: I need value_type for `eval`, dev_functor to call device specific/general functor depending on the traits, cont_type for eval_subexpr_if_needed and self is removed, it is just there temporarily.

include/evaluators/blas1_tree_evaluator.hpp
theoden8: So, the random thing is generated by my IDE, I will remove it. The reason to use it was because I was copying operations/ files into evaluators/, and the header guards would be the same. We can stick to the convention if we agree on one, I don't mind.

include/evaluators/blas1_tree_evaluator.hpp
```diff
@@ -0,0 +1,152 @@
...
+
+  Evaluator(Expression &expr):
+    l(Evaluator<LHS, Device>(expr.l)),
+    r(Evaluator<RHS, Device>(expr.r))
+  {
+    setBlocksize(256);
+    setGridsize(512);
+  }
+
+  size_t getSize() { return r.getSize(); }
```
theoden8: I'm not putting consts for two reasons: I would have to copy-paste a lot of consts and it would be harder to compile.

include/evaluators/blas2_tree_evaluator.hpp
```diff
@@ -0,0 +1,482 @@
...
+ *  limitations under the License.
+ *
+ *  SYCL-BLAS: BLAS implementation using SYCL
+ *
+ *  @filename blas2_tree_evaluator.hpp
+ *
+ **************************************************************************/
+
+#ifndef BLAS2_TREE_EVALUATOR_HPP_YDJ0AOH7
+#define BLAS2_TREE_EVALUATOR_HPP_YDJ0AOH7
```
theoden8: Same as above.

include/evaluators/blas2_tree_evaluator.hpp
```diff
@@ -0,0 +1,482 @@
...
+
+  Evaluator<RHS1, Device> r1;
+  Evaluator<RHS2, Device> r2;
+
+  size_t mult = 0;
+
+  Evaluator(Expression &expr)
+      : r1(Evaluator<RHS1, Device>(expr.r1)),
+        r2(Evaluator<RHS2, Device>(expr.r2)) {
+    mult = 0;  // why is it unused
```
theoden8: Okay, removed that

include/evaluators/blas2_tree_evaluator.hpp
theoden8: Did it, however, I am not sure what it is doing and whether it can or can not be set by the device.

include/evaluators/blas2_tree_evaluator.hpp
theoden8: I am not sure, it was there before. I will try to figure out if this code is redundant,

include/evaluators/blas_tree_evaluator_base.hpp
theoden8: Because we want to specify explicitly which types are supported by this blas library. I just shortened DetectScalar which was instantiating the same class N times for each primitive type.

include/executors/blas_packet_traits.hpp
theoden8: Yep, I will remove it.

include/executors/blas_packet_traits_sycl.hpp
theoden8: Don't actually need it since it is in the queue anyway, right?

include/executors/blas_packet_traits_sycl.hpp
theoden8: It is just a skeleton code for calculating nWG localsize globalsize etc depending on the device.

include/executors/blas_packet_traits_sycl.hpp
theoden8: Will discuss it with @mehdi-goli

include/executors/reduction.hpp
theoden8: added

include/executors/reduction.hpp
theoden8: renamed to reduction_sycl.hpp with header guard BLAS_REDUCTION_SYCL_HPP, to match other files.

include/executors/reduction.hpp
theoden8: We discussed it with @mehdi-goli and have not found any place this function is used.

include/views/view_sycl.hpp
theoden8: true, will remove

include/executors/blas_packet_traits_sycl.hpp
theoden8: I am not sure if it is a safer way to do it.

theoden8:
@lukeiwanski I will try to split it into smaller PRs once I reach a breakpoint, at the moment it is not completely clear how many more changes are needed to get through the errors, so I am submitting work-in-progress.

include/executors/blas_packet_traits_sycl.hpp
Ruyk: Returning a copy of the object is safer in this case, since its reference counted. This means the only thing you copy is a thin object that has only a pointer to the actual implementation. In ComputeCpp is implemented as http://en.cppreference.com/w/cpp/language/pimpl

include/executors/blas_packet_traits_sycl.hpp
Ruyk: yes, you don't need it

include/evaluators/blas_tree_evaluator_base.hpp
Ruyk: Ok, but then that should be on a separate include file together with other traits that are specific to the library, so the user can also use the same trait to check if its possible to use the type with syclblas

include/evaluators/blas2_tree_evaluator.hpp
Ruyk: I am not sure neither, maybe it was a temporary implementation of an algorithm? @josealiaga ?

include/evaluators/blas1_tree_evaluator.hpp
Ruyk: Those are not valid reasons to reject the suggesion. I don't understand what it means to be "harder to compile". There is no significant difference in compilation time using const. The compiler will not get more tired by adding more consts. And if you don't want to copy/paste you can always write const. 
An alternative is to have a base class with the basic properties of the nodes so you can derive getSize and other friends in all cases. That will save you the copy/paste, but its up to @mehdi-goli I guess if he wants to go through that?

include/evaluators/blas1_tree_evaluator.hpp
Ruyk: Just use *this directly or pass this as a pointer if you really have to. If you find yourself passing this to a function, you probably need to make it a member of the class.

include/evaluators/blas1_tree_evaluator.hpp
Ruyk: maybe you can add simply _COPY at the end so it is clear that its a temporary copy? In any case its a WIP branch so I am not particularly concerned.

theoden8: _COPY is not appropriate because it is a permanent copy.

include/evaluators/blas2_tree_evaluator.hpp
theoden8: Can't we use a batched reduction here?

josealiaga: There is something wrong because I can see the code ...  we will comment it on monday

josealiaga: There is something wrong because I can't see the code ...  we will see on monday

theoden8:
The last commit fixes blas1 operations which do not use reduction.

theoden8:
Changes:
(1) Added different reductions.
(2) Removed shared memory from executors: it's confusing, hard to debug, annoying and useless because only reduction needs scratch.
(3) Removed executor completely: there is no use of an empty class with no methods or fields, Using device instead, it's much nicer.

The main changes were made, sequentially, to blas1_tree_evaluator, reduction_sycl, executor_sycl and blas1_interface.

include/evaluators/blas2_tree_evaluator.hpp
```diff
@@ -0,0 +1,483 @@
...
+
+    size_t rowid = blqidR * rowSz + localid;  // rowid of the current thread
+    size_t colid = blqidC * colSz;  // first column of the current thread
+
+    size_t k;
+
+    // Copying  to the scratch
+    k = localid;
+    for (size_t j = colid + localid; j < colid + colSz; j += rowSz) {
+      if ((rowid < dimR) && (j < dimC)) scratch[k] = r2.eval(j);
```
Ruyk: use braces always, never one-liners

include/evaluators/blas2_tree_evaluator.hpp
```diff
@@ -0,0 +1,483 @@
...
+
+  value_type eval(cl::sycl::nd_item<1> ndItem) {
+    return eval(ndItem.get_global(0));
+  }
+};
+
+/*! RedRowMatVct.
+ * @brief CLASSICAL AXPY GEMV
+ */
+// #define ORIGINAL_CODE 1
```
Ruyk: can this commented define be removed?

include/evaluators/blas1_tree_evaluator.hpp
```diff
@@ -41,100 +41,42 @@ namespace blas {
...
-      value_type local_val = dev_functor::init(r);
-      for (size_t k = j; k < vecS; k += 2 * grdS) {
-        local_val = dev_functor::eval(local_val, r.eval(k));
-        if (k + blqS < vecS) {
-          local_val = dev_functor::eval(local_val, r.eval(k + blqS));
-        }
-      }
-      // Reduction inside the block
-      val = dev_functor::eval(val, local_val);
+    if (cont)){
```
Ruyk: seems weird parenthesis here?

include/evaluators/blas1_tree_evaluator.hpp
```diff
@@ -41,100 +41,42 @@ namespace blas {
...
-    }
-    if (localid == 0) {
-      l.eval(groupid) = scratch[localid];
-    }
-    return l.eval(groupid);
+    FullReducer<Self, Device>::run(this, dev, *result);
+    return true;
   }
 
+  value_type eval(size_t i) { return m_result[i]; }
```
Ruyk: if this is returning by value, the method can be const

include/evaluators/blas2_tree_evaluator.hpp
```diff
@@ -109,8 +109,7 @@ struct Evaluator<PrdRowMatVctMultExpr<LHS, RHS1, RHS2, RHS3>, Device> {
         r1(Evaluator<RHS1, Device>(expr.r1)),
         r2(Evaluator<RHS2, Device>(expr.r2)),
         r3(Evaluator<RHS3, Device>(expr.r3)),
-        nThr(2)
-  {}
+        nThr(2) {}
```
Ruyk: why 2? a comment is needed explaining why 2, possibly need to ask the original author

theoden8: Done that, putting 2 for now.

include/evaluators/blas2_tree_evaluator.hpp
theoden8: Yes, it's not used anywhere.

include/evaluators/blas1_tree_evaluator.hpp
theoden8: ops

include/evaluators/blas1_tree_evaluator.hpp
theoden8: The assignment is not fixed yet, so it makes sense to figure out whether it is value or reference when it starts to matter.

include/evaluators/blas2_tree_evaluator.hpp
theoden8: Ok

include/evaluators/blas1_tree_evaluator.hpp
Ruyk: I mean:
```
value_type eval(size_t) const { return m_result[i]; }
```

theoden8:
Only one error left:

```
In file included from /home/kirill/sycl-blas-fusion/tests/blas1_interface_test.cpp:32:
In file included from /home/kirill/sycl-blas-fusion/include/interface/blas1_interface_sycl.hpp:37:
/home/kirill/sycl-blas-fusion/include/executors/reduction_sycl.hpp:283:8: error: redefinition of 'FullReducer'
struct FullReducer {
       ^
/home/kirill/sycl-blas-fusion/include/evaluators/blas1_tree_evaluator.hpp:44:8: note: previous definition is here
struct FullReducer {
       ^
In file included from /home/kirill/sycl-blas-fusion/tests/blas1_interface_test.cpp:32:
In file included from /home/kirill/sycl-blas-fusion/include/interface/blas1_interface_sycl.hpp:37:
/home/kirill/sycl-blas-fusion/include/executors/reduction_sycl.hpp:292:37: error: unknown type name 'Device'
  static void run(EvaluatorT &expr, Device &dev) {
                                    ^
2 errors generated.
tests/CMakeFiles/blas1_interface_test_blas1_interface_test.cpp_0_integration_header.dir/build.make:61: recipe for target 'tests/blas1_interface_test.cpp.sycl' failed
make[2]: *** [tests/blas1_interface_test.cpp.sycl] Error 1
CMakeFiles/Makefile2:281: recipe for target 'tests/CMakeFiles/blas1_interface_test_blas1_interface_test.cpp_0_integration_header.dir/all' failed
make[1]: *** [tests/CMakeFiles/blas1_interface_test_blas1_interface_test.cpp_0_integration_header.dir/all] Error 2
Makefile:94: recipe for target 'all' failed
make: *** [all] Error 2
```

So, there is only one error left essentially for blas1 without iamax/iamin.

This is not finished, but very close to it. I will need to fix this error and unlock blas1/blas2/blas3.

theoden8:
simple_axpy_result is wrong, damn it.

include/evaluators/blas_functor_traits.hpp
```diff
@@ -0,0 +1,153 @@
...
+
+template <class Functor, typename ScalarT, typename Device>
+struct functor_traits : Functor {
+  /* static constexpr bool supported = false; */
+};
+
+template <typename ScalarT, typename Device>
+struct functor_traits<iniAddOp1_struct, ScalarT, Device> : iniAddOp1_struct {
+  /* static constexpr bool supported = true; */
+};
```
josealiaga: Should these functors be initialised somewhere? 
For instance. the initial value of the addition is zero

theoden8: It is inherited from iniAddOp1_struct by design

include/evaluators/blas1_tree_evaluator.hpp
```diff
@@ -0,0 +1,86 @@
...
+
+  size_t getSize() const { return r.getSize(); }
+  cont_type data() { return l.data(); }
+
+  bool eval_subexpr_if_needed(cont_type *cont, Device &dev) {
+    r.eval_subexpr_if_needed(NULL, dev);
+    FullReducer<Self, GenericReducerTwoStage<Self>>::run(*this, dev);
+    return true;
+  }
+
```
josealiaga: The cont argument is never used 
When is it decided if it is needed or not the subexpressions?

include/evaluators/blas_tree_evaluator.hpp
```diff
@@ -0,0 +1,302 @@
...
+  size_t getSize() const { return r1.getSize(); }
+  cont_type data() { return l1.data(); }
+
+  bool eval_subexpr_if_needed(cont_type *cont, Device &dev) {
+    l1.eval_subexpr_if_needed(NULL, dev);
+    l2.eval_subexpr_if_needed(NULL, dev);
+    cont_type data1 = l1.data(), data2 = l2.data();
+    r1.eval_subexpr_if_needed(&data1, dev);
+    r2.eval_subexpr_if_needed(&data2, dev);
+    return true;
```
josealiaga: I thought subexpression are computed on a expression tree not on real data.

include/evaluators/blas_tree_evaluator.hpp
```diff
@@ -0,0 +1,302 @@
...
+  }
+
+  value_type eval(size_t i) { return dev_functor::eval(r.eval(i)); }
+  value_type eval(cl::sycl::nd_item<1> ndItem) {
+    return eval(ndItem.get_global(0));
+  }
+};
+
+template <class Functor, class LHS, class RHS, typename Device>
+struct Evaluator<BinaryExpr<Functor, LHS, RHS>, Device> {
```
josealiaga: We should change LHS and RHS by RHS1 and RHS2

include/evaluators/blas_tree_evaluator.hpp
theoden8: This is assignment part, I have not fully fixed it yet, when it comes to reduction. I will discuss it tomorrow. The idea is you either, reduce with assignment or use the result elsewhere.

include/executors/blas1_tree_executor.hpp
```diff
@@ -28,253 +28,267 @@
...
-  using rhs_type = typename Evaluate<RHS>::type;
-  using input_type = BinaryOp<Operator, LHS, RHS>;
-  using type = BinaryOp<Operator, lhs_type, rhs_type>;
-
-  static type convert_to(input_type v, cl::sycl::handler &h) {
-    auto lhs = Evaluate<LHS>::convert_to(v.l, h);
-    auto rhs = Evaluate<RHS>::convert_to(v.r, h);
-    return type(lhs, rhs);
+template <typename Functor, typename LHS, typename RHS>
+struct Converter<Evaluator<BinaryExpr<Functor, LHS, RHS>, SYCLDevice>> {
```
josealiaga: Again, we should use RHS1 and RHS2 instead of LHS and RHS

include/executors/blas1_tree_executor.hpp
```diff
@@ -283,12 +297,19 @@ struct Evaluate<matrix_view<ScalarT, ContainerT>> {
  * @param Tree The Input Expression Tree.
  * @param handler The Command Group Handler used to create the accessors
  */
-template <typename Tree>
-auto make_accessor(Tree t, cl::sycl::handler &h) ->
-    typename Evaluate<Tree>::type {
-  return Evaluate<Tree>::convert_to(t, h);
+namespace detail {
```
josealiaga: The comment should be changed

include/executors/blas_device_sycl.hpp
```diff
@@ -0,0 +1,67 @@
...
+              }
+            } catch (cl::sycl::exception e) {
+              std::cerr << e.what() << std::endl;
+            }
+          }
+        })) {}
+
+  static void parallel_for_setup(size_t &localsize, size_t &nwg,
+                                 size_t &globalsize, size_t N) {
+    localsize = 256;
```
josealiaga: The value used to initiate localSize_ should an attribute of the SYCLDevice, because its optimum value could be different to each device

include/executors/blas_device_sycl.hpp
```diff
@@ -0,0 +1,67 @@
...
+    globalsize = N;
+    nwg = (globalsize + localsize - 1) / localsize;
+    globalsize = nwg * localsize;
```
josealiaga: nWg = (N + localsize - 1) / localsize;

include/executors/blas_packet_traits_sycl.hpp
```diff
@@ -0,0 +1,83 @@
...
+    has_dot = 1,
+    has_length = 1,
+    has_min = 1,
+    has_max = 1
+  };
+};
+
+template <>
+struct Packet_traits<float, SYCLDevice> {
+  using packet_type = cl::sycl::cl_float4;
```
josealiaga: You have defined a packed train for float. 
Does it mean the same template should be created for double, complex and double complex?

include/executors/reduction_sycl.hpp
```diff
@@ -0,0 +1,409 @@
+/***************************************************************************
```
josealiaga: We have to discuss about the reduction

include/operations/blas_operators.hpp
```diff
@@ -55,78 +55,31 @@ namespace blas {
...
-                       typename strip_asp<Type>::type>::value>::type * = 0) {
-    return cl::sycl::abs(val);
-  }
-
-  template <typename Type>
-  static Type eval(const Type &val,
-                   typename std::enable_if<std::is_floating_point<
-                       typename strip_asp<Type>::type>::value>::type * = 0) {
-    return cl::sycl::fabs(val);
+struct wang {
```
josealiaga: How do you compute the abs function in SYCLBLAS?
Morover, I don't know the meaning of wang

include/operations/blas_operators.hpp
```diff
@@ -139,12 +92,11 @@ SYCLBLAS_DEFINE_UNARY_OPERATOR(iniPrdOp1_struct,
...
 SYCLBLAS_DEFINE_UNARY_OPERATOR(posOp1_struct, (r))
 SYCLBLAS_DEFINE_UNARY_OPERATOR(negOp1_struct, (-r))
-SYCLBLAS_DEFINE_UNARY_OPERATOR(
-    sqtOp1_struct,
-    (static_cast<double>(cl::sycl::sqrt(static_cast<double>(r)))))
+SYCLBLAS_DEFINE_UNARY_OPERATOR(sqtOp1_struct, std::sqrt(r))
 SYCLBLAS_DEFINE_UNARY_OPERATOR(tupOp1_struct, r)
 SYCLBLAS_DEFINE_UNARY_OPERATOR(addOp1_struct, (r + r))
 SYCLBLAS_DEFINE_UNARY_OPERATOR(prdOp1_struct, (r * r))
+SYCLBLAS_DEFINE_UNARY_OPERATOR(absOp1_struct, std::abs(r))
```
josealiaga: Now, I know how the abs function is computed

include/operations/blas_operators.hpp
```diff
@@ -139,12 +92,11 @@ SYCLBLAS_DEFINE_UNARY_OPERATOR(iniPrdOp1_struct,
...
 SYCLBLAS_DEFINE_BINARY_OPERATOR(addOp2_struct, const_val::zero, (l + r))
 SYCLBLAS_DEFINE_BINARY_OPERATOR(prdOp2_struct, const_val::one, (l * r))
 SYCLBLAS_DEFINE_BINARY_OPERATOR(divOp2_struct, const_val::one, (l / r))
```
josealiaga: Again, to change (l,r) by (r1,r2)

include/operations/blas_trees.hpp
```diff
@@ -0,0 +1,201 @@
...
+  UnaryExpr(RHS &_r) : r(_r){};
+
+  size_t getSize() const { return r.getSize(); }
+};
+
+/*! BinaryOp.
+ * @brief Implements a Binary Operation (x OP z) with x and z vectors.
+ */
+template <typename Functor, typename LHS, typename RHS>
+struct BinaryExpr {
```
josealiaga: (LHS,RHS) -> (RHS1,RHS2)
(l,r) -> (r1,r2)

include/operations/blas_trees.hpp
```diff
@@ -0,0 +1,201 @@
...
+  using value_type = typename RHS::value_type;
+
+  LHS l;
+  RHS r;
+
+  JoinExpr(LHS &_l, RHS _r) : l(_l), r(_r){};
+
+  // PROBLEM: Only the RHS size is considered. If LHS size is different??
+  // If it is smaller, eval function will crash
+  size_t getSize() const { return r.getSize(); }
```
josealiaga: We can choose the maximum or the minimum
I think maximum is better, but we should avoid the computation on the small vector

include/operations/blas_trees.hpp
```diff
@@ -0,0 +1,201 @@
...
+#include <operations/blas_operators.hpp>
+
+namespace blas {
+
+/** Join.
+ * @brief Joins both sides of the expression in the single kernel.
+ */
+template <class LHS, class RHS>
+struct JoinExpr {
+  using value_type = typename RHS::value_type;
```
josealiaga: (LHS,RHS) -> (RHS1,RHS2)
(l,r) -> (r1,r2)

include/operations/blas_trees.hpp
```diff
@@ -0,0 +1,201 @@
...
+
+  LHS l;
+  RHS r;
+
+  BinaryExpr(LHS &_l, RHS &_r) : l(_l), r(_r){};
+
+  // PROBLEM: Only the RHS size is considered. If LHS size is different??
+  // If it is smaller, eval function will crash
+  size_t getSize() const { return r.getSize(); }
```
josealiaga: We can choose the maximum or the minimum.
Minimum -> some values are not used
Maximum -> Two options for the values which doesn't exist in the small vector: 0 or iniOp.

include/operations/blas_trees.hpp
```diff
@@ -0,0 +1,201 @@
...
+  using value_type = typename RHS::value_type;
+
+  LHS l;
+  RHS r;
+
+  AssignExpr(LHS &_l, RHS _r) : l(_l), r(_r){};
+
+  // PROBLEM: Only the RHS size is considered. If LHS size is different??
+  // If it is smaller, eval function will crash
+  size_t getSize() const { return r.getSize(); }
```
josealiaga: I think we should use l.getSize()
l is the small vector -> some values are not used
l is the large vector  -> Two options for the values which doesn't exist in r: 0 or iniOp.

include/operations/blas_trees.hpp
```diff
@@ -0,0 +1,201 @@
...
+ public:
+  using value_type = typename RHS1::value_type;
+
+  DoubleAssignExpr(LHS1 &_l1, LHS2 &_l2, RHS1 _r1, RHS2 _r2)
+      : l1(_l1), l2(_l2), r1(_r1), r2(_r2){};
+
+  // PROBLEM: Only the RHS size is considered. If LHS size is different??
+  // If it is smaller, eval function will crash
+  size_t getSize() const { return r2.getSize(); }
+};
```
josealiaga: We can choose the maximum or the minimum of lq.getSize() and l2.getSize(),
Minimum -> some values are not used
Maximum -> Two options for the values which doesn't exist in the small vector: 0 or iniOp.
Given that only one of the vectors defines the return value, we can use its size.

include/operations/blas_trees.hpp
josealiaga: We can use the size of the return vector, but I think vectors with different sizes could be joined

include/views/view_sycl.hpp
```diff
@@ -304,6 +265,8 @@ using BufferMatrixView = matrix_view<ScalarT, bufferT<ScalarT>>;
  */
 template <class ScalarT>
 struct matrix_view<ScalarT, bufferT<ScalarT>> {
+  constexpr static bool vectorizable =
+      Packet_traits<ScalarT, SYCLDevice>::Supported;
   using ContainerT = bufferT<ScalarT>;
```
josealiaga: It this the key for vectorize the code?

tests/blas1_interface_test.cpp
```diff
@@ -63,47 +92,39 @@ int main(int argc, char *argv[]) {
...
-      if (i == 0) {
-        diff = (elem * _cos - vX[i] * _sin) - ((vX[i] * _cos + elem * _sin));
-      } else if ((i + 1) == sizeV) {
-        diff += (elem * _cos - vX[i] * _sin) - ((vX[i] * _cos + elem * _sin));
-      }
-      i++;
-    });
-    nrmX = std::sqrt(nrmX);
-    nrmY = std::sqrt(nrmY);
+    /* std::for_each(std::begin(vZ), std::end(vZ), [&](double &elem) { */
```
josealiaga: These computation are useful to verify the correct execution of the SYCLBLAS interface

tests/blas1_interface_test.cpp
```diff
@@ -127,99 +148,105 @@ int main(int argc, char *argv[]) {
...
-#ifdef SHOW_VALUES
-    std::cout << "VALUES!! --> res = " << res << " , diff = " << diff
-              << " , err = " << diff - res << std::endl;
-#endif  //  SHOW_VALUES
-    if (std::abs((res - diff) / res) > ERROR_ALLOWED) {
-      std::cout << "ERROR!! --> res = " << res << " , diff = " << diff
-                << " , err = " << diff - res << std::endl;
-      returnVal += 64;
-    }
+    /* res = vR[0]; */
```
josealiaga: These computation verify the correct execution of the SYCLBLAS interface

include/evaluators/blas1_tree_evaluator.hpp
theoden8: It is reduction and assignment, this is the next thing to figure out

include/evaluators/blas_tree_evaluator.hpp
theoden8: Fair point, I will do that.

include/executors/blas1_tree_executor.hpp
theoden8: Will change.

include/executors/blas_device_sycl.hpp
theoden8: Ok, will change that

include/executors/blas_device_sycl.hpp
theoden8: Ok, this is a little confusing.

include/executors/blas_packet_traits_sycl.hpp
theoden8: It is better to first packetize float, and then packetize everything else.

include/operations/blas_operators.hpp
theoden8: Wang is a pseudo random number generator. Abs function is std::abs, but functor_traits<abs> is cl::sycl::abs, it is redefined for the device.

include/operations/blas_operators.hpp
theoden8: Will do

include/operations/blas_trees.hpp
theoden8: And here too.

include/operations/blas_trees.hpp
theoden8: I don't think we actually need join, will discuss it.

include/operations/blas_trees.hpp
theoden8: Here too.

include/executors/blas1_tree_executor.hpp
theoden8: I will change the comments once I can make sure the refactoring is finished.

include/operations/blas_trees.hpp
theoden8: Do we actually need assignment to smaller/larger vectors? If so, it makes sense to discuss it.

include/operations/blas_trees.hpp
theoden8: Same here, we will need to confirm with the team.

include/operations/blas_trees.hpp
theoden8: I will leave it for further consideration, at the moment it should not be hard to change the code to achieve that functionality, so it is not critical.

include/views/view_sycl.hpp
theoden8: No, vectorization is not yet explicitly written in this code, we will need to vectorize it once we've done with the fusion.

tests/blas1_interface_test.cpp
theoden8: This test becomes replaced with google expression test, but for now we can't verify the result because the reduction is disabled. We will need to figure out how to set up assignment of the reduction.

tests/blas1_interface_test.cpp
theoden8: Again, only axpy/copy/scal/etc should produce the correct result at this stage.

theoden8:
Waiting for placeholder accessor's default constructor and `cgh::require`.

(1) added buffer/placeholder accessor member field into reduction evaluator.
(2) added bind_to to converter to recursively require placeholder accessors for the current cgh.
(3) small refactor of reductions.
(4) NULL -> nullptr

include/interface/blas1_interface_sycl.hpp
```diff
@@ -71,260 +51,162 @@ void _axpy(Device &dev, int _N, ScalarT _alpha,
...
+     * \brief ASUM Takes the sum of the absolute values
+     * @param Device &dev
+     * @param _vx  VectorView
+     * @param _incx Increment in X axis
+     */
+    template <typename X, typename ScalarT, typename ContainerT>
+    auto _asum(int _N, X _x, int _offx, int _incx,
+               vector_view<ScalarT, ContainerT> _rs)
+        RETURNCXX11(make_expr<AssignExpr>(
+            _rs, make_addAbsReductionExpr(make_strdExpr(_x, _offx, _incx, _N))))
```
theoden8: This automatic formatting is so bad!

theoden8:
I'm trying to make Evaluator<StrideExpr> have Evaluator<BreakIfExpr> as a member, which would have both, Evaluator<BreakExpr> and Evaluator<EmptyExpr> and choose in runtime which one to use. There are two problems: (1) convert_to can not return two types from Converter<Evaluator<BreakIfExpr>> and (2) execute_tree does not execute recursively, meaning it is impossible to branch explicitly before enqueueing the kernel. Until this is completed, there can be no automatic fusion when passing an expression instead of a vector as a parameter to an interface function.

include/operations/blas_trees.hpp
```diff
@@ -216,18 +173,35 @@ struct StrideExpr {
...
+
+template <typename ScalarT, typename ContainerT> struct strdExpr_constructor<vector_view<ScalarT, ContainerT>> {
+  using RHS = vector_view<ScalarT, ContainerT>;
+  using rettype = StrideExpr<RHS, MakeHostPointer>;
+  static rettype make(RHS &r, long offset, long stride, size_t N) {
+    return rettype(r, offset, stride, N);
+  }
+};
+
```
josealiaga: The routine fails if different stride or offset is used in related operations on the same vector.

theoden8:
This pull request will be continued in #28.

include/operations/blas_trees.hpp
theoden8: Fixed

Ruyk:
This is addressed in other MR, so the main one can be closed.

PR #20, opened 2017-07-13
Basic testing infrastructure setup

theoden8:
This pull requests sets up the testing framework based on googletests. The changes include modification and addition of cmakelists and test class header.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,7 +24,7 @@ endif()
 
 if(USE_CLBLAS)
   message(STATUS "Will enable comparison with clBLAS")
-  set(CLBLAS_LIBRARIES "${CLBLAS_ROOT}/lib/x86_64-linux-gnu/libclBLAS.so")
+  set(CLBLAS_LIBRARIES "${CLBLAS_ROOT}/lib64/libclBLAS.so")
   set(CLBLAS_INCLUDE "${CLBLAS_ROOT}/include/")
 endif(USE_CLBLAS)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,7 +24,7 @@ endif()
 
 if(USE_CLBLAS)
   message(STATUS "Will enable comparison with clBLAS")
-  set(CLBLAS_LIBRARIES "${CLBLAS_ROOT}/lib64/libclBLAS.so")
+  set(CLBLAS_LIBRARIES "${CLBLAS_ROOT}/lib/x86_64-linux-gnu/libclBLAS.so")
   set(CLBLAS_INCLUDE "${CLBLAS_ROOT}/include/")
 endif(USE_CLBLAS)
 
@@ -42,5 +42,5 @@ if (USE_CLBLAS)
 endif()
 
 
+add_subdirectory(test)
 add_subdirectory(tests)
-
--- a/test/.gitignore
+++ b/test/.gitignore
@@ -1,0 +1,1 @@
+include/
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -1,0 +1,41 @@
+cmake_minimum_required(VERSION 2.8.2)
+
+project(syclblas_gtest)
+
+# compile google test, google mock
+
+include(ExternalProject)
+
+set(SYCLBLAS_TEST ${CMAKE_CURRENT_SOURCE_DIR})
+set(SYCLBLAS_TEST_INCLUDE "${SYCLBLAS_TEST}/include")
+
+ExternalProject_Add(
+  gtest
+  URL https://github.com/google/googletest/archive/release-1.8.0.zip
+  PREFIX ${SYCLBLAS_TEST_INCLUDE}/gtest
+  INSTALL_COMMAND ""
+)
+
+add_library(libgtest IMPORTED STATIC GLOBAL)
+add_dependencies(libgtest gtest)
+
+ExternalProject_Get_Property(gtest source_dir binary_dir)
+set_target_properties(libgtest PROPERTIES
+  IMPORTED_LOCATION "${binary_dir}/googlemock/gtest/libgtest.a"
+  IMPORTED_LINK_INTERFACE_LIBRARIES "${CMAKE_THREAD_LIBS_INIT}"
+)
+set(GTEST_LIB "${source_dir}/googletest/include")
+
+add_library(libgmock IMPORTED STATIC GLOBAL)
+add_dependencies(libgmock gmock)
+
+set_target_properties(libgmock PROPERTIES
+  "IMPORTED_LOCATION" "${binary_dir}/googlemock/libgmock.a"
+  "IMPORTED_LINK_INTERFACE_LIBRARIES" "${CMAKE_THREAD_LIBS_INIT}"
+)
+set(GMOCK_LIB "${source_dir}/googlemock/include")
+
+include_directories(
+  ${GTEST_LIB}
+  ${GMOCK_LIB}
+)
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -1,7 +1,5 @@
 cmake_minimum_required(VERSION 2.8.2)
 
-project(syclblas_gtest)
-
 # compile google test, google mock
 
 include(ExternalProject)
@@ -30,8 +28,8 @@ add_library(libgmock IMPORTED STATIC GLOBAL)
 add_dependencies(libgmock gmock)
 
 set_target_properties(libgmock PROPERTIES
-  "IMPORTED_LOCATION" "${binary_dir}/googlemock/libgmock.a"
-  "IMPORTED_LINK_INTERFACE_LIBRARIES" "${CMAKE_THREAD_LIBS_INIT}"
+  IMPORTED_LOCATION "${binary_dir}/googlemock/libgmock.a"
+  IMPORTED_LINK_INTERFACE_LIBRARIES "${CMAKE_THREAD_LIBS_INIT}"
 )
 set(GMOCK_LIB "${source_dir}/googlemock/include")
 
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8.2)
+cmake_minimum_required(VERSION 3.2.2)
 
 # compile google test, google mock
 
--- a/test/blas1_test.hpp
+++ b/test/blas1_test.hpp
@@ -1,5 +1,5 @@
-#ifndef BLAS1_TEST_HPP_DFQO1OHP
-#define BLAS1_TEST_HPP_DFQO1OHP
+#ifndef BLAS1_TEST_HPP
+#define BLAS1_TEST_HPP
 
 #include <cmath>
 #include <complex>
@@ -13,13 +13,14 @@
 
 #include <interface/blas1_interface_sycl.hpp>
 
-using namespace cl::sycl;
 using namespace blas;
 
 template <typename ClassName>
 struct option_size;
-#define RANDOM_SIZE UINT_MAX
-#define RANDOM_STRD UINT_MAX
+namespace {
+static const size_t RANDOM_SIZE = UINT_MAX;
+static const size_t RANDOM_STRD = UINT_MAX;
+}  // namespace special
 #define REGISTER_SIZE(size, test_name)          \
   template <>                                   \
   struct option_size<class test_name> {         \
@@ -69,6 +70,9 @@ class BLAS1_Test<blas1_test_args<ScalarT_, ExecutorType_>>
   virtual void TearDown() {}
 
   static size_t rand_size() {
+    // make sure the generated number is not too big for a type
+    // i.e. we do not want the sample size to be too big because of
+    // precision/memory restrictions
     size_t ret = rand() >> 5;
     int type_size =
         sizeof(ScalarT) * CHAR_BIT - std::numeric_limits<ScalarT>::digits10 - 2;
@@ -77,17 +81,42 @@ class BLAS1_Test<blas1_test_args<ScalarT_, ExecutorType_>>
            1;
   }
 
+  // it is important that all tests are run with the same test size
+  // so each time we access this function within the same program, we get the
+  // same
+  // randomly generated size
+  size_t test_size() {
+    static bool first = true;
+    static size_t N;
+    if (first) {
+      first = false;
+      N = rand_size();
+    }
+    return N;
+  }
+
+  // getting the stride in the same way as the size above
+  size_t test_strd() {
+    static bool first = true;
+    static size_t N;
+    if (first) {
+      first = false;
+      N = ((rand() & 1) * (rand() % 5)) + 1;
+    }
+    return N;
+  }
+
   template <typename DataType,
-            typename ValueType = typename DataType::value_type>
+            typename value_type = typename DataType::value_type>
   static void set_rand(DataType &vec, size_t _N) {
-    ValueType left(-1), right(1);
+    value_type left(-1), right(1);
     for (size_t i = 0; i < _N; ++i) {
-      vec[i] = ValueType(rand() % int(right - left) * 1000) * .001 - right;
+      vec[i] = value_type(rand() % int(right - left) * 1000) * .001 - right;
     }
   }
 
   template <typename DataType,
-            typename ValueType = typename DataType::value_type>
+            typename value_type = typename DataType::value_type>
   static void print_cont(const DataType &vec, size_t _N,
                          std::string name = "vector") {
     std::cout << name << ": ";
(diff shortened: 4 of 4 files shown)


test/.gitignore
```diff
@@ -0,0 +1 @@
+include/
```
lukeiwanski: Why do we ignore include?

test/blas1_test.hpp
```diff
@@ -0,0 +1,175 @@
+#ifndef BLAS1_TEST_HPP_DFQO1OHP
+#define BLAS1_TEST_HPP_DFQO1OHP
```
lukeiwanski: Why DFQO1OHP ?

test/blas1_test.hpp
```diff
@@ -0,0 +1,175 @@
...
+  ((option_size<test>::value == RANDOM_SIZE) ? test_size<TestClass>() \
+                                             : option_size<test>::value)
+#define TEST_STRD                                                     \
+  ((option_strd<test>::value == RANDOM_SIZE) ? test_strd<TestClass>() \
+                                             : option_strd<test>::value)
+// TEST_PREC determines the precision for the test based on the suggestion for
+// the type
+#define TEST_PREC option_prec<ScalarT, test>::value
+
+#endif
```
lukeiwanski: Can you add label ?

test/blas1_test.hpp
```diff
@@ -0,0 +1,175 @@
...
+#include <iostream>
+#include <vector>
+
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+
+#include <interface/blas1_interface_sycl.hpp>
+
+using namespace cl::sycl;
+using namespace blas;
```
lukeiwanski: Are we sure we want to do that?

test/.gitignore
theoden8: Because this is where gtest is being imported to by ExternalProject

test/blas1_test.hpp
theoden8: Will remove that.

test/blas1_test.hpp
```diff
@@ -0,0 +1,175 @@
...
+
+template <typename ClassName>
+struct option_size;
+#define RANDOM_SIZE UINT_MAX
+#define RANDOM_STRD UINT_MAX
```
gflegar: any reason for using `#define` instead of constants?

test/blas1_test.hpp
```diff
@@ -0,0 +1,175 @@
...
+  using TestClass = BLAS1_Test<TypeParam>;             \
+  using ExecutorType = typename TypeParam::executor_t; \
+  using test = class test_name;
+// TEST_SIZE determines the size based on the suggestion
+#define TEST_SIZE                                                     \
+  ((option_size<test>::value == RANDOM_SIZE) ? test_size<TestClass>() \
+                                             : option_size<test>::value)
+#define TEST_STRD                                                     \
+  ((option_strd<test>::value == RANDOM_SIZE) ? test_strd<TestClass>() \
+                                             : option_strd<test>::value)
```
gflegar: Can these be functions instead of macros?

test/blas1_test.hpp
```diff
@@ -0,0 +1,175 @@
...
+  using ScalarT = typename TypeParam::scalar_t;        \
+  using TestClass = BLAS1_Test<TypeParam>;             \
+  using ExecutorType = typename TypeParam::executor_t; \
+  using test = class test_name;
+// TEST_SIZE determines the size based on the suggestion
+#define TEST_SIZE                                                     \
+  ((option_size<test>::value == RANDOM_SIZE) ? test_size<TestClass>() \
+                                             : option_size<test>::value)
+#define TEST_STRD                                                     \
+  ((option_strd<test>::value == RANDOM_SIZE) ? test_strd<TestClass>() \
```
gflegar: `option_strd<test>::value == RANDOM_STRD`?

gflegar:
One more thing: you're missing the license comment.

test/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+  IMPORTED_LINK_INTERFACE_LIBRARIES "${CMAKE_THREAD_LIBS_INIT}"
+)
+set(GTEST_LIB "${source_dir}/googletest/include")
+
+add_library(libgmock IMPORTED STATIC GLOBAL)
+add_dependencies(libgmock gmock)
+
+set_target_properties(libgmock PROPERTIES
+  "IMPORTED_LOCATION" "${binary_dir}/googlemock/libgmock.a"
+  "IMPORTED_LINK_INTERFACE_LIBRARIES" "${CMAKE_THREAD_LIBS_INIT}"
```
Ruyk: The "IMPORTED_LOCATION" and "IMPORTED_LINK_INTERACE_LIBRARIES" quotes shouldn;t be there

test/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
+cmake_minimum_required(VERSION 2.8.2)
```
Ruyk: Why is this a separate CMake project ? cannot be added to the main one simply by adding the subdirectory?

test/blas1_test.hpp
```diff
@@ -0,0 +1,175 @@
...
+    return vector_view<ValueType, buffer<ValueType>>(buf);
+  }
+
+  template <typename DeviceSelector,
+            typename = typename std::enable_if<
+                std::is_same<ExecutorType, SYCL>::value>::type>
+  static cl::sycl::queue make_queue(DeviceSelector s) {
+    return cl::sycl::queue(s, [=](cl::sycl::exception_list eL) {
+      try {
+        for (auto &e : eL) std::rethrow_exception(e);
```
Ruyk: use always braces:
```
for (auto &e : eL) {
   std::rethrow_execption(e);
}
```

test/blas1_test.hpp
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  template <typename DataType,
+            typename ValueType = typename DataType::value_type>
+  static buffer<ValueType, 1> make_buffer(DataType &vec) {
+    return buffer<ValueType, 1>(vec.data(), vec.size());
+  }
+
+  template <typename ValueType>
+  static vector_view<ValueType, buffer<ValueType>> make_vview(
```
Ruyk: why make vview? make_vector_view seems clearer to me

test/blas1_test.hpp
```diff
@@ -0,0 +1,175 @@
...
+            typename = typename std::enable_if<
+                std::is_same<ExecutorType, SYCL>::value>::type>
+  static cl::sycl::queue make_queue(DeviceSelector s) {
+    return cl::sycl::queue(s, [=](cl::sycl::exception_list eL) {
+      try {
+        for (auto &e : eL) std::rethrow_exception(e);
+      } catch (cl::sycl::exception &e) {
+        std::cout << " E " << e.what() << std::endl;
+      } catch (...) {
+        std::cout << " An exception " << std::endl;
```
Ruyk: Catch and print STD exceptions 
```
catch (std::exception& e) {
    std::cerr << "Standard Exception: " << e.what() << std::endl;
}
```

CMakeLists.txt
```diff
@@ -42,5 +42,5 @@ if (USE_CLBLAS)
 endif()
 
 
+add_subdirectory(test)
 add_subdirectory(tests)
```
Ruyk: maybe tests can be renamed to samples ?

test/CMakeLists.txt
theoden8: it is added as subdirectory, I can remove the project thing though.

test/blas1_test.hpp
theoden8: It is used a lot, other than that I see no problems changing it.

CMakeLists.txt
theoden8: In a separate pull request

test/blas1_test.hpp
theoden8: Maybe it makes sense to remove `using namespace cl::sycl;`, but as a part of the project I think it makes sense to assume that we are using this library's namespace,

test/CMakeLists.txt
theoden8: Removed

test/blas1_test.hpp
theoden8: To avoid compilation errors I thought it would be cheaper to make it into macros than to split into .hpp and .cpp file to avoid symbol duplication in case this header is included in more than one file.

test/blas1_test.hpp
theoden8: Not sure I understand. We are not talking about C labels, right?

test/blas1_test.hpp
theoden8: If the stride is set to be random, it is intended to be some small random number, otherwise it is a specific value.

test/blas1_test.hpp
theoden8: Sure, except for that <test> is aliased to the current test in UNPACK_PARAM above. There should not be a problem changing it.

test/blas1_test.hpp
gflegar: `static const` or anonymous namespace?

test/blas1_test.hpp
gflegar: Sry, should have made it clearer: you have a typo, change `RANDOM_SIZE` -> `RANDOM_STRD` at line 169.

test/blas1_test.hpp
gflegar: I think he's talking about a comment to the `#endif` so it's clear what does this statement close:
```cpp
#endif  // BLAS1_TEST_HPP
```

test/blas1_test.hpp
gflegar: Or it could also be a test class member, and you initialize it to the correct value in the constructor.

test/blas1_test.hpp
theoden8: Ah ok

test/blas1_test.hpp
theoden8: Fixed, thank you

test/blas1_test.hpp
theoden8: Done

test/blas1_test.hpp
theoden8: Done

test/blas1_test.hpp
```diff
@@ -0,0 +1,175 @@
...
+  virtual void SetUp() {}
+  virtual void TearDown() {}
+
+  static size_t rand_size() {
+    size_t ret = rand() >> 5;
+    int type_size =
+        sizeof(ScalarT) * CHAR_BIT - std::numeric_limits<ScalarT>::digits10 - 2;
+    return (ret & (std::numeric_limits<size_t>::max() +
+                   (size_t(1) << (type_size - 2)))) +
+           1;
```
theoden8: Included the explanation.

test/blas1_test.hpp
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  template <typename DataType,
+            typename ValueType = typename DataType::value_type>
+  static void set_rand(DataType &vec, size_t _N) {
+    ValueType left(-1), right(1);
+    for (size_t i = 0; i < _N; ++i) {
+      vec[i] = ValueType(rand() % int(right - left) * 1000) * .001 - right;
```
theoden8: rand() is a temporary solution, we may want to use some other generator in future, which is to be decided.

test/blas1_test.hpp
theoden8: Good point, actually, I moved test_size and test_strd into the test class.

test/blas1_test.hpp
theoden8: Anonymous namespace.

theoden8:
@gflegar here is axpy unit test example:

```cpp
#include "blas1_test.hpp"

typedef ::testing::Types<
    blas1_test_args<float>,
    blas1_test_args<double>
> BlasTypes;

TYPED_TEST_CASE(BLAS1_Test, BlasTypes);

REGISTER_SIZE(::RANDOM_SIZE, axpy_test)
REGISTER_STRD(::RANDOM_STRD, axpy_test)
REGISTER_PREC(float, 1e-4, axpy_test)
REGISTER_PREC(double, 1e-6, axpy_test)
REGISTER_PREC(long double, 1e-7, axpy_test)

B1_TEST(axpy_test) {
  UNPACK_PARAM(axpy_test);
  size_t size = TEST_SIZE;
  size_t strd = TEST_STRD;
  ScalarT prec = TEST_PREC;

  ScalarT alpha(ScalarT(rand() % size_t(size * 1e2)) * 1e-2);
  std::vector<ScalarT> vX(size);
  std::vector<ScalarT> vY(size);
  std::vector<ScalarT> vZ(size, 0);
  TestClass::set_rand(vX, size);
  TestClass::set_rand(vY, size);

  for (auto &d : cl::sycl::device::get_devices()) {
    for (size_t i = 0; i < size; ++i) {
      if (i % strd == 0)
        vZ[i] = alpha * vX[i] + vY[i];
      else
        vZ[i] = vY[i];
    }

    auto q = TestClass::make_queue(d);
    Executor<ExecutorType> ex(q);
    {
      auto buf_vX = TestClass::make_buffer(vX);
      auto buf_vY = TestClass::make_buffer(vY);
      auto view_vX = TestClass::make_vview(buf_vX);
      auto view_vY = TestClass::make_vview(buf_vY);
      _axpy(ex, size, alpha, view_vX, strd, view_vY, strd);
    }
    for (size_t i = 0; i < size; ++i)
        ASSERT_NEAR(vZ[i], vY[i], prec);
  }
}
```

test/blas1_test.hpp
```diff
@@ -0,0 +1,207 @@
...
+        std::cout << "Standard Exception " << e.what() << std::endl;
+      } catch (...) {
+        std::cout << " An exception " << std::endl;
+      }
+    });
+  }
+};
+
+// unpacking the parameters within the test function
+// B is blas_templ_struct
```
theoden8: Forgot to update the comments, oops.

theoden8:
Probably makes sense to change the base for this PR to a new branch, and then make a PR from that to master.

test/blas1_test.hpp
```diff
@@ -0,0 +1,207 @@
...
+template <class ScalarT_, class ExecutorType_>
+class BLAS1_Test<blas1_test_args<ScalarT_, ExecutorType_>>
+    : public ::testing::Test {
+ public:
+  using ScalarT = ScalarT_;
+  using ExecutorType = ExecutorType_;
+
+  BLAS1_Test() {}
+
+  virtual ~BLAS1_Test() {}
```
Ruyk: can you make constructors and destructor default instead of { } ?

test/blas1_test.hpp
```diff
@@ -0,0 +1,207 @@
...
+
+  template <typename DeviceSelector,
+            typename = typename std::enable_if<
+                std::is_same<ExecutorType, SYCL>::value>::type>
+  static cl::sycl::queue make_queue(DeviceSelector s) {
+    return cl::sycl::queue(s, [=](cl::sycl::exception_list eL) {
+      try {
+        for (auto &e : eL) {
+          std::rethrow_exception(e);
+        }
```
Ruyk: note that only the first exception in the list will be processed, and the others will be discarded. Try catch should be inside the for loop to process all of them

test/blas1_test.hpp
```diff
@@ -0,0 +1,207 @@
...
+#define B1_TEST(name) TYPED_TEST(BLAS1_Test, name)
+#define UNPACK_PARAM(test_name)                        \
+  using ScalarT = typename TypeParam::scalar_t;        \
+  using TestClass = BLAS1_Test<TypeParam>;             \
+  using ExecutorType = typename TypeParam::executor_t; \
+  using test = class test_name;
+// TEST_SIZE determines the size based on the suggestion
+#define TEST_SIZE                                                       \
+  ((option_size<test>::value == ::RANDOM_SIZE) ? TestClass::test_size() \
+                                               : option_size<test>::value)
```
Ruyk: why can't this be simply a function that returns a boolean instead of a macro? it is much easier to debug

test/blas1_test.hpp
```diff
@@ -0,0 +1,207 @@
...
+#define TEST_STRD                                                       \
+  ((option_strd<test>::value == ::RANDOM_STRD) ? TestClass::test_strd() \
+                                               : option_strd<test>::value)
+// TEST_PREC determines the precision for the test based on the suggestion for
```
Ruyk: same here

test/blas1_test.hpp
```diff
@@ -0,0 +1,207 @@
...
+// the type
+#define TEST_PREC option_prec<ScalarT, test>::value
```
Ruyk: Can you not use an alias?

test/blas1_test.hpp
```diff
@@ -0,0 +1,207 @@
...
+    static bool first = true;
+    static size_t N;
+    if (first) {
+      first = false;
+      N = rand_size();
+    }
+    return N;
+  }
+
+  // getting the stride in the same way as the size above
```
Ruyk: functions can be moved around, therefore "above" may lose meaning. Each comment must explain the function on its own as much as possible

test/blas1_test.hpp
```diff
@@ -0,0 +1,207 @@
...
+  }
+};
+
+// unpacking the parameters within the test function
+// B is blas_templ_struct
+// TestClass is BLAS1_Test<B>
+// T is default (scalar) type for the test (e.g. float, double)
+// C is the container type for the test (e.g. std::vector)
+// E is the executor kind for the test (sequential, openmp, sycl)
+#define B1_TEST(name) TYPED_TEST(BLAS1_Test, name)
```
Ruyk: what is the point of this macro?

test/blas1_test.hpp
theoden8: Will do that.

test/blas1_test.hpp
theoden8: Included an explanation.

test/blas1_test.hpp
theoden8: Haven't noticed, i have been copying it around for a while, should change it everywhere.

test/blas1_test.hpp
theoden8: No point other than to hide TEST_F, TYPED_TEST etc. from the user code. If it seems confusing rather than helpful, I shall remove it.

test/blas1_test.hpp
theoden8: You mean, a function which returns a size? I don't see how a boolean helps.

Ruyk: yes, a function that returns a size, sorry too fast writing!

PR #21, opened 2017-07-13
Added basic benchmarking infrastructure

theoden8:
Added cmake lists and header file for benchmarking framework. Applied comments here from mehdi-goli#7.

### DIFF
--- a/bench/CMakeLists.txt
+++ b/bench/CMakeLists.txt
@@ -1,5 +1,4 @@
 cmake_minimum_required(VERSION 2.8.2)
-project(syclblas_benchmark)
 
 find_package(PkgConfig)
 
--- a/bench/CMakeLists.txt
+++ b/bench/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8.2)
+cmake_minimum_required(VERSION 3.2.2)
 
 find_package(PkgConfig)
 
--- a/bench/CMakeLists.txt
+++ b/bench/CMakeLists.txt
@@ -1,0 +1,41 @@
+cmake_minimum_required(VERSION 2.8.2)
+project(syclblas_benchmark)
+
+find_package(PkgConfig)
+
+message(STATUS "Path to the SYCL implementation ")
+
+set(SYCLBLAS_INCLUDE "${CMAKE_CURRENT_SOURCE_DIR}/../include")
+set(CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/../cmake/Modules")
+
+include(FindOpenCL)
+include(FindComputeCpp)
+
+pkg_search_module(CLBLAST REQUIRED clblast)
+pkg_check_modules(CLBLAST clblast OPTIONAL)
+if (CLBLAST_FOUND)
+  add_executable(clblast_benchmarks clblast_benchmark.cpp)
+  set_property(TARGET clblast_benchmarks PROPERTY CXX_STANDARD 11)
+  include_directories(${CLBLAST_INCLUDE_DIRS})
+  target_link_libraries(clblast_benchmarks PUBLIC ${CLBLAST_LIBRARIES})
+  target_link_libraries(clblast_benchmarks PUBLIC ${OpenCL_LIBRARIES})
+endif(CLBLAST_FOUND)
+
+pkg_search_module(CLBLAS REQUIRED clBLAS)
+pkg_check_modules(CLBLAS clBLAS OPTIONAL)
+if (CLBLAS_FOUND)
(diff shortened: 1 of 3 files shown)


theoden8:
As an example, here is syclblas benchmark for axpy:

```cpp
#include "blas_benchmark.hpp"

#include <interface/blas1_interface_sycl.hpp>

using namespace blas;

#define UNPACK_PARAM using ScalarT = TypeParam;

template <typename ExecutorType = SYCL>
class SyclBlasBenchmarker {
  cl::sycl::queue q;
  Executor<ExecutorType> ex;

 public:
  SyclBlasBenchmarker()
      : q(cl::sycl::default_selector(),
          [=](cl::sycl::exception_list eL) {
            try {
              for (auto &e : eL) std::rethrow_exception(e);
            } catch (cl::sycl::exception &e) {
              std::cout << " E " << e.what() << std::endl;
            } catch (...) {
              std::cout << " An exception " << std::endl;
            }
          }),
        ex(q) {}

  template <typename ValueType>
  static cl::sycl::buffer<ValueType, 1> mkbuffer(ValueType *data, size_t len) {
    return cl::sycl::buffer<ValueType, 1>(data, len);
  }

  template <typename ValueType>
  static vector_view<ValueType, cl::sycl::buffer<ValueType>> mkvview(
      cl::sycl::buffer<ValueType, 1> &buf) {
    return vector_view<ValueType, cl::sycl::buffer<ValueType>>(buf);
  }

  BENCHMARK_FUNCTION(axpy_bench) {
    UNPACK_PARAM;
    ScalarT *v1 = new_data<ScalarT>(size);
    ScalarT *v2 = new_data<ScalarT>(size);
    ScalarT alpha(2.4367453465);
    double flops;
    {
      auto buf1 = mkbuffer<ScalarT>(v1, size);
      auto buf2 = mkbuffer<ScalarT>(v2, size);
      auto vvw1 = mkvview(buf1);
      auto vvw2 = mkvview(buf2);
      flops = benchmark<>::measure(no_reps, size * 2, [&]() {
        _axpy(ex, size, alpha, vvw1, 1, vvw2, 1);
        q.wait_and_throw();
      });
    }
    release_data(v1);
    release_data(v2);
    return flops;
  }
};

BENCHMARK_MAIN_BEGIN(1 << 1, 1 << 24, 10);
SyclBlasBenchmarker<SYCL> blasbenchmark;

BENCHMARK_REGISTER_FUNCTION("axpy_float", axpy_bench<float>);
BENCHMARK_REGISTER_FUNCTION("axpy_double", axpy_bench<double>);

BENCHMARK_MAIN_END();
```

theoden8:
Probably makes sense to change the base for this PR to a new branch, and then make a PR from that to master.

bench/blas_benchmark.hpp
```diff
@@ -0,0 +1,127 @@
...
+ScalarT *new_data(size_t size, bool initialized = true) {
+  ScalarT *v = new ScalarT[size];
+  if (initialized) {
+    for (size_t i = 0; i < size; ++i) {
+      v[i] = 1e-3 * ((rand() % 2000) - 1000);
+    }
+  }
+  return v;
+}
+#define release_data(ptr) delete[] (ptr);
```
Ruyk: Can this be a function instead of a macro?

bench/blas_benchmark.hpp
```diff
@@ -0,0 +1,127 @@
...
+      dur += end - start;
+    }
+    return (double(flops) * numReps) / (dur.count() * 1e-9);
+  }
+
+  static constexpr const size_t text_name_length = 30;
+  static constexpr const size_t text_iterations_length = 15;
+  static constexpr const size_t text_flops_length = 10;
+
+  static std::string align_left(std::string &&text, size_t len,
```
Ruyk: why is this static?

bench/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
+cmake_minimum_required(VERSION 2.8.2)
```
Ruyk: minimum version of all SYCL Ecosystem projects is CMake 3.2.2

bench/blas_benchmark.hpp
theoden8: Yes, although it would do exactly the same, and there is no way I know to specify that we need a pointer that can be destroyed with `delete []`, not `delete`.

bench/blas_benchmark.hpp
theoden8: Because it is just a utility function, and all it uses are utility variables.

bench/CMakeLists.txt
theoden8: I will change that.

Ruyk:
I'll accept this for now assuming the travis fix will work.

PR #25, opened 2017-08-15
Added benchmarks for clblas, clblast and syclblas

theoden8:
Here I have benchmarks for syclblas (BLAS1 only). Because clblas and clblast benchmarks have a lot in common with this one, I will use the comments from this pr to improve the others before proposing a further change. See #21, mehdi-goli#7.

### DIFF
--- a/bench/clblas_benchmark.cpp
+++ b/bench/clblas_benchmark.cpp
@@ -325,79 +325,35 @@ ClBlasBenchmarker blasbenchmark;
 
 BENCHMARK_REGISTER_FUNCTION("scal_float", scal_bench<float>);
 BENCHMARK_REGISTER_FUNCTION("scal_double", scal_bench<double>);
-/* BENCHMARK_REGISTER_FUNCTION("scal_complex_float", */
-/*                             scal_bench<std::complex<float>>); */
-/* BENCHMARK_REGISTER_FUNCTION("scal_complex_double", */
-/*                             scal_bench<std::complex<double>>); */
 
 BENCHMARK_REGISTER_FUNCTION("axpy_float", axpy_bench<float>);
 BENCHMARK_REGISTER_FUNCTION("axpy_double", axpy_bench<double>);
-/* BENCHMARK_REGISTER_FUNCTION("axpy_complex_float", */
-/*                             axpy_bench<std::complex<float>>); */
-/* BENCHMARK_REGISTER_FUNCTION("axpy_complex_double", */
-/*                             axpy_bench<std::complex<double>>); */
 
 BENCHMARK_REGISTER_FUNCTION("asum_float", asum_bench<float>);
 BENCHMARK_REGISTER_FUNCTION("asum_double", asum_bench<double>);
-/* BENCHMARK_REGISTER_FUNCTION("asum_complex_float", */
-/*                             asum_bench<std::complex<float>>); */
-/* BENCHMARK_REGISTER_FUNCTION("asum_complex_double", */
-/*                             asum_bench<std::complex<double>>); */
 
 BENCHMARK_REGISTER_FUNCTION("nrm2_float", nrm2_bench<float>);
 BENCHMARK_REGISTER_FUNCTION("nrm2_double", nrm2_bench<double>);
-/* BENCHMARK_REGISTER_FUNCTION("nrm2_complex_float", */
-/*                             nrm2_bench<std::complex<float>>); */
-/* BENCHMARK_REGISTER_FUNCTION("nrm2_complex_double", */
-/*                             nrm2_bench<std::complex<double>>); */
 
 BENCHMARK_REGISTER_FUNCTION("dot_float", dot_bench<float>);
 BENCHMARK_REGISTER_FUNCTION("dot_double", dot_bench<double>);
-/* BENCHMARK_REGISTER_FUNCTION("dot_complex_float", */
-/*                             dot_bench<std::complex<float>>); */
-/* BENCHMARK_REGISTER_FUNCTION("dot_complex_double", */
-/*                             dot_bench<std::complex<double>>); */
 
 BENCHMARK_REGISTER_FUNCTION("iamax_float", iamax_bench<float>);
 BENCHMARK_REGISTER_FUNCTION("iamax_double", iamax_bench<double>);
-/* BENCHMARK_REGISTER_FUNCTION("iamax_complex_float", */
-/*                             iamax_bench<std::complex<float>>); */
-/* BENCHMARK_REGISTER_FUNCTION("iamax_complex_double", */
-/*                             iamax_bench<std::complex<double>>); */
 
 /* BENCHMARK_REGISTER_FUNCTION("iamin_float", iamin_bench<float>); */
 /* BENCHMARK_REGISTER_FUNCTION("iamin_double", iamin_bench<double>); */
-/* BENCHMARK_REGISTER_FUNCTION("iamin_complex_float", */
-/*                             iamin_bench<std::complex<float>>); */
-/* BENCHMARK_REGISTER_FUNCTION("iamin_complex_double", */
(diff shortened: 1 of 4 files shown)


bench/clblas_benchmark.cpp
```diff
@@ -0,0 +1,359 @@
...
+  template <>                                             \
+  struct clblasX##postfix<double> {                       \
+    static constexpr const auto func = &clblasD##postfix; \
+  };
+/* template <> struct clblasX##postfix<std::complex<float>> { \ */
+/*   static constexpr const auto value = clblasC##postfix; \ */
+/* }; \ */
+/* template <> struct clblasX##postfix<std::complex<double>> { \ */
+/*   static constexpr const auto value = clblasZ##postfix; \ */
+/* }; */
```
Ruyk: why is this code commented

bench/clblas_benchmark.cpp
```diff
@@ -0,0 +1,359 @@
...
+  template <>                                              \
+  struct clblasiX##postfix<double> {                       \
+    static constexpr const auto func = &clblasiD##postfix; \
+  };
+/* template <> struct clblasiX##postfix<std::complex<float>> { \ */
+/*   static constexpr const auto value = clblasiC##postfix; \ */
+/* }; \ */
+/* template <> struct clblasiX##postfix<std::complex<double>> { \ */
+/*   static constexpr const auto value = clblasiZ##postfix; \ */
+/* }; */
```
Ruyk: why is this code commented?

bench/clblas_benchmark.cpp
```diff
@@ -0,0 +1,359 @@
...
+    double flops;
+    {
+      ScalarT alpha(2.4367453465);
+      MemBuffer<ScalarT> buf1(context, size);
+      cl_event event;
+      flops = benchmark<>::measure(no_reps, size * 1, [&]() {
+        clblasXscal<ScalarT>::func(size, alpha, buf1.dev(), 0, 1, 1,
+                                   context._queue(), 0, NULL, &event);
+        clWaitForEvents(1, &event);
+        clReleaseEvent(event);
```
Ruyk: you should check for OpenCL errors always

bench/clblas_benchmark.cpp
```diff
@@ -0,0 +1,359 @@
...
+                                  context._queue(), 0, NULL, &event);
+        clWaitForEvents(1, &event);
+        clReleaseEvent(event);
+      });
+    }
+    return flops;
+  }
+
+  BENCHMARK_FUNCTION(iamax_bench) {
+    UNPACK_PARAM;
```
Ruyk: I don't think this UNPACK_PARAM macro is needed, its not clear what it does and complicates debugging. Writing the original code is easier to follow

bench/clblas_benchmark.cpp
```diff
@@ -0,0 +1,359 @@
...
+                                  buf2.dev(), 0, 1, scratch[1].dev(), 1,
+                                  context._queue(), 0, NULL, &events[2]);
+        clblasXnrm2<ScalarT>::func(size, bufr.dev(), 2, buf1.dev(), 0, 1,
+                                   scratch[2].dev(), 1, context._queue(), 0,
+                                   NULL, &events[3]);
+        clblasiXamax<ScalarT>::func(size, buf_i.dev(), 0, buf1.dev(), 0, 1,
+                                    scratch[3].dev(), 1, context._queue(), 0,
+                                    NULL, &events[4]);
+        clWaitForEvents(5, events);
+        for (int i = 0; i < 5; ++i) clReleaseEvent(events[i]);
```
Ruyk: Use braces always

bench/clblas_benchmark.cpp
```diff
@@ -0,0 +1,359 @@
...
+BENCHMARK_REGISTER_FUNCTION("nrm2_double", nrm2_bench<double>);
+
+BENCHMARK_REGISTER_FUNCTION("dot_float", dot_bench<float>);
+BENCHMARK_REGISTER_FUNCTION("dot_double", dot_bench<double>);
+
+BENCHMARK_REGISTER_FUNCTION("iamax_float", iamax_bench<float>);
+BENCHMARK_REGISTER_FUNCTION("iamax_double", iamax_bench<double>);
+
+/* BENCHMARK_REGISTER_FUNCTION("iamin_float", iamin_bench<float>); */
+/* BENCHMARK_REGISTER_FUNCTION("iamin_double", iamin_bench<double>); */
```
Ruyk: Why is this commented out?

bench/clblast_benchmark.cpp
```diff
@@ -0,0 +1,314 @@
...
+  /*     cl_event event; */
+  /*     flops = benchmark<>::measure(no_reps, [&]() { */
+  /*       clblast::Amin<ScalarT>(size, buf_i.dev(), 0, buf1.dev(), 0, 1, */
+  /*                               context._queue(), &event); */
+  /*       clWaitForEvents(1, &event); */
+  /*       clReleaseEvent(event); */
+  /*     }); */
+  /*   } */
+  /*   return flops; */
+  /* } */
```
Ruyk: ifdef out with a macro and explain why the code is disabled on the definition of the macro

bench/clblas_benchmark.cpp
theoden8: Removed, it is not as useful at the current stage I believe.

bench/clblas_benchmark.cpp
theoden8: Done that, will push with the next commit.

bench/clblas_benchmark.cpp
theoden8: It declares a c++ interface to complex float and complex double functions in clblas, not very useful for benchmarking.

bench/clblas_benchmark.cpp
theoden8: This code will no longer be there, replaced with `Event::wait(std::vector<Event> &&events)`.

bench/clblast_benchmark.cpp
theoden8: I think the reason was that iamin was not a part of clblast, but with release 1.0 it should be included.

bench/clblas_benchmark.cpp
josealiaga: Some benchmark could be also applied on complex functions

bench/clblas_benchmark.cpp
```diff
@@ -0,0 +1,353 @@
...
+        clblasXscal<ScalarT>::func(size, alpha, buf3.dev(), 0, 1, 1,
+                                   context._queue(), 0, NULL, &event3._cl());
+        Event::wait({event1, event2, event3});
+        Event::release({event1, event2, event3});
+      });
+    }
+    return flops;
+  }
+
+  BENCHMARK_FUNCTION(axpy3op_bench) {
```
theoden8: Not exactly, there is no problem if 1, 2 and 4 makes more sense. This was created initially to make a benchmark for fusion vs no fusion.

josealiaga: It would have been really interesting to talk about the operations before to be defined.
Have you thought to define the same operations for MAGMA?

theoden8: I would be happy to discuss it tomorrow.

bench/clblas_benchmark.cpp
Ruyk: Then use a macro, dont comment code. Comment explanations.

bench/clblas_benchmark.cpp
Ruyk: I am happy with this being clblas only for now, we can always add MAGMA later

josealiaga: It would be a good option for the next future

PR #26, opened 2017-08-16
added unit and expression tests

theoden8:
This pull request follows #20 and contains unit tests (checking one function), and expression tests, checking correctness of calling multiple functions and fusion. There is only one expression test, which mimics `tests/blas1_interface_test`, and a unit test for each blas1 function. For iamax/iamin the constants are only defined for float/double, and stride == 1 because otherwise it fails, not sure why. `_rot` fails, there is an error in implementation. But these errors shouldn't be a part of testing.


test/exprtest/blas1_iface_test.cpp
```diff
@@ -0,0 +1,156 @@
...
+  using ScalarT = typename TypeParam::scalar_t;
+  using ExecutorType = typename TypeParam::executor_t;
+  using TestClass = BLAS1_Test<TypeParam>;
+  using test = class interface1_test;
+
+  size_t size = TestClass::template test_size<test>();
+  size_t strd = TestClass::template test_strd<test>();
+  ScalarT prec = TestClass::template test_prec<test>();
+
+  std::cout << "size == " << size << std::endl;
```
Ruyk: Protect all stdout with a macro so it can be disabled easily. Alternatively use some function that allows to enable or disable a verbose mode, but at this point simply if/def cout with a VERBOSE macro

test/exprtest/blas1_iface_test.cpp
```diff
@@ -0,0 +1,156 @@
...
+      if (i == 0) {
+        ScalarT n1 = x;
+        ScalarT n2 = z;
+        _rotg(n1, n2, _cos, _sin);
+        diff = (z * _cos - x * _sin) - (x * _cos + z * _sin);
+      } else if (i == size - 1) {
+        diff += (z * _cos - x * _sin) - (x * _cos + z * _sin);
+      }
+      giv += ((x * _cos + z * _sin) * (z * _cos - x * _sin));
+    }
```
Ruyk: add some comments to this, can you explain what this code is doing? is it possible to extract part of it to a separate function so its clearer?

test/exprtest/blas1_iface_test.cpp
```diff
@@ -0,0 +1,156 @@
...
+      _dot(ex, size, view_vX, strd, view_vY, strd, view_vU);
+      _swap(ex, size, view_vX, strd, view_vY, strd);
+    }
+    ScalarT prec_sample =
+        std::max(std::numeric_limits<ScalarT>::epsilon() * size * 2,
+                 prec * ScalarT(1e1));
+    EXPECT_LE(prec_sample, prec * 1e4);
+    std::cout << "prec==" << std::fixed
+              << std::setprecision(std::numeric_limits<ScalarT>::digits10)
+              << prec_sample << std::endl;
```
Ruyk: VERBOSE macro

test/exprtest/main.cpp
```diff
@@ -0,0 +1,12 @@
+#include <cstdlib>
+
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+
+int main(int argc, char *argv[]) {
+  int seed = time(NULL) / 30;
```
Ruyk: use a constant seed so the tests can be reproduced.

test/unittest/blas1_asum_test.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+  std::vector<ScalarT> vX(size);
+  TestClass::set_rand(vX, size);
+  std::vector<ScalarT> vR(1, ScalarT(0));
+  ScalarT res(0);
+  for (size_t i = 0; i < size; i += strd) {
+    res += std::abs(vX[i]);
+  }
+
+  for (auto &d : cl::sycl::device::get_devices()) {
```
Ruyk: I think you should be able to select which device is running on the test runner , not test all devices using a for loop, 'cos that means devices cannot be skipped w/o modifying the source code.

test/unittest/blas1_axpy_test.cpp
```diff
@@ -0,0 +1,78 @@
...
+typedef ::testing::Types<blas1_test_args<float>, blas1_test_args<double> >
+    BlasTypes;
+
+TYPED_TEST_CASE(BLAS1_Test, BlasTypes);
+
+REGISTER_SIZE(::RANDOM_SIZE, axpy_test)
+REGISTER_STRD(::RANDOM_STRD, axpy_test)
+REGISTER_PREC(float, 1e-4, axpy_test)
+REGISTER_PREC(double, 1e-6, axpy_test)
+REGISTER_PREC(long double, 1e-7, axpy_test)
```
Ruyk: is this registering precision? isn't it too low?

test/unittest/blas1_rotg_test.cpp
```diff
@@ -0,0 +1,80 @@
...
+  TestClass::set_rand(vY, size);
+
+  for (auto &d : cl::sycl::device::get_devices()) {
+    std::vector<ScalarT> vZ(size, 0);
+    std::vector<ScalarT> vT(size, 0);
+    ScalarT _cos, _sin;
+
+    for (size_t i = 0; i < size; i += strd) {
+      ScalarT x = vX[i], y = vY[i];
+      _rotg(x, y, _cos, _sin);
```
Ruyk: is this because current _rotg implementation is still host only?

test/unittest/blas1_scal_test.cpp
```diff
@@ -0,0 +1,74 @@
...
+  using ScalarT = typename TypeParam::scalar_t;
+  using ExecutorType = typename TypeParam::executor_t;
+  using TestClass = BLAS1_Test<TypeParam>;
+  using test = class scal_test;
+
+  size_t size = TestClass::template test_size<test>();
+  size_t strd = TestClass::template test_strd<test>();
+  ScalarT prec = TestClass::template test_prec<test>();
+
+  ScalarT alpha((rand() % size * 1e2) * 1e-2);
```
Ruyk: why does need to be rand?

test/unittest/main.cpp
```diff
@@ -0,0 +1,16 @@
...
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+
+#include <fcntl.h>
+#include <unistd.h>
+
+int main(int argc, char *argv[]) {
+  int seed = 12345;
+  srand(seed);
+  /* set_seed(); */
```
Ruyk: why is this commented

test/unittest/main.cpp
```diff
@@ -0,0 +1,16 @@
...
+  std::cout << "seed: " << seed << std::endl;
```
Ruyk: use macro to ifdef cout

test/unittest/blas1_axpy_test.cpp
josealiaga: I think it is not low, but I expect relative error was computed

test/exprtest/blas1_iface_test.cpp
josealiaga: This code computes the final values when a givens rotation is applied, and then it computes the dot of the result vectors (giv) and the addition of the difference of the first and last component of the result vectors (diff)

test/unittest/blas1_rotg_test.cpp
theoden8: No, now that I look at it I doubt I understood the function right.

josealiaga: _rotg only computes the scalars which are required by _rot.
In theory, they can be computed where you prefer

test/unittest/blas1_axpy_test.cpp
theoden8: It probably should be higher, meaning I have to reduce random sizes as well. But relative error easily runs away, so I would not stick to that.

test/unittest/blas1_asum_test.cpp
theoden8: Like if I specify `-DSYCL_DEVICE=amd` it compiles as `cl::sycl::amd_selector d;`, but if I don't then as `cl::sycl::default_selector`?

test/unittest/blas1_scal_test.cpp
theoden8: I can fix it to some (pressing random numbers) value.

test/unittest/main.cpp
theoden8: I used to set seed from /dev/urandom, to run tests in a loop, but this behaviour was changed.This is a leftover which will be removed.

test/exprtest/main.cpp
theoden8: I suppose we can use --gtest_random_seed=... option, which defaults to 0 https://github.com/google/googletest/blob/master/googletest/src/gtest.cc#L4631

Ruyk:
Seems there are some formatting differences? Also, missing comments on tests. Apart from that looks ok.

test/exprtest/blas1_iface_test.cpp
theoden8: It's copied from tests/blas1_interface_test

theoden8:
There is a problem with the VERBOSE macro, since it is also used in stuff like vector view and causes compiler error.

theoden8:
device selector @josealiaga: default_selector, amd_selector, intel_selector, host_selector etc.

theoden8:
precision problem for some reason.

theoden8:
See #27, this pull request needs to be closed.

PR #27, opened 2017-08-18
added unit and expression tests (new)

theoden8:
Continued from #26.

There is a bug in reduction tests with the result being incorrect sometimes. I did not have enough time to understand all the reasons for that, but I fixed some of them, like wrong _N passed to the functions (now calculated as `(size+strd-1)/strd`. However, the formula seems not to be correct, and sometimes the result is -nan. One thorough gdb walkthrough should be enough to fix it.


josealiaga:
Note that for negative stride, the size is computed in a different way. 
Please, see the vector_view constructors

theoden8:
I know, but I need to get positive stride working at least.

josealiaga:
I can't find the source files, but do you use the same stride in the two kernels of the reduction?
The first kernel uses the stride to access the elements to be reduced, whereas the second kernel accesses to consecutive elements of the scratch vector.

theoden8:
I certainly do, the first uses evaluator to access an element, the second accesses global memory buffer by index.

theoden8:
I fail to reproduce the error on localsize=nwg=2 && input size = 2-10000. It only works for like hundreds of thousands of floats with precision failure 0.875 for numbers generated as multiples of 1/8 from -1 to 1.

theoden8:
Possibly, the presision mismatch was causing the problem. all tests pass. However, the check will fail due to old travis setup.

Ruyk:
LGTM

PR #28, opened 2017-08-18
Restructuring of SYCL BLAS for better fusion

theoden8:
This is a pull request of almost-complete SYCL BLAS1 restructuring. A few things not quite finished:

(1) `Intrinsic llvm.memmove.p1i8.p1i8.i64 has been generated in function SYCL_...` for iamax/iamin (fixed, l.evalref(i) = r.eval(i) must not access the same element).
(2) Because of the testing problem (see #27) the new interface can not be properly tested, but appears to be working (merged).
(3) StrideExpr: need to compare offt/strd with child's offt/strd, not with 0/1 (done).
(4) One-staged reduction (done).
(5) Tuning localsize/nwg for hardware (done).
(6) Documentation & Refactoring.

Features:

(1) Evaluators separate from expressions.
(2) Executing multiple kernels recursively in the same tree (requires computecpp 0.3.1+).
(2) Device instead of an executor.
(3) More generic interface (possible to pass buffers/subexpressions instead of vector views), and returning an expression which is not immediately evaluated.
(4) Conditional execution inside the interface functions depending on whether the N/strd/offt have changed or not.
(5) Samples & tests adjusted for the renewed interface.


theoden8:
Waiting for #27 and #30 to be merged.

theoden8:
@josealiaga for some reason on big input size 2-staged reduction on CPU is faster than 1-staged.

josealiaga:
Which is the local size? I obtained the best results by using 8 (equal to the number of threads)

theoden8:
Say, localsize = 4096 and nwg == 8, i.e. max shared memory size / shared memory per element / 2 and max number of work groups (cpu) respectively.

josealiaga:
For CPU, localsize=4096 is too big. I found that the best option was localsize=8, which it is equal to the number of threads of the i7-6700K when the hyperthreading is active.

About reduction, there is a threshold to decide when to use one or two stages. 

For GPU, I found 4096 was the maximum size to use 1-stage reduction, whereas biggest sizes should solved by using the 2-stages reduction. I think, I only analyzed the double case, therefore you can verify if the value is also valid for float case.
For CPU, I only tested until 16777216 (16M) because larger sizes are not useful for dense problem, although they are useful to know the threshold for the sparse case.
Thus, your tests have found the threshold for CPU case.

In my code, the reduction is defined as follows:
  auto sizeWG = LOCAL_SIZE;
  auto numWG = (_N <= FIRST_REDUCT_SIZE)? 1: LOCAL_SIZE;
where both values in capital letters are device dependent, 
  - AMD_R9: LOCAL_SIZE=256 & FIRST_REDUCT_SIZE = LOCAL_SIZE * 16;
  - i7-6700K: LOCAL_SIZE=8 & FIRST_REDUCT_SIZE = 16777216 ; 
The CPU threshold is determined as the maximum tested size, because I only consider 1-stage for the CPU case. Now, you can change this value by your bew threshold: 40 millions.

theoden8:
I use i7-6770HQ; with localsize=8 (for 1-staged reduction) the performance improved by a third.

```
asum_float_4194304            10             4314.755802 MFlops
asum_float_8388608            10             3903.078528 MFlops
asum_float_16777216           10             4017.370756 MFlops

nrm2_float_4194304            10             1238.022762 MFlops
nrm2_float_8388608            10             1306.515114 MFlops
nrm2_float_16777216           10             1352.508132 MFlops
```

However, for doubles it looks very bad:

```
asum_double_4194304           10             752.904718 MFlops
asum_double_8388608           10             818.199699 MFlops
asum_double_16777216          10             779.675634 MFlops

nrm2_double_4194304           10             705.329267 MFlops
nrm2_double_8388608           10             683.204067 MFlops
nrm2_double_16777216          10             665.965723 MFlops
```

On clblast, for comparison, we get 2k for both, floats and doubles on both operations (although our code is not vectorized).

Also, for normal operations like axpy I get 4.5 gflops with current localsize, but by changing it to  localsize=8 it decreases to 1.4k. Should this be changed to some other value too?

josealiaga:
I don't know the relationship between double and float computation on i7+SYCL. Perhaps, Mehdi knows this.

Only 2k for clblast? I think it is too small.

About nrm2, I agree you, it makes no sense a new kernel for the sqrt. Basically, the last thread, which writes the result on the output variable, should be compute the sqrt before the assignment. I talked to Mehdi about a similar problem, and he proposed me to create a function which could be applied before the assignment. This option could be done adding a new template parameter.

It is possible that different types of operations requires different localsizes. Basically, there are two types: coalescent operations and reductions. You can test the best option for each case, but in any case, I think 4096 is a really too big value for the CPU case.

theoden8:
Interesting, 8192 works best for coalescent operations, from 4 to 8192 it gradually increases.
I will ask Mehdi if he knows how to help, no matter how high is the localsize if fusion is not happening.

josealiaga:
I was confused. When you write 1.3k in fact it means 1.3kMflops, that is 1.3Gflops, isn't it?
Which clblas are you using?

I thought the maximum localsize in CPU was 4096, therefore it wouldn't be possible to use 8192.

In the tables, is 10  the number of repeats? 

About the dot results, the ratio equal to 2 between float and double execution is reasonable.
But, the iamax results show float is worse than double, is it correct?

theoden8:
Sorry, yes, everything is correct. clblas is failing with segfault on axpy kernel on my machine, so I can only compare with clblast. 8192 is the bound, I was using 4096 and lower because of shared memory size maximum size.

theoden8:
@josealiaga by adding `-Ofast` flag to device compiler (!) flag I achieved 4.5gflops on dot_float and 2.5gflops on dot_double. I made a workaround for nrm2 and now it strikes 4.5 gflops on nrm2_float and 3.7gflops on nrm2_double. asum_double reached 3.6 gflops. Only iamax keeps being depressing with 0.1gflops and 0.3gflops.

josealiaga:
Good news!! 
About the iamax, the problem could be the management of IndVal values or the condtion which is required to compute the maximum. To know where the problem is, you could obtain the performances of a new operation, the computation of the maximum of a vector. Basically, the code should be equal to asum changing addOp2_struct to maxOp2_struct.

theoden8:
maxOp2 runs same as dot/asum/nrm2, 4gflops or more, so I got back to iamax. It looks to me that the bottleneck is maxIndOp2::eval, because if I restructure it from
```cpp
template <typename ScalarT, typename Device>
struct functor_traits<maxIndOp2_struct, ScalarT, Device> : maxIndOp2_struct {
  template <typename R>
  static inline R abs(IndVal<R> r) {
    return functor_traits<absOp1_struct, R, Device>::eval(r.getVal());
  }

  template <typename L, typename R>
  static R eval(L l, R r) {
    if(abs(l) > abs(r)) {
      return l;
    }
    if(abs(l) < abs(r)) {
      return r;
    }
    return (l.getInd() < r.getInd()) ? l : r;
  }
};
```
to
```cpp
template <typename ScalarT, typename Device>
struct functor_traits<maxIndOp2_struct, ScalarT, Device> : maxIndOp2_struct {
  template <typename R>
  static inline R abs(IndVal<R> r) {
    return functor_traits<absOp1_struct, R, Device>::eval(r.getVal());
  }

  template <typename L, typename R>
  static R eval(L l, R r) {
    if(abs(l) > abs(r)) {
      return l;
    }
    if(abs(l) < abs(r)) {
      return r;
    }
    return l; // remove one branch
  }
};
```
The performance triples, although the result will not be correct.

I also checked using a printf that the bottleneck in the kernel occurs before local space barrier, i.e. in the first loop, so maybe it should become a mini reduction of some sort to avoid the delay that so much branching causes?

Also, I changed size_t into int... and now iamax_float peaks at 1gflops and iamax_double peaks at 0.2gflops.

josealiaga:
Could you try the following code?
  template <typename L, typename R>
  static R eval(L l, R r) {
    auto valL = abs(l);
    auto valR = abs(r);
    if(valL > valR) {
      return l;
    }
    if(valL < valR) {
      return r;
    }
    return (l.getInd() < r.getInd()) ? l : r;
  }

theoden8:
No difference. 1gflops and 0.2gflops.

josealiaga:
Which is the overhead of the creation of TupOp and valT1?

theoden8:
```
  template <typename L, typename R>
  static R eval(L l, R r) {
    return (l.getVal() < r.getVal()) ? l : r;
  }
```
Has the about same performance as before (1gflops/.6gflops), but `return l;`, `return r;` or any arithmetics on it work faster (3-5gflops). Branching, however, on value_type seems to bring a reasonable performance.

josealiaga:
I don't understand "... about same performance as before (1gflops/.6gflops), but return l;, return r; or any arithmetics on it work faster (3-5gflops)"
Have you measured the overhead of TupOp?

theoden8:
How do I measure it? Benchmark "asum" on top of tupleOp? In that case asum_float is 5gflops and iasum_float is 4gflops. max_float is 4gflops, and iamax_float is 1gflops. Similar for doubles.

josealiaga:
If you create TupOp and compute asum on it, how do you define the addition between two tuples? 
One option to measure the overhead could be avoid the reduction and assign valT1 as the first or the last element of TupOp. The problem will be if the compiler detects that no computation is made and it optimizes the code.

theoden8:
I define iasum as `eval(L l, R r) { return L(l.getInd(), abs(l) + abs(r)); }`. What do you mean by saying to avoid reduction? You mean to make an axpy-like operation for indval?

josealiaga:
If iasum creates TupOp and the reduction uses the operator you have written, the problem isn't the creation of TupOp. Therefore, the problem is maxIndOp2_struct.
An option could be, at least in theory, that each tread computes the maximum and later locates the index where the maximum is. This computation correspond with the first loop. The problem is how to implement it and if this option obtains better performances

theoden8:
Hm. If in the first loop I change

```cpp
for(int k = i; k < N; k += localsize*nwg) {
    val = dev_functor::eval(val, ev.eval(k));
}
```

into

```cpp
for(int k = i; k < N; k += localsize * nwg) {
    // auto x = ev.eval(k);
    // if(abs(x.getVal()) > abs(val.getVal())) {
    //     val = x;
    // }
    accum_functor_traits<dev_functor>::acc(val, ev.eval(k));
}
```

The performance changes to 2gflops for iamax_float and 2gflops for iamax_double. Is this good enough for all intends and purposes? Basically it is now possible to create an accumulator trait, which optimizes occumulation in the first for loop of a reduction stage.

By the way, gcc-5.4.0 starts segfaulting, but I don't think it is our bug.

theoden8:
Yes, it passes all the tests. The segfaul is during the compilation, not in runtime, it's the compiler's fault.

josealiaga:
Perfect!!!!

theoden8:
Found out clblast is tunable for a machine using a tool that works out of the box. Our performance for reduction is still competitive, great work @josealiaga!

josealiaga:
You are doing a really great work!!!!

theoden8:
Potentially, unrolling should be done by the compiler, but sometimes it doesn't work, so just in case pragma unroll is there.

theoden8:
Now instead of applying the stride on top of child expression, it will use an absolute index and then either, pick up an intermediate result or rhs.data() result (which, in all use-cases atm, is assignment LHS data). Finally, the fusion works the way it should.

josealiaga:
I think it makes no sense to compute (i - getDisp()) / getStrd() each time an eval is computed, it is too expensive. 
Moreover, which is the meaning of i in this expression? If you need apply this expression the only reason is because the index i is not the position to be computed. 
The question, why is the index ii generated. if it is not the position to be evaluated? 
If you say it works, it is perfect, but its working is really strange, and really very difficult to understand.

theoden8:
I agree that it's not very efficient. The plan was to compile three different kernels for the cases to_break, use_rhs_resullt and the inefficient one, and only use the last one when neither of the two were possible to use (i.e. if we have 3 kernels in the tree, we compile 27 kernels and use the most suitable one). All I can do know without remodelling this part is to create eval(i), which uses relative index to perform result[i] = r.eval(i), const_eval(i), uses absolute index and returns rhs_result[i], and abs_eval(i), which uses absolute index to access a temporary or constant buffer, i.e. the ugly part of the code with either, accessing result[(i-offt)/strd] or rhs_result[i].

josealiaga:
To explain my problem, I have to define a specific scenario.

If you want to compute the elements 1, 3, 5, 7 of vector VA (off=1&strd=2&N=4), the eval should be defined on the values (0, 1, 2, 3) never on the values (1, 3, 5, 7). Given that you need to transform the first set of values to the second one, to determine the real elements to be accessed, the expression [off+strd*i, i=0,1,2,3] should be used.

Later, if you want to use an intermediate buffer to store the result, its size would be equal to 4 and again the set of values where eval is defined should be (0, 1, 2, 3). Given that, in this case, the set of values are equal to the elements to be accessed on the intermediate result, then (off=0&strd=1). Therefore, applying the expression [off+strd*i, i=0,1,2,3] the correct set of values is obtained.

My problem is the following. I think you never should use the values (1, 3, 5, 7) in any eval, and therefore the expression [(i-off)/strd] never should be computed. As a summary,
  Correct transformation: (0, 1, 2, 3) -> (1, 3, 5, 7) by using [off+strd*i]
  Wrong transformation: (1, 3, 5, 7) -> (0, 1, 2, 3) by using [(i-off)/strd]

theoden8:
So, does it mean that strideexpr and breakexpr need to merge together? BreakExpr does not "know" what it's parent stride/offset, meaning that if we pass (0, 1, 2, 3) to breakexpr, it will not know whether these (0, 1, 2, 3) it received are the same (0, 1, 2, 3) that it indexes. I would like to avoid putting them into one class because they confuse me too much already. I also see the reason to use a vector view - because it has an optimized element access. Yet I am not sure how to make it easy to understand with all these conditions put into one (or two) nodes.

theoden8:
@Ruyk as soon as this is finished (excluding gemm), should this PR be merged to some branch and from there merged with gemm? This is too huge.

Ruyk:
@mehdi-goli should look into this before merging....

josealiaga:
Somewhere the values (0,1,2,3) are changed to the values (1,3,5,7), and sometimes the original values has to be recovered by using the "wrong transformation". A solution could be avoid the first change, and later to use the "correct transformation" when it is required.

theoden8:
Ok, I will make the change, it does seem a bit odd (and inefficient).

josealiaga:
Perfect!!

theoden8:
I'm stuck. I've suspected that merging both parts into one will be confusing, but I couldn't suspect it would become that and keep growing. One of the problem is that we need to consider parent stride>0, <0 and child stride <0, >0, i.e. all 4 cases, times each case can be optimized for |stride|==1, which makes it 8, all containing reverse-engineered index expressions... Moreover, the current testing setup does not test such cases, so I will need to rewrite at least some of the tests to use cblas. Here's what I'm having now:

```cpp
  value_type &evalref(size_t i) {
    if(!r.to_break) {
      // at the moment, this means that all indices match
      return r.evalref(i);
    } else if(r.use_rhs_result) {
      // this means we need to access rhs result by absolute index
      return r.rhs_evalref(offt + i * strd);
    }
    auto rstrd = r.getStrd();
    int disp = offt, rdisp = r.getDisp();
    if(rstrd == 1 || rstrd == -1) {
      auto pos = disp;
      if(strd > 0) {
        pos += i * strd;
      } else {
        pos += (N - i - 1) * strd;
      }
      if(pos < r.left_bound || pos > r.right_bound) {
        return r.rhs_evalref(pos);
      } else {
        int index = pos - rdisp;
        if(rstrd < 0) {
          index = right_bound - index;
        }
        return r.evalref(index);
      }
    }
    auto pos = disp;
    if(strd > 0) {
      pos += strd * i;
    } else {
      pos += (N - i - 1) * strd;
    }
    if(pos < r.left_bound || pos > r.right_bound) {
      return r.rhs_evalref(pos);
    }
    auto pos0 = pos - rdisp;
    if(pos0 % rstrd == 0) {
      if(rstrd > 0) {
        return r.evalref(pos0 / rstrd);
      } else {
        return r.evalref(r.getSize() - (-pos0 / rstrd) - 1);
      }
    } else {
      return r.rhs_evalref(pos);
    }
  }
```

This is not proprely tested.

theoden8:
I am afraid, in order to properly test it we will eventually have to use cblas, because the case when strideexpr and breakexpr both have negative strides was causing a short circuit in my brain for a while. Also, there is probably a faster way to access children indices, like given the remainders and strides' lcm we can compute the period, and maybe even use_rhs_result=true for when no indices match.

josealiaga:
For instance, I don’t know why rstrd and strd exist, because in theory eval is a function of the vector_view, and only strd should be defined. If you need r, it is because you are implementing an assign, but this operation was defined as a node, and nothing special should be made. I think, if a intermediate value has to be defined, in theory, a vector_view related to the scratch with strd=1 should be defined. 
Morever, as you say, to use both strides generates too many combinations, and therefore it is bad solution. I think there is a simpler solution that it usually is the best option, because to solve individual problems usually generates too complex solution.

About the strides, the only questions should be.
  * What happens if you use several strides in the same expression tree? 
  * How are they applied several strides?
The answer could be computing the product of the strides. Therefore,
  * [strd1=-1 & strd2=-1 => strd12=1]
  * [strd3= 4 & strd4= 2 => strd34=8]
The operator * was defined on vector_views to implement this operation.

Finally, I don’t think the solution was to use clblas.

theoden8:
I am not sure I understand. `eval` is now a function of evaluator, in case of vector_view, only `Evaluator<vector_view,Device>` has eval function. vector_view is an expression-level structure, and is supposed to be device-agnostic. strd is the new stride, and rstrd is the old stride. It actually turned out not having too many branching, considering that compile-time breaking can be made in future.
If you use several strides in the same tree, it applies them one by one without any problems. Yes, I have applied several strides in the reduction_fusion test and break_test. I meant `cblas`, not `clblas`. We should use lapack blas for testing, that's my point.

PR #29, opened 2017-08-22
adding forestexpr to fuse independent trees

theoden8:
@josealiaga This is a prototype of expression evaluating multiple independent trees. unfortunately, it only works for 1 tree due to standard layout error, but there should be a workaround.
I.e. this is like JoinExpr with variadic templates which only works for one subexpression.


josealiaga:
Perfect!! Could you locate and solve the error?

theoden8:
Two things I can't do. First, I need ComputeCpp 031, second, I will need help to fix standard layout problem on tuple of size >1.

josealiaga:
Will you need my help?

theoden8:
I think it will be possible to solve later, when this error is fixed in ComputeCpp.

josealiaga:
Ok. Thank you very much.

theoden8:
@josealiaga, I only pulled from #28, here I'm thinking of how to write something like std::tuple without writing a ton of templated code.

PR #56, opened 2018-07-05
Update CMakeLists to support the 'FindBLAS' module

AdamHarries:
This PR changes the way we look for OpenBLAS, to automatically search for a BLAS implementation (with OpenBLAS as our firm preference), rather than manually specifying the OpenBLAS root as a CMake option. 

Additionally, a number of docker and travis settings are changed to reflect the CMake changes.


Ruyk:
Have you tested locally with the default blas installation of ubuntu for example? (just do apt install libblas or any, and see if the cmake detects and works fine)

AdamHarries:
@Ruyk I've managed to track the bug down. It turns out that cmake versions 3.6 or lower don't have the right macros to detect OpenBLAS. I'm currently working on a workaround to allow us to use the "proper" `include(FindBLAS)` when we can (i.e. when it finds OpenBLAS properly), and fall back to assuming that `BLAS_DIR` (which we ask the user to specify, and where `FindBLAS` searches automatically) contains our OpenBLAS libraries/headers.

CMakeLists.txt
```diff
@@ -14,19 +14,12 @@ include(FindOpenCL)
...
-include(FindBLAS)
+message(STATUS "Searching for BLAS in ${BLAS_DIR}")
 
-if (DEFINED OPENBLAS_ROOT)
-  set(USE_OPENBLAS ON)
-else()
-  set(USE_OPENBLAS OFF)
-endif()
+# Hackily set this to see if it fixes it! 
+set(BLAS_DIR /tmp/OpenBLAS/build)
```
Ruyk: Is this for travis?

build.sh
```diff
@@ -1,5 +1,5 @@
 mkdir build
 cd build
-cmake .. -DCOMPUTECPP_PACKAGE_ROOT_DIR=$1 -DCMAKE_BUILD_TYPE=Release -DOPENBLAS_ROOT=/tmp/OpenBLAS/build
+cmake .. -DCOMPUTECPP_PACKAGE_ROOT_DIR=$1 -DCMAKE_BUILD_TYPE=Release -DBLAS_DIR=/tmp/OpenBLAS/build
```
Ruyk: if you use this script for travis, you shouldn't need the hack inside.
If you need it, then please add a comment explaining why the hack is there

Ruyk:
Some minor comments but looks good otherwise.

CMakeLists.txt
AdamHarries: This is for travis - I was basically just trying to debug *why* it wasn't finding OpenBLAS, before I realised it was a CMake version issue

AdamHarries:
Closing for now until I have a solution for the CMake version issues (we need to support 3.4, but OpenBLAS is only added to FindBLAS in version 3.6).

AdamHarries:
Closing for now until I have a solution for the CMake version issues (we need to support 3.4, but OpenBLAS is only added to FindBLAS in version 3.6).

AdamHarries:
I believe that I might have managed to add a workaround for the pre-3.6 cmake issues. Reopening to let travis test it.

AdamHarries:
@Ruyk @mehdi-goli Having looked over this again, I think it could probably be merged - pending any comments either of you might have of course.

Ruyk:
Looks fine to me

PR #57, opened 2018-07-19
Add GEMM transposed tests to unit test and remove invalid logging code

AdamHarries:
This PR implements the nine possible transposed/other variants for gemm, as well as removing code in logging macros which is invalid when compiled in debug mode. 

The `gemm` tests use a function `_gemm_test_impl`, which implementes the generic functionality common to each test. The individual variants are expressed using the `_GEMM_TEST` macro, which instantiates a test based on the `_gemm_test_impl` function.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -73,7 +73,7 @@ function mak {
 
 function tst {
     pushd build/test
-    ctest -VV --timeout 600
+    ctest -VV --timeout 1200
     popd
 }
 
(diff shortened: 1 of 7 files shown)


AdamHarries:
Okay, so it seems that having all the tests under one umbrella may be too slow for travis. I'll look into breaking them up a bit.

Ruyk:
Note that Travis is testing using the host device, which is known to be particularly slow on jenkins.

PR #60, opened 2018-07-26
WIP: Build system refactoring

mmha:
- General cleanup of the CMake scripts
- Updated FindComputeCpp
- Created CMake package
- Created Conan package
- Moved headers to resemble installed directory structure

Still TBD:
- Re-enable GTest as an ExternalProject?
- Remove dependency on inline variables (and therefore C++17)
- ~~Wait for codeplaysoftware/computecpp-sdk#130 to be merged~~
- Update Travis CI (maybe use [conan-io/conan-docker-tools](https://github.com/conan-io/conan-docker-tools)?)


CMakeLists.txt
```diff
@@ -1,43 +1,62 @@
-cmake_minimum_required (VERSION 3.2.2)
+cmake_minimum_required (VERSION 3.4.3)
```
AdamHarries: Check with @mehdi-goli to see whether we can do this or not. IIRC there's a board that we want to test on that doesn't support new(ish) cmakes (it is definitely below 3.6.0), but it could be that it supports up to version 3.4.3.

CMakeLists.txt
```diff
@@ -1,43 +1,62 @@
...
 
-project(sycl-blas)
+include(${CMAKE_BINARY_DIR}/conan_paths.cmake OPTIONAL)
+include(CTest)
 
-enable_testing()
+list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake/Modules")
+find_package(OpenCL REQUIRED)
+find_package(ComputeCpp REQUIRED)
+find_package(OpenBLAS REQUIRED)
```
AdamHarries: I'm not sure that this is the behavior we want here - although we should always have access to OpenBLAS, we might want to use a different BLAS library for tests and thus should let the user configure it.

CMakeLists.txt
mehdi-goli: @AdamHarries yes, this version is OK

CMakeLists.txt
mehdi-goli: @AdamHarries yes the OpenBLAS should not be mandatory. The default should be OpenBLAS but if user provide  any BLAS library, the user BLAS library should be used instead.

mmha: I removed OpenBLAS from the list of required libraries and added a search for other BLAS implementations, although the latter part is untested for now.

A 3rd party BLAS will only be required for the tests and won't be exported as a dependency.

PR #63, opened 2018-08-13
Add example applications

AdamHarries:
This PR adds a number of example applications that demonstrate usage of the sycl-blas. This addition is mainly motivated by wanting a smaller target for debugging and tweaking, rather than the (rather involved) test suite.

### DIFF
--- a/example/CMakeLists.txt
+++ b/example/CMakeLists.txt
@@ -1,0 +1,30 @@
+cmake_minimum_required(VERSION 3.2.2)
+
+project(syclblas_example)
+
+set(SYCLBLAS_EXAMPLE ${CMAKE_CURRENT_SOURCE_DIR})
+set(SYCLBLAS_EXAMPLE_INCLUDE "${SYCLBLAS_EXAMPLE}/include")
+
+if(SYCL_DEVICE)
+  add_definitions(-DSYCL_DEVICE=${SYCL_DEVICE})
+endif(SYCL_DEVICE)
+
+file(GLOB SYCLBLAS_EXAMPLE_SRCS ./*.cpp)
+
+foreach(example ${SYCLBLAS_EXAMPLE_SRCS})
+
+  message(STATUS " Getting ${example} ")
+
+  get_filename_component(SOURCE_NAME ${example} NAME_WE)
+  message(STATUS " Adding ${SOURCE_NAME} ")
+
+#   include_directories(${COMPUTECPP_INCLUDE_DIRECTORY})
+
+  add_executable(${SOURCE_NAME} ${CMAKE_CURRENT_SOURCE_DIR}/${SOURCE_NAME}.cpp  )
+  target_compile_options(${SOURCE_NAME} PUBLIC ${HOST_COMPILER_OPTIONS})
+  target_link_libraries(${SOURCE_NAME} PUBLIC ${CMAKE_THREAD_LIBS_INIT})
+
+  add_sycl_to_target(${SOURCE_NAME} ${CMAKE_CURRENT_BINARY_DIR}
+    ${CMAKE_CURRENT_SOURCE_DIR}/${SOURCE_NAME}.cpp )
+
+endforeach()
--- a/example/blas1_copy.cpp
+++ b/example/blas1_copy.cpp
@@ -1,0 +1,64 @@
+#include <iostream>
+
+#include "queue/sycl_iterator.hpp"
+#include <interface/blas1_interface.hpp>
+#include <interface/blas2_interface.hpp>
+#include <interface/blas3_interface.hpp>
+
+using namespace blas;
+
+int main(int argc, char const *argv[]) {
+  using ScalarT = float;
+  std::cout << "Example building, cmake working" << std::endl;
+
+  size_t size = 128;
+  long strd = 8;
+
+  std::cout << "size == " << size << std::endl;
+  std::cout << "strd == " << strd << std::endl;
+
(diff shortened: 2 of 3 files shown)


example/blas1_copy.cpp
```diff
@@ -0,0 +1,86 @@
...
+        std::rethrow_exception(e);
+      } catch (cl::sycl::exception &e) {
+        std::cout << "E " << e.what() << std::endl;
+      } catch (std::exception &e) {
+        std::cout << "Standard Exception " << e.what() << std::endl;
+      } catch (...) {
+        std::cout << " An exception " << std::endl;
+      }
+    }
+  });
```
Ruyk: can you move the lambda outside of the queue constructor? will make the example clearer i think.

example/blas1_copy.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+  auto gpu_vX = blas::helper::make_sycl_iteator_buffer<ScalarT>(vX, size);
+  auto gpu_vY = blas::helper::make_sycl_iteator_buffer<ScalarT>(size);
+
+  std::cout << "Built iterator buffers for vX and vY" << std::endl;
+
+  try {
+    _copy(ex, (size + strd - 1) / strd, gpu_vX, strd, gpu_vY, strd);
+  } catch (cl::sycl::exception &e) {
+    std::cout << "E " << e.what() << std::endl;
```
Ruyk: since it is an example, we need a wee clearer error messages

example/blas1_copy.cpp
```diff
@@ -0,0 +1,86 @@
...
+    std::cout << "Standard Exception " << e.what() << std::endl;
+  } catch (...) {
+    std::cout << " An exception " << std::endl;
+  }
+
+  std::cout << "Copy command run" << std::endl;
+
+  ex.copy_to_host(gpu_vY, vY.data());
+
+  std::cout << "Copied data to host" << std::endl;
```
Ruyk: does copy_to_host have an explicit wait? I thought we have changed that

example/blas1_copy.cpp
```diff
@@ -0,0 +1,86 @@
...
+  // check that vX and vY are the same
+  for (size_t i = 0; i < size; ++i) {
+    if (i % strd == 0) {
+      std::cout << "checking: vX[" << i << "] == vY[" << i << "]";
+      std::cout << " with values: [" << vX[i] << ", " << vY[i] << "]"
+                << std::endl;
+      assert(vX[i] == vY[i]);
+    } else {
+      assert(0 == vY[i]);
```
Ruyk: don't use assert for an example, rather use a function to check the results and then ensure the return value of the application is not 0 if there is an error. 
Assert can be disabled by compilation options, which you don't really want on an example

example/blas1_copy.cpp
```diff
@@ -0,0 +1,86 @@
...
+  });
+
+  std::cout << "Built selector and queue" << std::endl;
+
+  Executor<SYCL> ex(q);
+
+  std::cout << "Built executor" << std::endl;
+
+  auto gpu_vX = blas::helper::make_sycl_iteator_buffer<ScalarT>(vX, size);
+  auto gpu_vY = blas::helper::make_sycl_iteator_buffer<ScalarT>(size);
```
Ruyk: iteator?

example/blas1_copy.cpp
```diff
@@ -0,0 +1,86 @@
...
+#include <interface/blas3_interface.hpp>
+
+using namespace blas;
+
+int main(int argc, char const *argv[]) {
+  using ScalarT = float;
+  std::cout << "Example building, cmake working" << std::endl;
+
+  size_t size = 128;
+  long strd = 8;
```
Ruyk: why long and not size_t?

example/blas1_copy.cpp
```diff
@@ -0,0 +1,86 @@
...
+                << std::endl;
+      assert(vX[i] == vY[i]);
+    } else {
+      assert(0 == vY[i]);
+    }
+  }
+
+  std::cout << "Finished execution" << std::endl;
+
+  return 0;
```
Ruyk: return non zero if there was an error

example/blas1_copy.cpp
```diff
@@ -0,0 +1,86 @@
...
+  std::cout << "strd == " << strd << std::endl;
+
+  // create two vectors: vX and vY
+  std::vector<ScalarT> vX(size);
+  std::vector<ScalarT> vY(size, 0);
+
+  // initialise the vector
+  ScalarT left(-1), right(1);
+  for (size_t i = 0; i < size; ++i) {
+    vX[i] = rand() % int((right - left) * 8) * 0.125 - right;
```
Ruyk: use explicit casts (static_cast)

example/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+file(GLOB SYCLBLAS_EXAMPLE_SRCS ./*.cpp)
+
+foreach(example ${SYCLBLAS_EXAMPLE_SRCS})
+
+  message(STATUS " Getting ${example} ")
+
+  get_filename_component(SOURCE_NAME ${example} NAME_WE)
+  message(STATUS " Adding ${SOURCE_NAME} ")
+
+#   include_directories(${COMPUTECPP_INCLUDE_DIRECTORY})
```
Ruyk: why is this commented?

Ruyk:
Good work, thanks for the example

PR #66, opened 2018-08-16
Break blas3 gemm tests into separate compilation units to reduce memory footprint

AdamHarries:
This PR breaks the blas level 3 gemm tests into separate files/compilation units so that they can be compiled on low-memory systems. Previously, the large gemm test file would be compiled into one *large* object, the size of which would exceed the available memory on some systems. 

In order to keep the tests folder readable, this PR also creates subdirectories for each level of the blas.


test/unittest/CMakeLists.txt
```diff
@@ -9,20 +9,36 @@ include_directories(${SYCLBLAS_TEST})
...
+  ${SYCLBLAS_UNITTEST}/blas1/blas1_rotg_test.cpp
+  ${SYCLBLAS_UNITTEST}/blas1/blas1_iamax_test.cpp
+  ${SYCLBLAS_UNITTEST}/blas1/blas1_iamin_test.cpp
+
+  # Blas 2 tests
+  ${SYCLBLAS_UNITTEST}/blas2/blas2_gemv_test.cpp
+  ${SYCLBLAS_UNITTEST}/blas2/blas2_ger_test.cpp
+
+  # Blas 3 tests
+  # ${SYCLBLAS_UNITTEST}/blas3/blas3_gemm_test.cpp
```
Ruyk: why is this commented?

AdamHarries: No reason - I just forgot to clean that bit up! I'll fix it now.

PR #67, opened 2018-08-17
cleaning BLAS2 operations; removing the dead codes.

mehdi-goli:


include/operations/blas2_trees.hpp
```diff
@@ -1015,16 +970,14 @@ struct Ger_Col {
             }
           }
         }
+        // This barrier is mandatory to be sure the data is on the shared
+        // memory
+        ndItem.barrier(cl::sycl::access::fence_space::local_space);
       }
     } else {
-      auto shrSz1 = (shrMemSize / 2);
+      auto shrSz1 = (shrMemSize >> 2);
```
Ruyk: this should be >> 1

include/operations/blas2_trees.hpp
```diff
@@ -1015,16 +970,14 @@ struct Ger_Col {
             }
           }
         }
+        // This barrier is mandatory to be sure the data is on the shared
+        // memory
+        ndItem.barrier(cl::sycl::access::fence_space::local_space);
```
Ruyk: Be aware that this barrier changes the behaviour from before. In this case, you always wait even if the loop trip is one, whereas in the original code it will only wait if there was more than one iteration.
It may be fine, but I am not sure about the intentions of @josealiaga  when writing that original code

PR #68, opened 2018-08-17
Move test for double support to test library, and add cmake options

AdamHarries:
This PR moves the test for double support from the definition site of each test, to the internal definition, and adds a cmake option to disable tests with double precision floats.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,6 +38,30 @@ set(COMPUTECPP_DEVICE_COMPILER_FLAGS
     "${COMPUTECPP_DEVICE_COMPILER_FLAGS} -no-serial-memop -sycl-compress-name")
 message(STATUS "${COMPUTECPP_DEVICE_COMPILER_FLAGS}")
 
+# Check to see if we've disabled double support in the tests
+option(NO_DOUBLE_SUPPORT "Disable double support when testing." off)
+if(NO_DOUBLE_SUPPORT)
+  # Define NO_DOUBLE_SUPPORT for the host cxx compiler
+  add_definitions(-DNO_DOUBLE_SUPPORT)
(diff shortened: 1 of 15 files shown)


CMakeLists.txt
```diff
@@ -38,6 +38,30 @@ set(COMPUTECPP_DEVICE_COMPILER_FLAGS
...
+endif()
+
+# If the user has specified a specific workgroup size for tests, pass that on to the compiler
+if(WGSIZE)
+  add_definitions(-DWGSIZE=${WGSIZE})
+  list(APPEND COMPUTECPP_DEVICE_COMPILER_FLAGS -DWGSIZE=${WGSIZE})
+endif()
+
+# If the user has specified that we should use naive gemm, enable that
+option(NAIVEGEMM "Default to naive GEMM implementations" off)
```
Ruyk: seems this desperately needs an _ between naive and gemm?

test/blas_test.hpp
```diff
@@ -81,10 +81,27 @@ struct blas_templ_struct {
...
+
+// specialisation (with define guard) for double
+// #define NO_DOUBLE_SUPPORT
+#ifndef NO_DOUBLE_SUPPORT
+  template <class MetadataT_ = void, class ExecutorType_ = SYCL>
+  using blas_test_double = blas_templ_struct<double, MetadataT_, ExecutorType_>;
+#else 
+  template <class MetadataT_ = void, class ExecutorType_ = SYCL>
+  using blas_test_double = ::testing::internal::None;
+#endif
```
Ruyk: Interesting!

CMakeLists.txt
mehdi-goli: @AdamHarries I fixed code. So you can add _ between the NAIVE and GEMM. Also between the WG and SIZE

PR #73, opened 2018-08-28
Add GEMM to benchmark suite, and refactor

AdamHarries:


bench/syclblas/syclblas_benchmark_level2.cpp
```diff
@@ -25,17 +25,75 @@
...
+
+  benchmark<>::flops_units_t flops = benchmark<>::measure(
+      reps, n_fl_ops,
+      [&]() {
+        auto event = _gemv(ex, *t_str, m, n, alpha, m_a_gpu, m, v_b_gpu, incX,
+                           beta, v_c_gpu, incY);
+        ex.wait(event);
+      });
+
+      auto event = ex.copy_to_host(v_c_gpu, c_v_gpu_result.data(), rlen);
```
Ruyk: why the indent here?

bench/syclblas/syclblas_benchmark_level2.cpp
```diff
@@ -25,17 +25,75 @@
...
+  ex.template deallocate<ScalarT>(v_c_gpu);
+
+  return flops;
+}
+
+SUITE(ADD(gemv))
+
+
+// TODO: This doesn't seem to work with transposed matrices? 
+// Is it a problem with the routine, or the inputs?
```
Ruyk: Create a github issue for this

bench/blas_benchmark.hpp
```diff
@@ -47,8 +51,37 @@ ScalarT *new_data(size_t size, bool initialized = true) {
   }
   return v;
 }
+
+template <typename ScalarT>
+ScalarT *new_const_data(size_t size, ScalarT value = 0) {
+  ScalarT *v = new ScalarT[size];
```
Ruyk: do we need a raw pointer? can we use unique_ptr at least?

bench/blas_benchmark.hpp
```diff
@@ -100,28 +133,58 @@ struct benchmark {
   }
 };
 
+// why on earth do we need this??
```
Ruyk: good question

bench/range.hpp
```diff
@@ -0,0 +1,238 @@
+#ifndef RANGE_HPP
```
Ruyk: file is missing license header

bench/range.hpp
```diff
@@ -0,0 +1,238 @@
...
+      : v(_low), low(_low), high(_high), mult(_mult) {}
+  SizeRange(const SizeRange& sr)
+      : v(sr.v), low(sr.low), high(sr.high), mult(sr.mult) {}
+  T peek() { return v; }
+  T yield() {
+    T r = v;    // cache the current value
+    v *= mult;  // increment
+    return r;   // return
+  }
+  bool finished() {
```
Ruyk: missing final keyword

bench/range.hpp
```diff
@@ -0,0 +1,238 @@
...
+
+template <typename Range1, typename Range2, typename Range3, typename Range4,
+          typename Range5>
+Range5D<Range1, Range2, Range3, Range4, Range5> nd_range(Range1 r1, Range2 r2,
+                                                         Range3 r3, Range4 r4,
+                                                         Range5 r5) {
+  return Range5D<Range1, Range2, Range3, Range4, Range5>(r1, r2, r3, r4, r5);
+}
+
+#endif  // include guard
```
Ruyk: is this coming from elsewhere? is there a link to the source?

bench/syclblas/syclblas_benchmark_level3.cpp
```diff
@@ -0,0 +1,96 @@
...
+        << std::get<4>(params);
+  return fname.str();
+}
+
+BENCHMARK(gemm, blas_level_3) {
+  using ScalarT = ElemT;
+
+  size_t m = std::get<0>(params);
+  size_t k = std::get<1>(params);
+  size_t n = std::get<2>(params);
```
Ruyk: the three above can be const

bench/syclblas/syclblas_benchmark_level3.cpp
```diff
@@ -0,0 +1,96 @@
...
+  char const *t_a = std::get<3>(params);
+  char const *t_b = std::get<4>(params);
```
Ruyk: why not std::string?

bench/syclblas/syclblas_benchmark_level2.cpp
AdamHarries: After re-running the code, it does actually seem to work for transposed inputs. I think I managed to fix the bug, but didn't remove the comment.

bench/blas_benchmark.hpp
AdamHarries: It turns out that we don't actually need these functions at all - I think they're left over from an earlier refactoring pass. Removing.

bench/range.hpp
AdamHarries: Do you mean the source code in range.hpp? No, it's all written by me...

bench/syclblas/syclblas_benchmark_level3.cpp
AdamHarries: the BLAS interface expects an old c-style string, so avoiding converting to/from string saves a little cognitive overhead.

Ruyk: Ok, let it be...

PR #74, opened 2018-09-19
Removing unnecessary casting from UNARY operators

mehdi-goli:


Ruyk:
Looks good, make sure you squash commits

PR #75, opened 2018-09-20
Target based gemm optimisation

mehdi-goli:
Add "Target" option to enable/disable specific target optimisation at compile time.


CMakeLists.txt
```diff
@@ -56,10 +56,32 @@ endif()
...
   list(APPEND COMPUTECPP_DEVICE_COMPILER_FLAGS -DNAIVE_GEMM)
 endif()
 
+if(DEFINED TARGET)
+message(STATUS "TARGET is defined")
+  if(${TARGET} STREQUAL "INTEL_GPU")
+    message(STATUS "${TARGET} device is chosen")
+    add_definitions(-DINTEL_GPU)
+    list(APPEND COMPUTECPP_DEVICE_COMPILER_FLAGS -DINTEL_GPU)
+  # If the user has specified  RCAR as a target bnackend the optimisation for all other device will be disabled
```
Ruyk: bnackend -> backend

include/interface/blas3_interface.hpp
```diff
@@ -133,20 +129,36 @@ cl::sycl::event _gemm(Executor& ex, char _TransA, char _TransB, IndexType _M,
...
+  else if ((ex.get_device_type() ==
+            Executor::Queue_Interface_Type::device_type::SYCL_RCAR_CVENGINE) ||
+           (ex.get_device_type() ==
+            Executor::Queue_Interface_Type::device_type::SYCL_RCAR_HOST_CPU))
+#endif
+#if defined(DYNAMIC) || defined(RCAR)
+  {
+    BIND_DEFAULT TO_TPARAMS(16, false, 4, 4, 4, 4);
+  }
+#endif
```
Ruyk: This combination of if/else is quite confusing. 
Seems to me you are trying not to compile altogether some combination of parameters, but is not clear what are those. Can you rewrite this to
```
#if defined(RCAR)
// Valid combinations
#elif DYNAMIC
// Valid combinations
...
```

In this particular case, if there is some replication, its is OK in principle.

In general, seems to me we simply should extract R-Car path to a separate backend all together if we specifically want to disable some paths of code at compile time (and template SFINAE cannot do the trick)

include/operations/blas_operators.hpp
```diff
@@ -147,9 +147,7 @@ SYCLBLAS_DEFINE_UNARY_OPERATOR(iniPrdOp1_struct,
                                (constant<R, const_val::one>::value))
 SYCLBLAS_DEFINE_UNARY_OPERATOR(posOp1_struct, (r))
 SYCLBLAS_DEFINE_UNARY_OPERATOR(negOp1_struct, (-r))
-SYCLBLAS_DEFINE_UNARY_OPERATOR(
-    sqtOp1_struct,
-    (static_cast<double>(cl::sycl::sqrt(static_cast<double>(r)))))
+SYCLBLAS_DEFINE_UNARY_OPERATOR(sqtOp1_struct, (cl::sycl::sqrt(r)))
```
Ruyk: I've reviewed a separate MR that has this patch? is this re-appearing?

include/interface/blas3_interface.hpp
mehdi-goli: I was trying to avoid the duplication of the code. So I will duplicate and rewrite it. It will be easier. The problem is not the backend. The problem is the launching an optimised kernel by calculating  certain compile time parameters for each backend based on run-time information. The specialisation constant will solve the issue :)

include/operations/blas_operators.hpp
mehdi-goli: It has been merged in master but I don't know why it is again shown here. I checked the master file and it shows that it is there.

include/interface/blas3_interface.hpp
mehdi-goli: But in general, I think it is better to have a separate backend for R-Car, due to architectural difference.

include/interface/blas3_interface.hpp
```diff
@@ -133,20 +129,31 @@ cl::sycl::event _gemm(Executor& ex, char _TransA, char _TransB, IndexType _M,
...
   }
+#elif defined(INTEL_GPU)
+  BIND_DATA_SIZE(1024, 4096, 1024) TO_TPARAMS(128, false, 4, 4, 16, 16);
+  BIND_DATA_SIZE(10, 1024, 1024) TO_TPARAMS(128, false, 2, 2, 8, 8);
+  BIND_DEFAULT TO_TPARAMS(128, false, 8, 8, 8, 8);
+#elif defined(RCAR)
+  BIND_DEFAULT TO_TPARAMS(16, false, 4, 4, 8, 4);
+#else  // any other specified devices
+  BIND_DATA_SIZE(10, 1024, 1024) TO_TPARAMS(128, true, 1, 1, 16, 16);
+  BIND_DEFAULT TO_TPARAMS(128, false, 8, 8, 16, 16);
```
Ruyk: can you add some comments explaining all this numbers? Is easy to lose track of all the different parameters.

PR #76, opened 2018-09-25
Update the FindComputeCpp cmake module to the most recent version.

AdamHarries:
This PR updates the FindComputeCPP cmake module to the most recent version, and updates the various usages of it (in build.sh for example) in order to use the new variables etc. 

In future, it may be better to add the sdk as a submodule, so that we do not need to manually update the SDK, but this may require some discussion.


Ruyk:
We discussed the submodule a couple of times, but seems people don't like to depend on other projects. We could maybe have the FindComputeCpp as a subtree?

Ruyk:
(No comments on the PR from my side BTW!)

AdamHarries:
Subtrees might be a good idea. We can discuss this offline at some point. Right now though, unfortunately, the current cmake module causes linker errors, so those need to be fixed first.

PR #79, opened 2018-10-02
Benchmark integration 

AdamHarries:
This PR integrates the clBLAST benchmarks with the new benchmarking framework, in order to examine the performance on the existing sycl-blas gemv implementation in intel gpu and cpu platforms.


bench/CMakeLists.txt
```diff
@@ -11,63 +11,58 @@ include(FindOpenCL)
...
-# pkg_check_modules(CLBLAS clBLAS)
+# Add syclblas benchmarks.
+file(GLOB SYCL_BENCHMARK_SRCS
+  ${CMAKE_CURRENT_SOURCE_DIR}/syclblas/syclblas_benchmark_level1.cpp
+  ${CMAKE_CURRENT_SOURCE_DIR}/syclblas/syclblas_benchmark_level2.cpp
+  ${CMAKE_CURRENT_SOURCE_DIR}/syclblas/syclblas_benchmark_level3.cpp
+)
+foreach(syclblas_benchmark ${SYCL_BENCHMARK_SRCS})
+  get_filename_component(sycl_bench_exec ${syclblas_benchmark} NAME_WE)
+  set(sycl_bench_exec, ${syclblas_benchmark})
```
Ruyk: why there is a comma there?

bench/CMakeLists.txt
```diff
@@ -11,63 +11,58 @@ include(FindOpenCL)
...
-#   pkg_search_module(CLBLAS OPTIONAL clBLAS)
-#   pkg_check_modules(CLBLAS clBLAS OPTIONAL)
-#   add_executable(clblas_benchmarks clblas_benchmark.cpp)
-#   set_property(TARGET clblas_benchmarks PROPERTY CXX_STANDARD 11)
-#   include_directories(${CLBLAS_INCLUDE_DIRS})
-#   target_link_libraries(clblas_benchmarks ${CLBLAS_LIBRARIES})
-#   target_link_libraries(clblas_benchmarks ${OpenCL_LIBRARIES})
-# endif(CLBLAS_FOUND)
+# Try and also build clBLAST benchmarks.
+# Manually search for CLBLAST
```
Ruyk: inconsistent usage of clBLAS , CLBLAST, CLBlast... in various places. What is the correct one?

bench/CMakeLists.txt
```diff
@@ -11,63 +11,58 @@ include(FindOpenCL)
...
+if (DEFINED CLBLAST_ROOT) 
+  message(STATUS "CLBlast root specified as : ${CLBLAST_ROOT}")
+  set(CMAKE_PREFIX_PATH ${CLBLAST_ROOT})
+else() 
+  # set a default path if not specified
+  set(CMAKE_PREFIX_PATH "/usr/lib/clblast")
+endif()
```
Ruyk: This else should show a message explaining what path is using by default

bench/clblast/clblast_benchmark_level1.cpp
```diff
@@ -0,0 +1,281 @@
...
+      });
+
+  return flops;
+}
+
+BENCHMARK(scal3op, clblast_level_1) {
+  using ScalarT = ElemT;
+  size_t size = params;
+
+  ScalarT alpha(2.4367453465);
```
Ruyk: If these values are random, please make a note. Otherwise create a constant explaining why those values are selected

bench/clblast/clblast_benchmark_level3.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+  size_t a_size = m * k;
+  std::vector<ScalarT> a_host = benchmark<>::random_data<ScalarT>(a_size);
+  MemBuffer<ScalarT> a(*ex, a_host.data(), a_size);
+
+  size_t b_size = k * n;
+  std::vector<ScalarT> b_host = benchmark<>::random_data<ScalarT>(b_size);
+  MemBuffer<ScalarT> b(*ex, b_host.data(), b_size);
+
+  size_t c_size = m * n;
```
Ruyk: all this sizes can be const

bench/common/blas_benchmark.hpp
```diff
@@ -372,13 +251,14 @@ std::vector<benchmark_instance<ElemT, ExecutorT, ParamT>*> benchmark_suite();
 template <typename Ex, typename ParamT>
 int main_impl(Range<ParamT>* range_param, const unsigned reps, Ex ex,
               output_type output = output_type::STDOUT) {
+  benchmark<>::output_headers(output);
   auto fbenchmarks = benchmark_suite<float, Ex, ParamT>();
   for (auto b : fbenchmarks) {
     run_benchmark(b, range_param, reps, ex, output);
   }
 
 #ifndef NO_DOUBLE_SUPPORT
```
Ruyk: This double negation is a bit confusing. Could we change the macro so is #ifdef  DOUBLE_SUPPORT instead?

bench/syclblas/syclblas_benchmark_level2.cpp
```diff
@@ -23,32 +23,33 @@
...
 #include <interface/blas2_interface.hpp>
 
 using namespace blas;
 
-BENCHMARK_NAME_FORMAT(blas_level_2) {
+BENCHMARK_NAME_FORMAT(syclblas_level_2) {
   std::ostringstream fname;
-  fname << name() << "_" << std::get<0>(params) << "_" << std::get<1>(params)
-        << "_" << std::get<2>(params);
+  fname << typeid(ElemT).name() << "_" << name() << "_" << std::get<0>(params)
```
Ruyk: This introduces a dependency on RTTI for the benchmark, can we avoid it?

bench/CMakeLists.txt
AdamHarries: It appears to be shorthand for "TARGET", i.e. the following are equivalent: 
`set(sycl_bench_exec, ${syclblas_benchmark})`
`set(TARGET sycl_bench_exec ${syclblas_benchmark})`
I'll leave it as the latter, as it should make the intent clearer to later readers.

bench/CMakeLists.txt
AdamHarries: According to github, the correct one should be "CLBlast". Unfortunately, however, cmake variables are case sensitive, so we sometimes need to style it as CLBLAST, as that's what it expects. I'll clean it up where I can, but there are some places where it's not possible.

bench/CMakeLists.txt
AdamHarries: Done

bench/clblast/clblast_benchmark_level1.cpp
AdamHarries: I've added a `random_scalar` method to the benchmarking framework, that should let us unify all the random scalars, and make it clear that they're random (in the same way as `random_data` does for vectors/matrices).

bench/syclblas/syclblas_benchmark_level2.cpp
AdamHarries: I *think* we can. I've thrown together a solution that uses function template specialisation instead, as follows: 

```
// blas_benchmark.hpp
template <typename T>
const inline std::string type_string() {
  return "unknown";
}

template <>
const inline std::string type_string<float>() {
  return "float";
}


template <>
const inline std::string type_string<double>() {
  return "double";
}

// ...
// Inside the benchmark struct
template<typename T>
static std::string typestr() { 
  return type_string<T>();
}
```

Which we use as follows in (say) `syclblas_benchmark_level2.cpp`:

```
BENCHMARK_NAME_FORMAT(syclblas_level_2) {
  std::ostringstream fname;
  fname << benchmark<>::typestr<ElemT>() << "_" << name() << "_" << std::get<0>(params)
        << "_" << std::get<1>(params) << "_" << std::get<2>(params);
  return fname.str();
}
```
It gives us the functionality we want (being able to distinguish between benchmarks based on the scalar type), and *should* decide on the type string statically.

bench/common/blas_benchmark.hpp
mehdi-goli: I think it would be good to create a separate PR for fixing this issue

AdamHarries: I agree - I'm happy to look at it tomorrow if people want? I think it might be best done after merging the two currently open+active PRs.

mehdi-goli: agreed!

PR #84, opened 2018-11-02
Make `buffer_allocator` non-typed

ProGTX:
`buffer_allocator` does not take a template parameter in the SYCL 1.2.1 spec.


ProGTX:
Yes, if #88 compiles, then this PR is redundant

PR #85, opened 2018-11-06
Foundations for rewriting GEMV

AdamHarries:
This PR starts on the foundations for rewriting GEMV, in order to simplify, clean, and optimise the codebase. In particular, it makes the following contributions: 

- A new gemv implementation has been added, currently somewhat naive, but performing well on some platforms. 
- The existing gemv implementation has been deprecated, but not removed, for situations where a new implementation does not perform as well as we want. 
- Internally, a number of parameters such as accesses and transpositions have been "typed" - i.e. changed from arbitrary integers to types representing the semantics of the argument.


include/operations/blas2/gemv_deprecated.hpp
```diff
@@ -0,0 +1,628 @@
...
+template <unsigned int interLoop = 1, bool Lower = true, bool Diag = true,
+          bool Upper = true, bool Unit = false, typename LHS, typename Matrix_t,
+          typename Vector_t>
+Gemv_Row<interLoop, Lower, Diag, Upper, Unit, LHS, Matrix_t, Vector_t> make_Gemv_Row(
+    LHS &l, Matrix_t &matrix, Vector_t &vector, typename Vector_t::IndexType nWG_row,
+    typename Vector_t::IndexType nWG_col, typename Vector_t::IndexType shrMemSize) {
+  return Gemv_Row<interLoop, Lower, Diag, Upper, Unit, LHS, Matrix_t, Vector_t>(
+      l, matrix, vector, nWG_row, nWG_col, shrMemSize);
+}
+
```
mehdi-goli: There is a big space gap here

include/operations/blas2/gemv.hpp
```diff
@@ -0,0 +1,96 @@
...
+
+  value_type eval(IndexType i) {
+    auto dim = vector.getSize();
+
+    // initialise val to the correct type.
+    auto val = iniAddOp1_struct::eval(vector.eval(0));
+
+    for (IndexType j = 0; j < dim; j++) {
+      auto prod = prdOp2_struct::eval(matrix.eval(i, j), vector.eval(j));
+      val = addOp2_struct::eval(val, prod);
```
mehdi-goli: it might be better to add one operator as a template parameter here (lets call it MAD) and the eval function for your functor operator applies  val= cl::sycl::mad(matrix.eval(i,j), vector.eval(j), val)

include/operations/blas2/gemv.hpp
```diff
@@ -0,0 +1,96 @@
...
+    }
+    return l.eval(i) = val;
+  }
+
+  value_type eval(cl::sycl::nd_item<1> ndItem) {
+    IndexType globalId = ndItem.get_global_id(0);
+    IndexType globalSize = ndItem.get_global_range(0);
```
mehdi-goli: extra parameter: here you can simply call eval(ndItem.get_global_id(0)); So no need to rewrite the code

include/operations/blas2/gemv.hpp
```diff
@@ -0,0 +1,96 @@
...
+
+  void bind(cl::sycl::handler &h) {
+    l.bind(h);
+    matrix.bind(h);
+    vector.bind(h);
+  }
+};
+
+template <typename Output_t, typename Matrix_t, typename Vector_t>
+Gemv<Output_t, Matrix_t, Vector_t> make_naive_gemv(Output_t &l,
```
mehdi-goli: If it performs better, don't call it naive

test/unittest/blas2/blas2_gemv_test.cpp
```diff
@@ -147,3 +144,121 @@ TYPED_TEST(BLAS_Test, gemv_test_transposed) {
   ex.template deallocate<ScalarT>(v_b_gpu);
   ex.template deallocate<ScalarT>(v_c_gpu);
 }
+
+// ***********************
+// Testing jose's versions:
```
mehdi-goli: change it to legacy version

include/types/access_types.hpp
```diff
@@ -0,0 +1,97 @@
...
+ private:
+ /**
+ * @enum Layout
+ * @brief The possible layout options for a matrix, expressed algebraically.
+ */
+enum class Layout { RowMajor, ColMajor };
+  const Layout layout;
+
+  static inline Layout layout_from_transposition(Transposition &t) { 
+     if (t.isNormal()) {
```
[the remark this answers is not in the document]
AdamHarries: The problem is that there's also the "conjugate" option - which as I'm aware is subtly different from transposed. In this case, having an "isNormal" is the behaviour that we want most often, so that we know what implementation to pick.

include/operations/blas2/gemv.hpp
AdamHarries: I think this might be suited to later development, after this initial implementation is merged, perhaps?

PR #88, opened 2018-11-12
Significant refactorings to ensure compatibility with ComputeCpp 1.0.2

AdamHarries:
This PR includes a number of changes to ensure compatibility with ComputeCpp 1.0.2. 

Specifically: 
  - The computecpp-sdk is now included as a submodule
  - The copy of `FindComputeCpp.cmake` has been deleted, and replaced with the copy from the computecpp-sdk submodule. 
  - `include/queue/pointer_alias.hpp` and `include/queue/pointer_mapper.hpp` have been replaced with the copies in the computecpp-sdk submodule. 
  - `cl::sycl::default_allocator` has been removed wherever necessary.


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "include/computecpp-sdk"]
+	path = include/computecpp-sdk
```
DuncanMcBain: Can I suggest putting this in modules/ or similar? It's not really just for includes, given that it has other code in it. I always envisioned users being able to set the include dir to `submodule-path/computecpp-sdk/include` so their code could `#include <vptr/virtual_pointer.hpp>` and that would be that.

include/queue/queue_sycl.hpp
```diff
@@ -26,7 +26,7 @@
 #define QUEUE_SYCL_HPP
 
 #include <CL/sycl.hpp>
-#include <queue/pointer_mapper.hpp>
+#include <computecpp-sdk/include/vptr/virtual_ptr.hpp>
```
DuncanMcBain: If you set the include directory for your targets, you'll be able to do vptr/virtual_pointer.hpp which I think is much nicer.

.gitmodules
AdamHarries: That's a really good point - especially given that we're now using the FindComputeCpp.cmake from the sdk.

CMakeLists.txt
```diff
@@ -5,7 +5,8 @@ project(sycl-blas)
 enable_testing()
 
 set(SYCLBLAS_INCLUDE ${CMAKE_CURRENT_SOURCE_DIR}/include)
-set(CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake/Modules)
+list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake/Modules)
+list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/include/computecpp-sdk/cmake/Modules)
 
 include(FindOpenCL)
 include(FindComputeCpp)
```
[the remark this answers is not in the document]
AdamHarries: Nope - I had some cmake issues (unrelated to the FindComputeCpp), so I reverted it in case that was the issue. I think using find_package() would probably be best as well, so I'll go back to that.

.gitmodules
Ruyk: We typically use "external" for this

AdamHarries: ah, I pushed a commit moving this to 'modules' right as you posted this! I'll change it to external - it's not a big change.

PR #89, opened 2018-11-20
Fix gemm bug

mehdi-goli:


mehdi-goli:
This has been embedded in PR#90

PR #93, opened 2018-11-26
Example changes for SYCL-DNN support.

r-potter:
This MR contains the various changes required for SYCL-DNN support. Some of these have already been applied manually, and this MR also pulls in a patch of Mehdi's so it might be smarter to manually apply the individual changes required, rather than just accept the MR. The main changes are:

Fix kernels to respect offsets (required for batch matmul) i.e. the "+ _A.getDisp()" changes.
Fixes to resolve multiple definition via constexpr statics
Modify get_buffer and get_offset to hand const T* pointers
Modify kernel selection to support small matmuls

I've also just noticed some spurious event.waits() that I introduced for debugging that need stripping out.


Ruyk:
This has been added internally on SYCL-BLAS already, so the MR is no longer needed.

PR #94, opened 2018-11-29
Benchmark framework extensions

AdamHarries:
This PR adds functionality to the benchmarks in the form of performance reporting based on both overall, and event timings.

### DIFF
--- a/bench/clblast/clblast_benchmark_level1.cpp
+++ b/bench/clblast/clblast_benchmark_level1.cpp
@@ -40,20 +40,19 @@ BENCHMARK(scal, clblast_level_1) {
   using ScalarT = ElemT;
   const size_t size = params;
 
-  // Alpha 
+  // Alpha
   ScalarT alpha = benchmark<>::random_scalar<ScalarT>();
 
   std::vector<ScalarT> buf1_host = benchmark<>::random_data<ScalarT>(size);
   MemBuffer<ScalarT> buf1(*ex, buf1_host.data(), size);
 
   Event event;
-  benchmark<>::flops_units_t flops =
-      benchmark<>::measure(reps, size * 1, [&]() {
-        clblast::Scal<ScalarT>(size, alpha, buf1.dev(), 0, 1, (*ex)._queue(),
-                               &event._cl());
-        event.wait();
-        event.release();
-      });
+  benchmark<>::datapoint_t flops = benchmark<>::measure(reps, size * 1, [&]() {
+    clblast::Scal<ScalarT>(size, alpha, buf1.dev(), 0, 1, (*ex)._queue(),
+                           &event._cl());
+    event.wait();
+    return event;
+  });
 
   return flops;
 }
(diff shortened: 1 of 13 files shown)


bench/common/blas_benchmark.hpp
```diff
@@ -75,54 +75,219 @@ const inline std::string& type_string<double>() {
...
+        break;
+      case CL_OUT_OF_RESOURCES:
+        std::cerr << "cl_out_of_resources in profiling call" << std::endl;
+        break;
+      case CL_OUT_OF_HOST_MEMORY:
+        std::cerr << "cl_out_of_host_memory in profiling call" << std::endl;
+        break;
+    }
+    // If we've reached this point, something's gone wrong - set the error flag
+    all_ok = false;
```
Ruyk: why not use the default of the switch for this?

bench/common/blas_benchmark.hpp
```diff
@@ -75,54 +75,219 @@ const inline std::string& type_string<double>() {
...
+  check_call(clGetEventProfilingInfo(e._cl(), CL_PROFILING_COMMAND_START,
+                                     sizeof(cl_ulong), &start_time, NULL));
+  check_call(clGetEventProfilingInfo(e._cl(), CL_PROFILING_COMMAND_START,
+                                     sizeof(cl_ulong), &end_time, NULL));
+  e.release();
+  // return the delta
+  if (all_ok) {
+    return (end_time - start_time);
+  } else {
+    return (cl_ulong)(-1);
```
Ruyk: cl_ulong is unsigned long, but you return -1?

bench/common/blas_benchmark.hpp
```diff
@@ -75,54 +75,219 @@ const inline std::string& type_string<double>() {
...
+  void output_data(const std::string& name, int no_reps,
+                   output_type output = output_type::STDOUT) {
+    // Calculate GFlop/s overall, and for the kernel.
+    // As it turns out, Flop/Ns (i.e. floating point operations per nanosecond)
+    // is the same as GFlops/S (i.e. billion floating point operations per
+    // second), so we don't need to do any duration casts between time
+    // intervals, or between flops and gigaflops - we can just compute Flop/Ns
+    // directly.
+    auto _mean_overall_flops = (double)(_n_fl_ops) / _mean_overall_time.count();
+    auto _mean_kernel_flops = (double)(_n_fl_ops) / _mean_kernel_time.count();
```
Ruyk: this should be static cast instead of C cast

bench/syclblas/syclblas_benchmark_level1.cpp
```diff
@@ -45,10 +45,11 @@ BENCHMARK(scal, syclblas_level_1) {
 
   auto in = ex.template allocate<ScalarT>(size);
   ex.copy_to_device(v1.data(), in, size);
-  benchmark<>::flops_units_t flops =
-      benchmark<>::measure(reps, size * 1, [&]() {
+  benchmark<>::datapoint_t flops =
+      benchmark<>::measure(reps, size * 1, [&]() -> cl::sycl::event {
```
Ruyk: why the explicit return type?

bench/common/blas_benchmark.hpp
AdamHarries: good point!

bench/common/blas_benchmark.hpp
AdamHarries: Yes - if there's a problem with the event, we don't really want to fail (this is not a test, after all), but we want to produce the largest number we can. It occurs that this should probably be the equivalent of UINT_MAX for cl_ulong

bench/syclblas/syclblas_benchmark_level1.cpp
AdamHarries: It's not required, but it produces errors more clearly if we forget to return an event.

bench/clblast/clblast_benchmark_level1.cpp
```diff
@@ -71,13 +70,12 @@ BENCHMARK(axpy, clblast_level_1) {
...
-        clblast::Axpy<ScalarT>(size, alpha, buf1.dev(), 0, 1, buf2.dev(), 0, 1,
-                               (*ex)._queue(), &event._cl());
-        event.wait();
-        event.release();
-      });
+  benchmark<>::datapoint_t flops = benchmark<>::measure(reps, size * 2, [&]() {
+    clblast::Axpy<ScalarT>(size, alpha, buf1.dev(), 0, 1, buf2.dev(), 0, 1,
+                           (*ex)._queue(), &event._cl());
+    event.wait();
+    return event;
```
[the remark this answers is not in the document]
AdamHarries: the event.release has been moved to the measuring code outside the lambda - inside timef, in blas_benchmark.cpp

PR #95, opened 2018-12-05
Set `COMPUTECPP_USER_FLAGS`

ProGTX:
Set `COMPUTECPP_USER_FLAGS` instead of `COMPUTECPP_DEVICE_COMPILER_FLAGS`, as required by `FindComputeCpp.cmake`. This is required because of this change to ComputeCpp SDK: https://github.com/codeplaysoftware/computecpp-sdk/pull/149

### DIFF
--- a/cmake/Modules/ConfigureSYCLBLAS.cmake
+++ b/cmake/Modules/ConfigureSYCLBLAS.cmake
@@ -1,7 +1,7 @@
 
 # We add some flags to workaround OpenCL platform bugs, see ComputeCpp documentation
 # COMPUTECPP_USER_FLAGS are used when calling add_sycl_to_target
-set(COMPUTECPP_USER_FLAGS -no-serial-memop -Xclang -cl-mad-enable -O3)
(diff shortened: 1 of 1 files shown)


AdamHarries:
I think you might need to merge in the most recent master, as this is currently timing out on Travis - which suggests to me that the changes that Mehdi made to speed up the Level 3 blas tests aren't in the branch that you're trying to merge.

AdamHarries:
Okay, so after a bit of digging, the tests are timing out because we _also_ set `COMPUTECPP_USER_FLAGS` in the Docker to `-fsycl-split-modules=20`, which speeds up the recompilation that happens every time we run the GEMM. 

I'm not sure what we should do - I have a similar change to this, where we set the `COMPUTECPP_DEVICE_COMPILER_FLAGS` using `list(APPEND...)`, which seems to work. I'm not sure which flags are most appropriate for the library configuration though, or whether we can just update the Docker (or the CMake configuration) to append to user flags instead.

PR #96, opened 2018-12-10
Tiled axpy

AdamHarries:
This PR adds a variant of axpy that supports "tiling". If the conditions are correct, the axpy implementation will attempt to use "tiles" in order to compute sub-ranges of the axpy calculation. At present, the tiles use the default address space, but in future this can be changed to specific address spaces (e.g. private memory) in order to improve performance on specific platforms.

### DIFF
--- a/include/interface/blas1_interface.hpp
+++ b/include/interface/blas1_interface.hpp
@@ -55,14 +55,31 @@ typename Executor::Return_Type _axpy(Executor &ex, IndexType _N, T _alpha,
                                      ContainerT1 &_vy, IncrementType _incy) {
   auto vx = make_vector_view(ex, _vx, _incx, _N);
   auto vy = make_vector_view(ex, _vy, _incy, _N);
+  cl::sycl::event ret;
 
-  std::cout << "vx size: " << vx.size() << std::endl;
-  std::cout << "vy size: " << vy.size() << std::endl;
+  auto tmp = make_vector_view()
 
-  auto scalOp = make_op<ScalarOp, prdOp2_struct>(_alpha, vx);
-  auto addOp = make_op<BinaryOp, addOp2_struct>(vy, scalOp);
-  auto assignOp = make_op<Assign>(vy, addOp);
-  auto ret = ex.execute(assignOp);
+                 std::cout
+             << "vx size: " << vx.size_ << std::endl;
+  std::cout << "vy size: " << vy.size_ << std::endl;
+
+  auto tile_size = 128;
+  auto tiles = (vx.size_ / tile_size) + 1;
+  for (int i = 0; i < tiles; i++) {
+    // Make on chip buffer ??
+    // Copy to on chip buffer ??
+    // Perfrom
+    auto scalOp = make_op<ScalarOp, prdOp2_struct>(_alpha, vx);
+    auto addOp = make_op<BinaryOp, addOp2_struct>(vy, scalOp);
+    auto assignOp = make_op<Assign>(vy, addOp);
+    ret = ex.execute(assignOp);
+    // Copy back into buffer.
+  }
+
+  // auto scalOp = make_op<ScalarOp, prdOp2_struct>(_alpha, vx);
+  // auto addOp = make_op<BinaryOp, addOp2_struct>(vy, scalOp);
+  // auto assignOp = make_op<Assign>(vy, addOp);
+  // auto ret = ex.execute(assignOp);
   return ret;
 }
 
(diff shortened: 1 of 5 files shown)


include/interface/blas1_interface.hpp
```diff
@@ -48,20 +48,89 @@ namespace blas {
...
                                      ContainerT1 &_vy, IncrementType _incy) {
-  auto vx = make_vector_view(ex, _vx, _incx, _N);
-  auto vy = make_vector_view(ex, _vy, _incy, _N);
+  cl::sycl::event ret;
+
+  // For now, only use the tiled version when:
+  //   - _N `mod` TileSize == 0
+  //   - TileSize `mod` _incx == 0
+  //   - _incx == _incy
+  //   - (sizeof(ElemT) * TileSize ) < _N (see SYCL copy bug)
```
Ruyk: SYCL copy bug -> ComputeCpp copy bug, its because our implementation, not because the spec!

test/blas_test.hpp
```diff
@@ -178,6 +178,24 @@ class BLAS_Test<blas_test_args<ScalarT_, MetadataT_, ExecutorType_>>
...
+  }
+
+  template <typename DataType,
+            typename value_type = typename DataType::value_type>
+  static void set_monotonic(DataType &vec, size_t _N, value_type increment) {
+    value_type v = value_type(0);
+    for (size_t i = 0; i < _N; ++i) {
+      vec[i] = v;
+      v += increment;
+    }
```
Ruyk: isn't this std::iota? why the need for a static method instead of using the algorithm?

test/blas_test.hpp
```diff
@@ -178,6 +178,24 @@ class BLAS_Test<blas_test_args<ScalarT_, MetadataT_, ExecutorType_>>
     }
   }
 
+  template <typename DataType,
+            typename value_type = typename DataType::value_type>
+  static void set_const(DataType &vec, size_t _N, value_type c) {
+    for (size_t i = 0; i < _N; ++i) {
+      vec[i] = c;
+    }
```
Ruyk: this is std::fill_n

test/unittest/blas1/blas1_axpy_test.cpp
```diff
@@ -194,6 +190,188 @@ TYPED_TEST(BLAS_Test, axpy_test_vpr) {
...
+    }
+  }
+
+  SYCL_DEVICE_SELECTOR d;
+  auto q = TestClass::make_queue(d);
+  Executor<ExecutorType> ex(q);
+  auto gpu_vX = blas::helper::make_sycl_iterator_buffer<ScalarT>(vX, size);
+  auto gpu_vY = ex.template allocate<ScalarT>(size);
+  ex.copy_to_device(vY.data(), gpu_vY, size);
+  _axpy<8>(ex, (size + strd - 1) / strd, alpha, gpu_vX, strd, gpu_vY, strd);
```
Ruyk: why 8? is it hardcoding the tile-size? use a clear constant if thats the case

include/interface/blas1_interface.hpp
AdamHarries: Oops! Good catch, thanks!

test/blas_test.hpp
AdamHarries: Yes - and it turns out that this function isn't even used, so I'm going to remove it.

test/blas_test.hpp
AdamHarries: Yup - and it also turns out that this function isn't even used any more (it was useful for debugging), so I'll remove it.

test/unittest/blas1/blas1_axpy_test.cpp
AdamHarries: 8 is a tile size that we know works here. I've pulled it out into a `constexpr size_t const` variable so that readers can see it explicitly.

include/interface/blas1_interface.hpp
```diff
@@ -48,20 +48,89 @@ namespace blas {
...
+        auto vy_tile_acc =
+            vy_tile.template get_access<cl::sycl::access::mode::read>(
+                cgh, cl::sycl::range<1>(TileSize), cl::sycl::id<1>(0));
+        auto vy_range_acc =
+            vy.template get_access<cl::sycl::access::mode::write>(
+                cgh, cl::sycl::range<1>(TileSize), cl::sycl::id<1>(i));
+        cgh.copy(vy_tile_acc, vy_range_acc);
+      });
+    }
+
```
Ruyk: you should put this out on a separate function so this one is not that big

AdamHarries:
We're moving this to an internal repository, as it's been pointed out to me that there's no point developing this publicly because it won't be any use without internal ComputeCpp operations.

PR #104, opened 2019-01-30
Add attach detach buffer methods

Rbiessy:
This lets user provide existing buffers to SYCL-BLAS while still using the pointer mapper.

### DIFF
--- a/include/queue/queue_sycl.hpp
+++ b/include/queue/queue_sycl.hpp
@@ -106,6 +106,17 @@ class Queue_Interface<SYCL> {
   inline void deallocate(T *p) const {
     cl::sycl::codeplay::SYCLfree(static_cast<void *>(p), *pointerMapperPtr_);
   }
+  template <typename T, typename ContainerT>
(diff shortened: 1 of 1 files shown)


Ruyk:
What are this methods used for and where? is this for the user interface? Can you add some tests and documentation to ease understanding what the intentions are?
Looking at this I actually think we can also do the other way around, and add an "attach" and "detach" methods to the PointerMapper so you can get a pointer from an existing buffer

Rbiessy:
I'm sorry this PR is not very detailed because I spoke with @mehdi-goli IRL. The idea was more to share code with him, he is the one working on it. These methods let you use buffers that were not created by SYCL-BLAS as long as they are attached.
I'm not sure which way is cleaner since what you propose means exposing the pointer mapper to the user. Either way is fine with me but this is blocker for me right now.

Rbiessy:
We won't go for this solution and use the `buffer_iterator` when possible instead.

PR #106, opened 2019-03-13
Travis fixes

AdamHarries:
This PR adds a dependency on python-pip and enum34 to the Dockerfile, so that we can consistently and correctly build the new "library" version of the sycl-blas on Travis. 

This fix also revealed a bug when passing COMPUTECPP_USER_FLAGS via CMAKE_ARGS, which this PR also fixes. 

Finally, this PR also adds "run_docker.sh", which provides a simple wrapper script around docker with some "sensible defaults", to make local testing of travis configurations easy and fast.

### DIFF
--- a/run_docker.sh
+++ b/run_docker.sh
@@ -6,7 +6,7 @@ export IMPL=COMPUTECPP
 export CXX_COMPILER=g++-7
 export CC_COMPILER=gcc-7
 export TARGET=opencl
-export GIT_SLUG="AdamHarries/sycl-blas"
+export GIT_SLUG="codeplaysoftware/sycl-blas"
 export GIT_BRANCH="master"
 
 
--- a/run_docker.sh
+++ b/run_docker.sh
@@ -1,14 +1,21 @@
 #!/bin/bash
 
 # This script runs the SYCL-BLAS tests using the provided Dockerfile.
-# The intention is to provide a seamless alternative to .travis.yml, so that developers can locally test changes in a (somewhat) platform-agnostic manner without the usual delay that happens with travis.
-   - IMPL=COMPUTECPP CXX_COMPILER=clang++-6.0 CC_COMPILER=clang-6.0 TARGET=opencl
+# The intention is to provide a seamless alternative to .travis.yml, so that
+# developers can locally test changes in a (somewhat) platform-agnostic manner
+# without the usual delay that travis testing entails.
+#
+# By default, this script will compile the SYCL-BLAS with g++-7. Other compilers
+# can be enabled by changing the `CXX_COMPILER` and `CC_COMPILER` environment
+# variables, e.g.:
+#   export CXX_COMPILER=clang++-6.0
+#   export CC_COMPILER=clang-6.0
+# Targets and git "slug" are also equally configurable. By default, the target
+# is OpenCL, and the git repository cloned is codeplay's sycl blas master.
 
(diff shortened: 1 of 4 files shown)


AdamHarries:
It looks like we have a new linker error with clang - I've tested it on various versions, and it seems very reproducible. I can't, however, actually fix the issue, so we may need to defer merging this PR.

run_docker.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+# This script runs the SYCL-BLAS tests using the provided Dockerfile.
+# The intention is to provide a seamless alternative to .travis.yml, so that developers can locally test changes in a (somewhat) platform-agnostic manner without the usual delay that happens with travis.
+   - IMPL=COMPUTECPP CXX_COMPILER=clang++-6.0 CC_COMPILER=clang-6.0 TARGET=opencl
```
mehdi-goli: I don't see any issue using clang-7.0. unless there is a particular reason, feel free to switch to clang-7.0

run_docker.sh
```diff
@@ -0,0 +1,25 @@
...
+
+export IMPL=COMPUTECPP
+# export CXX_COMPILER=g++-7
+export CXX_COMPILER=clang++-6.0
+# export CC_COMPILER=gcc-7
+export CC_COMPILER=clang-6.0
```
mehdi-goli: same as above

run_docker.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+# This script runs the SYCL-BLAS tests using the provided Dockerfile.
+# The intention is to provide a seamless alternative to .travis.yml, so that developers can locally test changes in a (somewhat) platform-agnostic manner without the usual delay that happens with travis.
+   - IMPL=COMPUTECPP CXX_COMPILER=clang++-6.0 CC_COMPILER=clang-6.0 TARGET=opencl
+
+export IMPL=COMPUTECPP
+# export CXX_COMPILER=g++-7
+export CXX_COMPILER=clang++-6.0
```
mehdi-goli: same as above

run_docker.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+# This script runs the SYCL-BLAS tests using the provided Dockerfile.
+# The intention is to provide a seamless alternative to .travis.yml, so that developers can locally test changes in a (somewhat) platform-agnostic manner without the usual delay that happens with travis.
+   - IMPL=COMPUTECPP CXX_COMPILER=clang++-6.0 CC_COMPILER=clang-6.0 TARGET=opencl
+
+export IMPL=COMPUTECPP
+# export CXX_COMPILER=g++-7
```
mehdi-goli: remove commented code

run_docker.sh
```diff
@@ -0,0 +1,25 @@
...
+export CXX_COMPILER=clang++-6.0
+# export CC_COMPILER=gcc-7
```
mehdi-goli: remove commented code

run_docker.sh
AdamHarries: Clang 7 isn't available in ubuntu xenial (16.04), so we can only support clang 6. If we want to switch our travis to clang 7, we'd also have to upgrade to 18.04 - which we might want to do, but I think it's beyond the scope of this PR.

PR #108, opened 2019-03-18
Roofline-model analysis: adaptation of Google Benchmarks and creation of Python scripts

Nyrio:
Included in this MR:
*  Modifications of the Google Benchmarks
   * Converted the legacy CLBLAST GEMM benchmark to Google Benchmarks, adapted the structure of the `gbench` folder and cmake files accordingly.
   * Modified the GEMM benchmark to save the number of floating point operations and bytes processed
   * Fixed event time for GEMM benchmarks: now counting the average, best and total
   * A few other changes to the GEMM benchmark such as switching to ranges instead of loops to define the bounds.
* Python scripts: the `roofline` folder contains a bunch of Python files to generate the roofline graph.

I will continue to move the legacy benchmarks to Google Benchmarks, and make the same changes in the other benchmarks that I did for GEMM, but this is not the purpose of this MR.


gbench/clblast/blas3/gemm.cpp
```diff
@@ -0,0 +1,160 @@
...
+void BM_Gemm(benchmark::State& state) {
+  // Standard test setup.
+  char const* t_a = benchmark::utils::from_transpose_enum(
+      static_cast<benchmark::utils::Transposition>(state.range(0)));
+  char const* t_b = benchmark::utils::from_transpose_enum(
+      static_cast<benchmark::utils::Transposition>(state.range(1)));
+  const index_t m = static_cast<index_t>(state.range(2));
+  const index_t k = static_cast<index_t>(state.range(3));
+  const index_t n = static_cast<index_t>(state.range(4));
+/* "c"*/
```
AdamHarries: Could you explain this commented "c"? If not, please remove it.

gbench/clblast/cli_device_selector.hpp
```diff
@@ -0,0 +1,107 @@
+/* Copyright (c) 2015 The Khronos Group Inc.
```
AdamHarries: @mehdi-goli @Ruyk Is this license correct? I believe this file has been copied/moved/adapted from the old benchmarks, but I don't know if the copyright is strictly correct, or if it should be copyright Codeplay.

gbench/clblast/utils.hpp
```diff
@@ -0,0 +1,118 @@
...
+    return 0xFFFFFFFFFFFFFFFF;
+  }
+}
+
+/**
+ * @fn warmup
+ * @brief Warm up to avoid benchmarking data transfer
+ */
+template <typename F, typename... Args>
+inline cl_ulong warmup(F func, Args &&... args) {
```
AdamHarries: Is this not `void` instead of `cl_ulong`?

Nyrio: Yes, thanks for catching that

gbench/clblast/blas3/gemm.cpp
Nyrio: I have no idea why it's there, thanks for catching that

PR #115, opened 2019-03-27
Added and fixed GEMV unittests

RossBrunton:
Added tests for the alpha, beta and lda parts of gemv, and also
made one of the tests actually check its output.

### DIFF
--- a/test/unittest/blas2/blas2_gemv_test.cpp
+++ b/test/unittest/blas2/blas2_gemv_test.cpp
@@ -25,119 +25,105 @@
 
 #include "blas_test.hpp"
 
-typedef ::testing::Types<blas_test_float<>, blas_test_double<> > BlasTypes;
+template <typename T>
+using combination_t = std::tuple<int, int, bool, T, T, int, int, int>;
 
-TYPED_TEST_CASE(BLAS_Test, BlasTypes);
+template <typename scalar_t>
+void run_test(const combination_t<scalar_t> combi) {
+  using type_t = blas_test_args<scalar_t, void>;
+  using blas_test_t = BLAS_Test<type_t>;
+  using executor_t = typename type_t::executor_t;
 
-REGISTER_PREC(float, 1e-4, gemv_test)
-REGISTER_PREC(double, 1e-8, gemv_test)
+  int m;
+  int n;
+  bool trans;
+  scalar_t alpha;
+  scalar_t beta;
+  int incX;
+  int incY;
+  int lda_mul;
+  std::tie(m, n, alpha, beta, trans, incX, incY, lda_mul) = combi;
 
-TYPED_TEST(BLAS_Test, gemv_test) {
-  using scalar_t = typename TypeParam::scalar_t;
-  using ExecutorType = typename TypeParam::executor_t;
-  using TestClass = BLAS_Test<TypeParam>;
-  using test = class gemv_test;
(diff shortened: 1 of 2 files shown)


test/unittest/blas2/blas2_gemv_test.cpp
```diff
@@ -145,6 +145,186 @@ TYPED_TEST(BLAS_Test, gemv_test_transposed) {
...
+
+TYPED_TEST(BLAS_Test, gemv_test_legacy_lda) {
+  using scalar_t = typename TypeParam::scalar_t;
+  using ExecutorType = typename TypeParam::executor_t;
+  using TestClass = BLAS_Test<TypeParam>;
+  using test = class gemv_test_legacy_lda;
+
+  int m = 125;
+  int n = 127;
+  int lda = m * 2;
```
mehdi-goli: @RossBrunton  this lda is not valid in blas
 LDA must be at least
*           max( 1, m ).
*           Unchanged on exit.

test/unittest/blas2/blas2_gemv_test.cpp
```diff
@@ -145,6 +145,186 @@ TYPED_TEST(BLAS_Test, gemv_test_transposed) {
...
+  using test = class gemv_test_legacy_beta;
+
+  int m = 125;
+  int n = 127;
+  int lda = m;
+  int incX = 1;
+  int incY = 1;
+  const char* t_str = "n";  // Testing the no transpose matrix
+  scalar_t prec = TestClass::template test_prec<test>();
+  scalar_t alpha = scalar_t(1.0);
```
mehdi-goli: in the main test you can set std::arry<scalar_t, 4>(0.0, 1.0, 4.0, 5.0); for alpha and beta, similarly for m, and n, you can look into the gemm_def.hpp if you want to test multiple sizes and values. In this way we can remove alot of code duplication

test/unittest/blas2/blas2_gemv_test.cpp
RossBrunton: But m * 2 > max(1, m).

test/unittest/blas2/blas2_gemv_test.cpp
RossBrunton: My concern with this is that it hides the reasons the tests are failing. For example, if you can see that the "beta" tests are failing, it means that something to do with "beta" is wrong. I can see where code duplication could be an issue though.

test/unittest/blas2/blas2_gemv_test.cpp
mehdi-goli: max( 1, m ) this means that for m>1 lda cannot be bigger than m.

RossBrunton:
Need to make it so that trans/no trans are in different files to avoid duplication.

test/unittest/blas2/blas2_gemv_test.cpp
RossBrunton: The spec says "LDA must be at least max( 1, m ).", assuming m is 100, this would be "LDA must be at least 100".

LDA is the total number of rows of the matrix which A is a submatrix of, it doesn't make sense for the submatrix to be larger than the outer matrix.

mehdi-goli: LDA is the leading dimension of Matrix A, if the matrix A is M*N and the matrix is colum major the LDA is M.

PR #116, opened 2019-03-28
Removed experimental GEMV

RossBrunton:
Previously, there was a "legacy" GEMV and an unnamed "experimental"
GEMV. The experimental one didn't work, so it's been removed, now
the old legacy is the new default.


Nyrio:
Thanks for doing this. Looks good to me but there are still a few references to `gemv_legacy` in the code, can you check the following files?:
* `cmake/CmakeFunctionHelper.cmake` line 397
* `src/operations/blas2/gemv_legacy.hpp`: why is this still here?
* `src/operations/blas2_trees.hpp`: line 30
* `src/interface/blas2/CMakeLists.txt`: line 31
* `src/interface/blas2/gemv_legacy.cpp.in`: why is it still named that way?

PR #117, opened 2019-03-28
Change opencl package to reduce rate of asum exceptional failures. 

AdamHarries:
This PR changes the intel opencl package to a more up to date one, in order to avoid failures in the asum test. Additionally, the verbosity of the `make` command is reduced in order to make following the output of the build process easier.


build.sh
```diff
@@ -65,7 +65,7 @@ function configure {
 }
 
 function mak {
-    pushd build && make VERBOSE=1 -j$NPROC
+    pushd build && make -j$NPROC
```
RossBrunton: Can pushd fail? If so, do you want to popd?

Nyrio: Why this change?

mehdi-goli: @Nyrio this is because the length of the log would be more that 10000 lines in travis and then travis cannot shows the logs and any result should be seen in a raw log. That is why we change that to be able to see the travis result.

mehdi-goli: @RossBrunton  this is because the length of the log would be more that 10000 lines in travis and then travis cannot shows the logs and any result should be seen in a raw log. That is why we change that to be able to see the travis result.

RossBrunton: We have safeguards in place for when pushd fails... But if it does, then nothing will be on the stack when popd runs.

(This is nitpicky, and I don't see it as a blocker)

RossBrunton: But that's a different issue, so shouldn't be fixed here.

PR #120, opened 2019-04-09
Benchmark csv runtime config

Nyrio:
CSV runtime configuration of the benchmarks.


benchmark/syclblas/main.cpp
```diff
@@ -13,29 +13,35 @@ void free_device_selector() { cdsp.reset(); }
 ExecutorPtr Global::executorInstancePtr;
```
mehdi-goli: I think now we should be able to remove global variable

Nyrio: Oh yes I forgot about the executor, sorry.

Nyrio: Fixed

benchmark/clblast/blas3/gemm.cpp
```diff
@@ -57,24 +66,25 @@ void BM_Gemm(benchmark::State& state) {
...
+  clblast::Transpose a_tr = blas_benchmark::utils::translate_transposition(t_a);
+  clblast::Transpose b_tr = blas_benchmark::utils::translate_transposition(t_b);
 
   // Specify the layout. As with GEMV, this needs to be kColMajor, and results
   // in errors otherwise. It may be that this is incorrect (especially for
   // performance reasons), so may need to be revisited.
   auto layout = clblast::Layout::kColMajor;
 
-  ExecutorType* ex = Global::executorInstancePtr.get();
+  ExecutorType* ex = executorInstancePtr.get();
```
AdamHarries: Do we still need an executorInstancePtr when we're dynamically constructing the benchmarks? If so, then why is it no longer in a global namespace?

benchmark/clblast/cli_device_selector.hpp
```diff
@@ -38,65 +38,34 @@
 #include <utility>
 #include <vector>
 
+#include "cli_args.hpp"
+
 class cli_device_selector {
   std::string program_name;
 
-  void usage() {
```
AdamHarries: Where has the usage gone? I'm assuming that it's now part of the command line arg parsing code?

benchmark/cli_args.hpp
```diff
@@ -0,0 +1,43 @@
+#ifndef CLI_ARGS_HPP
```
AdamHarries: Could this file be named more descriptively - for example `benchmark_cli_arg_defs.hpp`. When reading other code that includes it, it's not clear whether or not it's just definitions for general CLI argument parsing, or for benchmarks specifically.

benchmark/common_utils.hpp
```diff
@@ -5,15 +5,91 @@
...
 #include <memory>
+#include <sstream>
+#include <string>
+#include <vector>
 
 #include "blas_meta.h"
+#include "cli_args.hpp"
 
 using index_t = int;
+using Blas3Param = std::tuple<std::string, std::string, int, int, int>;
```
AdamHarries: this should be named according to the conventions that we have - i.e. `blas_3_param_t`

benchmark/syclblas/CMakeLists.txt
```diff
@@ -8,15 +8,15 @@ include_directories(${SYCLBLAS_BENCH})
...
-  ${SYCLBLAS_BENCH}/blas2/gemv.cpp
+  # ${SYCLBLAS_BENCH}/blas1/axpy.cpp
+  # ${SYCLBLAS_BENCH}/blas1/asum.cpp
+  # ${SYCLBLAS_BENCH}/blas1/dot.cpp
+  # ${SYCLBLAS_BENCH}/blas1/iamax.cpp
+  # ${SYCLBLAS_BENCH}/blas1/iamin.cpp
+  # ${SYCLBLAS_BENCH}/blas1/nrm2.cpp
+  # ${SYCLBLAS_BENCH}/blas1/scal.cpp
+  # # Level 2 blas
+  # ${SYCLBLAS_BENCH}/blas2/gemv.cpp
```
AdamHarries: commented code

benchmark/syclblas/main.cpp
```diff
@@ -8,40 +8,43 @@
...
   // googlebench, as otherwise we may not be able to delete the queue before we
   // exit (if Initialise calls exit(0)), and dump some information about it
   cl::sycl::queue q = cl::sycl::queue(
       *cdsp.get(), {cl::sycl::property::queue::enable_profiling()});
-  benchmark::utils::print_queue_information(q);
+  blas_benchmark::utils::print_queue_information(q);
 
   // Create a sycl blas executor from the queue
-  Global::executorInstancePtr = std::unique_ptr<SyclExecutorType>(
+  ExecutorPtr executorInstancePtr = std::shared_ptr<SyclExecutorType>(
```
AdamHarries: This doesn't need to be a pointer any more - we can just create an executor, and then pass a pointer to it into blas_benchmark::create_benchmark, as the lifetime of the executor will be longer than the lifetime of the benchmarks.

benchmark/test.csv
```diff
@@ -0,0 +1,33 @@
+n,n,1,2,3
```
AdamHarries: please move this to a folder with ranges. e.g. `benchmark/configurations/test.csv` or `benchmark/ranges/test.csv`

benchmark/CMakeLists.txt
```diff
@@ -53,4 +53,4 @@ endif(SYCL_DEVICE)
 
 include_directories(${BLAS_BENCH})
 add_subdirectory(${BLAS_BENCH}/syclblas)
-add_subdirectory(${BLAS_BENCH}/clblast)
+# add_subdirectory(${BLAS_BENCH}/clblast)
```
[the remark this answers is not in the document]
Nyrio: Sure, this is because I only made the changes for gemm at the moment so the others won't compile

benchmark/clblast/CMakeLists.txt
```diff
@@ -32,15 +32,15 @@ if (CLBLAST_FOUND)
...
-    ${CLBLAST_BENCH}/blas2/gemv.cpp
+    # ${CLBLAST_BENCH}/blas1/asum.cpp
+    # ${CLBLAST_BENCH}/blas1/axpy.cpp
+    # ${CLBLAST_BENCH}/blas1/dot.cpp
+    # ${CLBLAST_BENCH}/blas1/iamax.cpp
+    # ${CLBLAST_BENCH}/blas1/iamin.cpp
+    # ${CLBLAST_BENCH}/blas1/nrm2.cpp
+    # ${CLBLAST_BENCH}/blas1/scal.cpp
+    # # Level 2 blas
+    # ${CLBLAST_BENCH}/blas2/gemv.cpp
```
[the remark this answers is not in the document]
Nyrio: Temporary. I will made the same changes for these files once we have reviewed them for gemm, at the moment they won't compile

benchmark/argparse.hpp
```diff
@@ -0,0 +1,495 @@
+/* Copyright (c) 2017, Hilton Bristow
```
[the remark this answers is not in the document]
Nyrio: If we fork-and-submodule an argument parser, maybe we should find a better one.This one does the job here but it's quite limited and I had to hack it a bit to use it with google benchmark (which has its own args)

benchmark/clblast/blas3/gemm.cpp
Nyrio: No, I should pass it by reference. I will make the change

benchmark/clblast/cli_device_selector.hpp
Nyrio: Yes, it is handled by the arg parser

benchmark/test.csv
Nyrio: Actually I added this file by mistake, I will remove it

benchmark/clblast/blas3/gemm.cpp
Nyrio: Done

benchmark/cli_args.hpp
Nyrio: Done

benchmark/syclblas/main.cpp
Nyrio: Done

benchmark/test.csv
Nyrio: Done

benchmark/common_utils.hpp
Nyrio: Done

benchmark/argparse.hpp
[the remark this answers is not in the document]
mehdi-goli: @Nyrio you should use clara not this one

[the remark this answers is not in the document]
Nyrio: @mehdi-goli done. Had to hack it to work with googlebench but clara is definitely better.

mehdi-goli:
Please add a readme file explaining the benchmarks? how to use it? CSVfiles format, how to generate CSV using python

Nyrio:
I made the changes for all benchmarks now, and added a README to the benchmark folder with documentation about how to use the benchmarks.

benchmark/common_utils.hpp
```diff
@@ -5,15 +5,147 @@
...
+    constexpr int dmin = 64, dmax = 1024;
+    for (std::string t : {"n", "t"})
+      for (int m = dmin; m <= dmax; m *= 2)
+        for (int n = dmin; n <= dmax; n *= 2)
+          blas2_default.push_back(std::make_tuple(t, m, n));
+    return blas2_default;
+  } else
+    return parse_csv_file<blas2_param_t>(
+        args.csv_param, [&](std::vector<std::string>& v) {
+          return std::make_tuple(v[0].c_str(), std::stoi(v[1]),
```
RossBrunton: How will this work if only 0 or 1 values were parsed from the csv?

benchmark/common_utils.hpp
```diff
@@ -62,10 +194,10 @@ const std::array<Transposition, 3> possible_transpositions(
  * @fn to_transpose_enum
  * @brief Translates from a transposition string to an enum.
  */
-static inline Transposition to_transpose_enum(const char* t) {
-  if (t[0] == 't') {
+static inline Transposition to_transpose_enum(std::string t) {
+  if (t == "t") {
```
RossBrunton: Enum values can be upper case as well.

benchmark/clara.hpp
```diff
@@ -0,0 +1,1260 @@
+// Copyright 2017 Two Blue Cubes Ltd. All rights reserved.
```
mehdi-goli: Move this to the external folder create folder called clara

benchmark/clblast/cli_device_selector.hpp
```diff
@@ -38,65 +38,34 @@
 #include <utility>
```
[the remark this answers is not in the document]
Nyrio: As discussed offline, there is the issue that I have two different device selectors in the `syclblas` and  `clblast` folders, with only 20 lines in common

benchmark/common_utils.hpp
Nyrio: I think we discussed about checking the CSV format or not. It adds a lot of code to check it properly and provide relevant information about the failure, but what we can do, maybe, is to catch the exceptions, print a message, and then either fail or use defaults. Any opinion on this?

benchmark/common_utils.hpp
Nyrio: What do you mean? This string is initially read from the CSV file (or from defaults) and is supposed to be in lower case.

benchmark/clblast/blas2/gemv.cpp
```diff
@@ -54,98 +61,83 @@ void BM_Gemv(benchmark::State& state) {
...
-  } while (!gemv_range.finished());
+template <typename scalar_t>
+void register_benchmark(blas_benchmark::Args& args, ExecutorType* exPtr) {
+  auto gemm_params = blas_benchmark::utils::get_params<blas2_param_t>(args);
+
+  for (auto p : gemm_params) {
+    std::string ts;
+    int m, n;
+    std::tie(ts, m, n) = p;
+    int t = (int)blas_benchmark::utils::to_transpose_enum(ts);
```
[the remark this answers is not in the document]
Nyrio: Fixed, thanks for catching that.

benchmark/common_utils.hpp
Nyrio: I prefer to fix a convention and always use the same case, because I don't want a file to look like:
```
n,n,64,64,64
n,N,64,128,64
T,n,64,256,64
...
```

benchmark/clara.hpp
Nyrio: Done

benchmark/common_utils.hpp
```diff
@@ -5,15 +5,147 @@
...
+
+template <>
+inline std::vector<blas1_param_t> get_params<blas1_param_t>(Args& args) {
+  if (args.csv_param.empty()) {
+    warning_no_csv();
+    std::vector<blas1_param_t> blas1_default;
+    for (int size = 4096; size <= 1048576; size *= 2)
+      blas1_default.push_back(size);
+    return blas1_default;
+  } else
```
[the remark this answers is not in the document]
Nyrio: Done

benchmark/common_utils.hpp
RossBrunton: Whether or not you generate a csv file like that, unless it's an entirely closed system, a user may write or generate a file with all upper case. Somewhere down the line someone is going to use "T" and not realise it means "n" instead.

benchmark/common_utils.hpp
RossBrunton: I'm not really bothered about having a perfectly crafted error message. Perhaps parse_csv_file or the param_t could have the number of fields per record, and it just says "invalid csv file!" if they don't match. I'd rather not have an invalid csv file causing UB.

benchmark/common_utils.hpp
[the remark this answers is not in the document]
RossBrunton: It doesn't look like you have? After the else, I mean.

benchmark/common_utils.hpp
Nyrio: It doesn't do that anymore. Everything that is not `n`, `t` or `c` now generates the error: `Unrecognized transpose type: x`

Nyrio: You're right that using `n` as default is not a good idea, I fixed that

benchmark/common_utils.hpp
Nyrio: I have added some basic checking with 3 possible messages (+ printing the line where it fails):
* `invalid number of parameters (x expected)`
* `invalid parameter`
* `unrecognized transpose type: x`

benchmark/common_utils.hpp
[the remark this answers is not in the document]
Nyrio: Oh right, I thought you were talking about the `for` loop. Fixed now

benchmark/clblast/cli_device_selector.hpp
```diff
@@ -38,65 +38,34 @@
...
-              << std::endl;
-    std::cout << " --device  DEVICE" << std::endl
-              << "    Select a device (best effort) for running the benchmark."
-              << std::endl;
-    std::cout << "    e.g. intel:cpu, amd:gpu etc" << std::endl;
-  }
-
  public:
-  std::string device_vendor;
+  std::string vendor_name;
```
AdamHarries: Is there any reason why this was changed? I somewhat prefer the whole device_<...> naming scheme, but if there's a reason to change it, then do so.

benchmark/syclblas/main.cpp
```diff
@@ -8,40 +8,35 @@
 std::unique_ptr<cli_device_selector> cdsp;
 void free_device_selector() { cdsp.reset(); }
 
-// Declare the executorInstancePtr (it is already `extern` declared in utils.hpp
-// but it must be explicitly declared at least once.)
```
AdamHarries: declared *exactly* once

AdamHarries: otherwise there'll be linker errors

benchmark/clblast/cli_device_selector.hpp
Nyrio: To use the same names in both CLBLAST and SYCL-BLAS device selectors. I can revert that if you prefer though

benchmark/syclblas/main.cpp
Nyrio: This comment (and this variable) have been deleted, and there is now a local variable called `executor` instead. Anything you wish to modify here?

Nyrio: This comment (and this variable) have been deleted, and there is now a local variable called executor instead. Anything you wish to modify here?

PR #121, opened 2019-04-09
Changed GEMM test logic

RossBrunton:
It now works similarly to BLAS2 GEMV tests in that it uses a gtest
TEST_P test thing. This also covers different transposation values,
so they are no longer in seperate files.

In addition the batched/nonbatched versions are split up into
seperate files.

This also required changing assert_t_equal to a macro so it fails
on the first assert, rather than printing all of them.


test/blas_test.hpp
```diff
@@ -204,19 +204,8 @@ class BLAS_Test<blas_test_args<ScalarT_, MetadataT_, ExecutorType_>>
...
-void assert_t_equal<float>(float a, float b) {
-  ASSERT_FLOAT_EQ(a, b);
-}
-
-template <>
-void assert_t_equal<double>(double a, double b) {
-  ASSERT_DOUBLE_EQ(a, b);
-}
+#define ASSERT_T_EQUAL(T, val1, val2)                                         \
+  ASSERT_PRED_FORMAT2(::testing::internal::CmpHelperFloatingPointEQ<T>, val1, \
```
mehdi-goli: You can directly put that in the test instead of wrapping it in a macro.

test/unittest/blas3/blas3_gemm_test.cpp
```diff
@@ -0,0 +1,141 @@
...
+  std::array<int, 2> dim_b = {k, n};
+  std::array<int, 2> dim_c = {m, n};
+
+  int lda = ((transa != 'n') ? dim_a[1] : dim_a[0]) * lda_mul;
+  int ldb = ((transb != 'n') ? dim_b[1] : dim_b[0]) * ldb_mul;
+  int ldc = dim_c[0] * ldc_mul;
+
+  std::vector<scalar_t> a_m(_size(dim_a, lda_mul));
+  std::vector<scalar_t> b_m(_size(dim_b, ldb_mul));
+  std::vector<scalar_t> c_m_gpu(_size(dim_c, ldc_mul), scalar_t(10.0));
```
mehdi-goli: you can use std::rand here

test/unittest/blas3/blas3_gemm_test.cpp
```diff
@@ -0,0 +1,141 @@
...
+  std::vector<scalar_t> c_m_cpu(_size(dim_c, ldc_mul), scalar_t(10.0));
```
mehdi-goli: abd you can set this one to whatever c_m_gpu computed

test/blas_test.hpp
RossBrunton: My reasoning for having it a macro is that it allows the code to be less verbose and more clear, and also updatable if a different method is to be called.

"ASSERT_T_EQUAL" is clearer than "ASSERT_PRED_FORMAT2".

PR #122, opened 2019-04-18
Fix for TRMV of a unit triangle

RossBrunton:
It was reading from the matrix, rather than the vector for the
diagonal value.


mehdi-goli:
looks fine to me

PR #124, opened 2019-04-18
Added/updated tests for BLAS2 functions

RossBrunton:
This change adds an (updated) test for GER, SYMV, SYR, SYR2, and
TRMV.

As part of this, the reference blas system was rewritten and now
uses the proper cblas.h header.

Please note: This requires the fix for unit triangles in TRMV.

### DIFF
--- a/external/cblas/cblas.h
+++ b/external/cblas/cblas.h
@@ -1,0 +1,577 @@
+/* CBlas header, downloaded from http://www.netlib.org/blas/#_cblas */
+
+#ifndef CBLAS_H
+#define CBLAS_H
+#include <stddef.h>
+
+/*
+ * Enumerated and derived types
+ */
+#define CBLAS_INDEX size_t  /* this may vary between platforms */
+
+enum CBLAS_ORDER {CblasRowMajor=101, CblasColMajor=102};
+enum CBLAS_TRANSPOSE {CblasNoTrans=111, CblasTrans=112, CblasConjTrans=113};
+enum CBLAS_UPLO {CblasUpper=121, CblasLower=122};
+enum CBLAS_DIAG {CblasNonUnit=131, CblasUnit=132};
+enum CBLAS_SIDE {CblasLeft=141, CblasRight=142};
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/*
+ * ===========================================================================
+ * Prototypes for level 1 BLAS functions (complex are recast as routines)
+ * ===========================================================================
+ */
+float  cblas_sdsdot(const int N, const float alpha, const float *X,
+                    const int incX, const float *Y, const int incY);
+double cblas_dsdot(const int N, const float *X, const int incX, const float *Y,
+                   const int incY);
+float  cblas_sdot(const int N, const float  *X, const int incX,
+                  const float  *Y, const int incY);
+double cblas_ddot(const int N, const double *X, const int incX,
+                  const double *Y, const int incY);
+
+/*
+ * Functions having prefixes Z and C only
+ */
+void   cblas_cdotu_sub(const int N, const void *X, const int incX,
+                       const void *Y, const int incY, void *dotu);
+void   cblas_cdotc_sub(const int N, const void *X, const int incX,
+                       const void *Y, const int incY, void *dotc);
+
+void   cblas_zdotu_sub(const int N, const void *X, const int incX,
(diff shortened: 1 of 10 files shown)


test/cblas.h
```diff
@@ -0,0 +1,575 @@
+#ifndef CBLAS_H
+#define CBLAS_H
```
mehdi-goli: Move this file to the external project clblas so as it is a third-party project and it is not part of sycl-blas

external/cblas/cblas.h
```diff
@@ -0,0 +1,577 @@
+/* CBlas header, downloaded from http://www.netlib.org/blas/#_cblas */
+
```
mehdi-goli: You may not need cblas.h . I think all the blas libraries handles c++ demangling. So you may not need the extern C. you should be to directly call sgemv_ or dgemv_ , see the below example

test/system_reference_blas.hpp
```diff
@@ -26,56 +26,140 @@
...
+  (f_or_d<scalar_t, decltype(F), decltype(D)>::get(F, D))
+
+// =======
+// Level 2
+// =======
+template <typename scalar_t>
+void gemv(const char *trans, int m, int n, scalar_t alpha, const scalar_t a[],
+          int lda, const scalar_t b[], int incX, scalar_t beta, scalar_t c[],
+          int incY) {
+  auto f = COROUTINE_SELECT(&cblas_sgemv, &cblas_dgemv);
```
mehdi-goli: this will be COROUTINE_SELECT(sgemv_ , dgemv_)

test/system_reference_blas.hpp
```diff
@@ -26,56 +26,140 @@
...
+    case 'N':
+      return CblasNonUnit;
+  }
+}
+}  // namespace
+
+namespace reference_blas {
+
+template <typename scalar_t, typename float_t, typename double_t>
+struct f_or_d {};
```
mehdi-goli: You can use c++ variadic template with function pointer and its parameters as a template function input to call:
template<typename func, typename ... params> func(func f, params... prs) {
f(pr...);
}

external/cblas/cblas.h
RossBrunton: I don't follow what you are saying here, sorry!

test/system_reference_blas.hpp
RossBrunton: I'm not that familiar with varadic templates, I'm afraid. But is there a way to get the nth parameter of an argument and template bundle? There seems to be a way using std::tuple?

I'm not what one would gain from using varadic templates in this case either.

mehdi-goli: it is more for maintainability. As the next step is adding conjugate version which requires two more types to be added

external/cblas/cblas.h
RossBrunton: Done!

test/system_reference_blas.hpp
RossBrunton: I've been staring at this a while now, and I can't for the life of me figure out how I would use variadic templates for this.

mehdi-goli: It can be something like this
template<typename type, typename func_t1, typename funct_2> struct Func;
template<typename FuncT1, Funct2> struct Func<float, FunctT1, FuncT2> struct Func{

template <typename Ret, typename ... Args> Ret cal_func(Args .... args) {
return FuncT1(args...); 
}
};
// similarly for double 
in your code you call Func<datatype, sxxx, dxxx>::calfunc(parameters);

test/system_reference_blas.hpp
```diff
@@ -26,56 +26,140 @@
...
+
+template <>
+struct TypeDispatcher<float> {
+  template <typename ret_t = void, typename floatfn_t, typename doublefn_t,
+            typename... Args>
+  static ret_t call(floatfn_t ffn, doublefn_t dfn, Args... args) {
+    return ffn(args...);
+  }
+};
+template <>
```
mehdi-goli: This should wrap in #if DOUBLE_SUPPORT

mehdi-goli:
should be rebased to use the CBLAS submodule.

RossBrunton:
Looks like the blas2_trmv_test is failing for some reason.

RossBrunton:
Hmm... Had this issue before actually. I thought I checked it, but maybe it statically links with it.

PR #126, opened 2019-04-24
Set the initial values for iamax and iamin correctly

RossBrunton:
This change changes the base value for the maximum value to (0,
0.0). If the first element is 0.0, this is the output, if it is
greater than 0.0, then it will be treated as larger and get set.


RossBrunton:
This actually doesn't work at all, even for 0.0. It was working earlier. :C

PR #129, opened 2019-04-25
WIP: Extend BufferIterator

Rbiessy:
These are changes required by SYCL-DNN in order to be able to use the BufferIterator instead of the PointerMapper. The BufferIterator needs to behave like a pointer so we need to have:
* a default constructor
* a copy constructor able to construct `BufferIterator<const T>` from `BufferIterator<T>`
* `BufferIterator<T>` implicitly convertible to `BufferIterator<const T>`


mehdi-goli:
If you want to handle it like a pointer, you need to handle the nullptr behaviour as well, as in SYCLBLAS buffer size of 1 has a meaning since dot product and ... will return buffer size of one as an output. So in this case you need to distinguish between buffer_Iterator with a default constructor and buffer iterator with size 1.

PR #130, opened 2019-04-25
Unit tests now take in a --device parameter

RossBrunton:
This takes a colon seperated pair thing, which specifies the device
to run on. This also moves the CLI selector to a common location.


test/blas_test.hpp
```diff
@@ -204,6 +211,13 @@ class BLAS_Test<blas_test_args<ScalarT_, MetadataT_, ExecutorType_>>
   }
 };
 
+template<typename scalar_t>
+cl::sycl::queue make_queue() {
```
mehdi-goli: make queue should not take type. Why do you need to pass type to make queue?

test/unittest/main.cpp
```diff
@@ -25,11 +25,46 @@
...
-int main(int argc, char *argv[]) {
+struct Args args{};
+
+namespace {
+/**
+ * @fn parse_args
+ * @brief Returns a structure containing the data extracted from the
+ * command-line arguments.
+ */
+void parse_args(int argc, char** argv) {
```
mehdi-goli: If a user does not pass a device type, the default selector should be picked

test/unittest/main.cpp
```diff
@@ -25,11 +25,46 @@
...
+  bool show_help;
+
+  auto parser = clara::Help(show_help) |
+                clara::Opt(args.device, "device")["--device"](
+                    "Select a device (best effort) for running the benchmark.");
+
+  auto res = parser.parse(clara::Args(argc, argv));
+  if (!res) {
+    std::cerr << "Error in command line: " << res.errorMessage() << std::endl;
+    exit(1);
```
mehdi-goli: If a user passes a device that does not exist you should give a warning saying that any device based on the order in your clinfo will be selected, you dont need to exit

test/blas_test.hpp
RossBrunton: We need to create a BLAS_Test instance of the appropriate type. Unless we don't want to use the BLAS_Test thing anymore? Which is probably the correct move here.

test/unittest/main.cpp
RossBrunton: That's not what this code is doing, this code only checks that the --device parameter is valid (i.e. followed by an argument on the argument list).

In terms of legality checking, I think if the selector doesn't match exactly, the program should fail.

Nyrio: The way device selectors are working, the program shouldn't fail. An invalid name will just result in a random device being used (best effort but with irrelevant search criteria)

test/blas_test.hpp
```diff
@@ -204,6 +211,18 @@ class BLAS_Test<blas_test_args<ScalarT_, MetadataT_, ExecutorType_>>
...
+  using type_t = blas_test_args<scalar_t, void>;
+  using blas_test_t = BLAS_Test<type_t>;
+  if (args.device.empty()) {
+    std::cout << "Using default selector\n";
```
RossBrunton: ... Yeah, I should remove this. Good old VSCode's git thing.

test/unittest/main.cpp
RossBrunton: Should be done now, both the SYCL benchmarks and the unit tests now will use the default selector if args.device is empty.

mehdi-goli:
Printing the selected device type is useful

mehdi-goli:
CLBLAST and systemBLAS should not be dependent on SYCL

test/unittest/CMakeLists.txt
```diff
@@ -28,7 +28,7 @@ project(syclblas_unittest)
 
 set(SYCLBLAS_UNITTEST ${CMAKE_CURRENT_SOURCE_DIR})
 
-include_directories(${SYCLBLAS_TEST} ${BLAS_INCLUDE_DIRS})
+include_directories(${SYCLBLAS_TEST} ${BLAS_INCLUDE_DIRS} "${SYCLBLAS_TEST}/../external/clara")
```
RossBrunton: What is this doing here?

Nyrio: You're the author, are you asking this question to yourself?

RossBrunton: Just making a note to myself.

PR #131, opened 2019-04-25
Updated BLAS1 tests to match BLAS2/3 tests

RossBrunton:
This uses a parametized test now, and compares their values to
those produced by the reference blas (rather than hardcoding the
algorithm).

### DIFF
--- a/test/blas_test.hpp
+++ b/test/blas_test.hpp
@@ -33,6 +33,7 @@
 #include <iomanip>
 #include <iostream>
 #include <vector>
+#include <random>
 
 #include <gtest/gtest.h>
 
@@ -42,6 +43,7 @@
 #include "utils/system_reference_blas.hpp"
 #include "utils/cli_device_selector.hpp"
 #include "utils/print_queue_information.hpp"
+#include "utils/float_comparison.hpp"
 
 struct Args {
   std::string program_name;
@@ -240,8 +242,19 @@ inline cl::sycl::queue make_queue() {
   return q;
 };
 
-#define ASSERT_T_EQUAL(T, val1, val2)                                         \
-  ASSERT_PRED_FORMAT2(::testing::internal::CmpHelperFloatingPointEQ<T>, val1, \
-                      val2)
+/**
+ * @fn random_data
+ * @brief Generates a random vector of scalar values, using a uniform
+ * distribution.
+ */
+template <typename scalar_t>
+static inline void fill_random(std::vector<scalar_t> &vec) {
+  std::random_device rd;
+  std::mt19937 gen(rd());
+  std::uniform_real_distribution<> dis(-1.0, 1.0);
+  for (scalar_t &e : vec) {
+    e = dis(gen);
+  }
+}
 
 #endif /* end of include guard: BLAS_TEST_HPP */
--- a/test/unittest/blas3/blas3_gemm_test.cpp
+++ b/test/unittest/blas3/blas3_gemm_test.cpp
@@ -81,9 +81,9 @@ void run_test(const combination_t<scalar_t> combi) {
   std::vector<scalar_t> c_m_gpu(_size(dim_c, ldc_mul));
   std::vector<scalar_t> c_m_cpu(_size(dim_c, ldc_mul));
 
-  blas_test_t::set_rand(a_m, _size(dim_a, lda_mul));
-  blas_test_t::set_rand(b_m, _size(dim_b, ldb_mul));
-  blas_test_t::set_rand(c_m_gpu, _size(dim_c, ldc_mul));
+  fill_random(a_m);
+  fill_random(b_m);
+  fill_random(c_m_gpu);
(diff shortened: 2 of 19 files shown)


RossBrunton:
Locally the rotg test seems to be failing. Lets see what Travis says about that.

mehdi-goli:
Require to fix the precision issue for Intel:GPU

test/blas_test.hpp
```diff
@@ -204,8 +204,13 @@ class BLAS_Test<blas_test_args<ScalarT_, MetadataT_, ExecutorType_>>
   }
 };
 
-#define ASSERT_T_EQUAL(T, val1, val2)                                         \
-  ASSERT_PRED_FORMAT2(::testing::internal::CmpHelperFloatingPointEQ<T>, val1, \
-                      val2)
+constexpr const int ulp_required_in_assert = 20;
+
+#define ASSERT_T_EQUAL(T, val1, val2)                                   \
```
mehdi-goli: create a function called precision Computation and return the precision. So in this case you dont need macro for asser_near and you can directly put it in the code

test/blas_test.hpp
```diff
@@ -204,8 +204,13 @@ class BLAS_Test<blas_test_args<ScalarT_, MetadataT_, ExecutorType_>>
...
+  do {                                                                  \
+    T precision = std::numeric_limits<T>::epsilon() *                   \
+                  std::abs((val1) + (val2)) * ::ulp_required_in_assert; \
```
Nyrio: Shouldn't it be `fabs`?

Nyrio: Also if you can put a link or explanation of where that 20 comes from that would be nice

RossBrunton: std::abs and std::fabs do the same thing if cmath is imported. Because C++.

test/blas_test.hpp
RossBrunton: Should be done now.

test/blas_test.hpp
Nyrio: Are you absolutely sure about this? I don't see floating point versions there: https://en.cppreference.com/w/cpp/numeric/math/abs

mehdi-goli:
Can you please the std::uniform distribution for data as used in the benchmark. Also, can you use the new floating point precision validation algorithm for test

Nyrio:
@RossBrunton this is the code used for the benchmarks:
```cpp
#include <random>
...

/**
 * @fn random_data
 * @brief Generates a random vector of scalar values, using a uniform
 * distribution.
 */
template <typename scalar_t>
static inline std::vector<scalar_t> random_data(size_t size,
                                                bool initialized = true) {
  std::vector<scalar_t> v = std::vector<scalar_t>(size);

  std::random_device rd;
  std::mt19937 gen(rd());
  std::uniform_real_distribution<> dis(-2.0, 2.0);
  for (int i = 0; i < v.size(); ++i) {
    v[i] = dis(gen);
  }
  return v;
}
```

test/blas_test.hpp
```diff
@@ -204,8 +204,14 @@ class BLAS_Test<blas_test_args<ScalarT_, MetadataT_, ExecutorType_>>
   }
 };
 
-#define ASSERT_T_EQUAL(T, val1, val2)                                         \
-  ASSERT_PRED_FORMAT2(::testing::internal::CmpHelperFloatingPointEQ<T>, val1, \
-                      val2)
+// The "distance" required between two floats for precision_limit
+// Higher numbers = less precision, 40 was obtained through experimentation
+// This results in values in xe-5 range for floats, and xe-14 for doubles
+constexpr const int ulp_required_for_precision = 40;
```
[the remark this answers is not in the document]
RossBrunton: This is no longer relevant.

test/unittest/blas3/blas3_gemm_test.cpp
```diff
@@ -116,15 +116,14 @@ void run_test(const combination_t<scalar_t> combi) {
...
+        policy_handler.copy_to_host(m_c_gpu + _base(dim_c, ldc_mul, bs),
+                                    c_m_gpu.data() + _base(dim_c, ldc_mul, bs),
+                                    _size_batch(dim_c, ldc_mul));
     policy_handler.wait(event);
   }
 
-  for (int i = 0; i < _size(dim_c, ldc_mul); ++i) {
-    ASSERT_T_EQUAL(scalar_t, c_m_gpu[i], c_m_cpu[i]);
-  }
+  ASSERT_TRUE(utils::compare_vectors(c_m_gpu, c_m_cpu));
```
mehdi-goli: You can call this ASSERT_NEAR or ASSERT_ALMOST_EQUAL

RossBrunton: What do you mean? Do you want me to get the threshold and use ASSERT_NEAR on it?

Nyrio: We discussed it offline, you can ignore that

PR #132, opened 2019-04-29
Patch: use index_t instead of hardcoded int for matrix and vector dimensions

Nyrio:


Nyrio:
> Overall looks good. Remove the SDK update from this MR

Fixed

PR #133, opened 2019-05-01
Created an option to compile clblast benchmarks and changed import method

Nyrio:


benchmark/clblast/CMakeLists.txt
```diff
@@ -18,48 +18,38 @@ else()
...
-
-    message("-- Created benchmark: ${clblast_bench_exec}")
-    install(TARGETS bench_clblast_${clblast_bench_exec} RUNTIME DESTINATION sycl_blas/bin)
-  endforeach()
-else()
-  message(STATUS "CLBlast not found, only building benchmarks for syclblas")
-endif()
+find_package(CLBlast REQUIRED)
+
+message(STATUS "CLBlast found!")
```
mmha: `find_package` prints out a message already.

Nyrio: Uh actually no

Nyrio:
```
-- CLBlast root not specified by user, searching default path /usr/lib/clblast
-- Created benchmark: asum
-- Created benchmark: axpy
...
```

Nyrio: It only prints something if it fails

Nyrio: Done

benchmark/clblast/CMakeLists.txt
```diff
@@ -1,5 +1,7 @@
 cmake_minimum_required(VERSION 3.2.2)
 
+include(FindCLBlast)
```
mmha: Find Module should not be included. Just `find_package` them

Nyrio: Fixed

benchmark/clblast/utils.hpp
```diff
@@ -13,6 +13,35 @@ void create_benchmark(blas_benchmark::Args &args, ExecutorType *exPtr);
 
 namespace utils {
 
+inline void print_device_information(cl_device_id device) {
+  char device_name[64];
+  cl_device_type *device_type = new cl_device_type;
```
mmha: Make this a `std::unique_ptr`.

mmha: In fact, why is this on the heap, even?

benchmark/clblast/utils.hpp
```diff
@@ -13,6 +13,35 @@ void create_benchmark(blas_benchmark::Args &args, ExecutorType *exPtr);
...
+  clGetDeviceInfo(device, CL_DEVICE_NAME, 64 * sizeof(char), device_name, NULL);
```
mmha: - make this `nullptr`, please
- There is some code duplication here: The size of the array you are writing to is stated here and at the point of declaration above. It's better to use  `sizeof(device_name)`. It would be even better to use `std::string` here because the length of the string might exceed 64 bytes:

```c++
size_t device_name_length = 0;
clGetDeviceInfo(device, CL_DEVICE_NAME, 0, nullptr, &device_name_length);
if(!device_name_length) {
    return;
}
std::string device_name(device_name_length);
clGetDeviceInfo(device, CL_DEVICE_NAME, 0, std::addressof(device_name.front()), nullptr);
```

cmake/Modules/FindSB_CLBlast.cmake
```diff
@@ -0,0 +1,40 @@
...
+# *  @filename FindCLBlast.cmake
+# *
+# **************************************************************************/
+
+find_package(OpenCL)
+find_path(SB_CLBLAST_INCLUDE_DIR clblast.h HINTS ${CLBLAST_ROOT} PATH_SUFFIXES include)
+find_library(SB_CLBLAST_LIBRARY NAME clblast libclblast.so HINTS ${CLBLAST_ROOT} PATH_SUFFIXES lib)
+
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(MYCLBlast REQUIRED_VARS SB_CLBLAST_LIBRARY SB_CLBLAST_INCLUDE_DIR OpenCL_FOUND)
```
mmha: Should be `SB_CLBlast`.

cmake/Modules/FindSB_CLBlast.cmake
```diff
@@ -0,0 +1,40 @@
...
+
+if(MYCLBlast_FOUND AND NOT TARGET clblast)
```
mmha: Same here.

cmake/Modules/FindSB_CLBlast.cmake
```diff
@@ -0,0 +1,40 @@
...
+# *
+# *  SYCL-BLAS: BLAS implementation using SYCL
+# *
+# *  @filename FindCLBlast.cmake
+# *
+# **************************************************************************/
+
+find_package(OpenCL)
+find_path(SB_CLBLAST_INCLUDE_DIR clblast.h HINTS ${CLBLAST_ROOT} PATH_SUFFIXES include)
+find_library(SB_CLBLAST_LIBRARY NAME clblast libclblast.so HINTS ${CLBLAST_ROOT} PATH_SUFFIXES lib)
```
mmha: `libclblast.so` is searched for by CMake already.

benchmark/clblast/CMakeLists.txt
```diff
@@ -1,65 +1,40 @@
...
+  # Level 2 blas
+  ${CLBLAST_BENCH}/blas2/gemv.cpp
+  # Level 3 blas
+  ${CLBLAST_BENCH}/blas3/gemm.cpp
+)
+
+# Add individual benchmarks for each method
+foreach(clblast_benchmark ${CLBLAST_BENCH_SRCS})
+  get_filename_component(clblast_bench_exec ${clblast_benchmark} NAME_WE)
+  set(clblast_bench_exec, ${clblast_benchmark})
```
mmha: Remove the `.`

benchmark/clblast/utils.hpp
Nyrio: It shouldn't. Fixed

benchmark/clblast/utils.hpp
Nyrio: Thanks. Your code doesn't work but I've used a slight variation of it.

cmake/Modules/FindSB_CLBlast.cmake
Nyrio: Thanks for catching that

cmake/Modules/FindSB_CLBlast.cmake
Nyrio: Right, fixed

benchmark/clblast/CMakeLists.txt
Nyrio: I don't understand your comment. Which `.`?

cmake/Modules/FindSB_CLBlast.cmake
Nyrio: Right, fixed

benchmark/clblast/CMakeLists.txt
Nyrio: Do you mean the `,`?

mmha: Oh I made a typo. Yes, the comma.

CMakeLists.txt
```diff
@@ -90,6 +90,7 @@ if(${BLAS_ENABLE_TESTING})
 endif()
 
 option(BLAS_ENABLE_BENCHMARK "Whether to enable benchmarking" ON)
+option(BUILD_CLBLAST_BENCHMARKS "Whether to build CLBLAST benchmarks" OFF)
```
AdamHarries: For consistency, could this be `CLBLAST_ENABLE_BENCHMARKS` instead? Just a minor quibble.

Nyrio: I didn't want it to be confusing given that `BLAS_ENABLE_BENCHMARK` acts at an higher level than `BUILD_CLBLAST_BENCHMARKS` (i.e if the value of the former is `OFF`, the latter will be ignored)

PR #136, opened 2019-05-06
Implicit conversion

mehdi-goli:
This PR allows implicit conversion of non-const buffer to be a Buffer of const elements required by SYCL-DNN.


src/container/sycl_iterator.hpp
```diff
@@ -90,6 +93,13 @@ BufferIterator<element_t, codeplay_policy>::operator++(int i) {
   return temp_iterator;
 }
 
+// implicitly converting non-const buffer to const buffer
+template <typename element_t>
+inline BufferIterator<element_t, codeplay_policy>::
+operator BufferIterator<const element_t, codeplay_policy>() const {
+  return BufferIterator<const element_t, codeplay_policy>(*this);
+}
```
cjdb: I think this should be a constructor instead of a conversion operator. You may need to SFINAE out the case when `element_t` is `const` to avoid overload resolution ambiguity.

include/container/sycl_iterator.h
```diff
@@ -67,9 +67,19 @@ class BufferIterator<element_t, codeplay_policy> {
...
+      class = typename std::enable_if<
+          !std::is_const<other_scalar_t>::value ||
+          (std::is_const<other_scalar_t>::value && std::is_const<U>::value &&
+           !std::is_same<U, other_scalar_t>::value)>::type>
   BufferIterator(const BufferIterator<other_scalar_t, codeplay_policy>& other);
 
+  /*!
+   * @brief implicitly converting non-const buffer to const buffer.
+   */
+  // operator BufferIterator<const element_t, codeplay_policy>() const;
```
jwlawson: Probably don't want commented code

PR #140, opened 2019-05-09
Moved system_reference_blas.hpp to under utils

RossBrunton:


Nyrio:
This is the pre-#131 version, right? Should be updated when #131 is merged then.

PR #141, opened 2019-05-09
Floating-point numbers comparison util for tests and benchmarks

Nyrio:


mehdi-goli:
LGTM!

include/utils/float_comparison.hpp
```diff
@@ -0,0 +1,109 @@
...
+  return static_cast<scalar_t>(0.005);
+}
+
+template <>
+inline double getRelativeErrorMargin<double>() {
+  return 0.000001;
+}
+
+/**
+ * Indicates the tolerated margin for absolute differences
```
mehdi-goli: Put a comment and explain how did you came up with these numbers?

Nyrio: I've added a few comments

Nyrio:
@mehdi-goli it's only for floating-point numbers. For integers an exact match can be expected (e.g for iamin and iamax I check that the index is the exact same)

include/utils/float_comparison.hpp
Nyrio: I have added even more explanations. I hope it's clear.
Also I've made a change to properly handle the case of zero

PR #143, opened 2019-05-15
Provide iamin reference implementation

Nyrio:
iamin is an extension to blas and it is not included in `libopenblas.so` when cross-compiling for ARM. Therefore we need to provide our own reference implementation.


include/utils/system_reference_blas.hpp
```diff
@@ -28,13 +28,27 @@
...
+#include <cmath>
 
-extern "C" {
-// i*amin is an extension
-CBLAS_INDEX cblas_isamin(const int N, const float *X, const int incX);
-CBLAS_INDEX cblas_idamin(const int N, const double *X, const int incX);
-CBLAS_INDEX cblas_icamin(const int N, const void *X, const int incX);
-CBLAS_INDEX cblas_izamin(const int N, const void *X, const int incX);
+// i*amin is an extension, provide an implementation
+template<typename scalar_t>
```
RossBrunton: Can this be in an anonymous namespace if it's not public?

include/utils/system_reference_blas.hpp
```diff
@@ -28,13 +28,27 @@
...
+inline int iamin(const int N, const scalar_t *X, const int incX)
+{
+  int best = 0;
+  for(int i = 1; i < N; i++) {
```
RossBrunton: You aren't handling incX.

Nyrio: Oh, my bad. Fixed

include/utils/system_reference_blas.hpp
Nyrio: Done

include/utils/system_reference_blas.hpp
```diff
@@ -79,6 +72,27 @@ CBLAS_DIAG c_diag(char x) {
...
 }
+
+// i*amin is an extension, provide an implementation
+template<typename scalar_t>
+inline int iamin(const int N, const scalar_t *X, const int incX)
+{
+  int best = 0;
+  for(int i = 1; i < N; i += incX) {
+    if(std::abs(X[i]) < std::abs(X[best])) {
+      best = i;
```
RossBrunton: This should return i / incX

Nyrio: Thanks for catching that @RossBrunton , fixed

RossBrunton: lgtm

RossBrunton:
I see no isssues.

PR #145, opened 2019-05-15
Adding gemm batched benchmark

Nyrio:


benchmark/clblast/CMakeLists.txt
```diff
@@ -29,10 +35,9 @@ foreach(clblast_benchmark ${CLBLAST_BENCH_SRCS})
   add_executable(bench_clblast_${clblast_bench_exec} ${clblast_benchmark} ${CLBLAST_BENCH}/main.cpp ${CLBLAST_BENCH}/clwrap.cpp)
   set_target_properties(bench_clblast_${clblast_bench_exec} PROPERTIES CXX_STANDARD 11)
   add_dependencies(bench_clblast_${clblast_bench_exec} benchmark)
-  target_link_libraries(bench_clblast_${clblast_bench_exec} PUBLIC ${OpenCL_LIBRARIES})
-  target_link_libraries(bench_clblast_${clblast_bench_exec} PUBLIC benchmark)
+  target_link_libraries(bench_clblast_${clblast_bench_exec} PUBLIC benchmark ${OpenCL_LIBRARIES} ${BLAS_LIBRARIES})
   target_link_libraries(bench_clblast_${clblast_bench_exec} PRIVATE clblast)
-  target_include_directories(bench_clblast_${clblast_bench_exec} PUBLIC ${ComputeCpp_INCLUDE_DIRS} ${SYCLBLAS_INCLUDE} ${CLARA_INCLUDE})
+  target_include_directories(bench_clblast_${clblast_bench_exec} PUBLIC ${SYCLBLAS_INCLUDE} ${CLARA_INCLUDE} ${BLAS_INCLUDE_DIRS})
```
mehdi-goli: This should use THIRD_PARTY_INCLUDE

benchmark/common_utils.hpp
```diff
@@ -364,7 +437,7 @@ inline cl_ulong time_events(event_t first_event,
  * (both overall and event time, returned in nanoseconds in a tuple of double)
  */
 template <typename function_t, typename... args_t>
-static std::tuple<double, double> timef(function_t func, args_t&&... args) {
```
mehdi-goli: Keep static keyword here.

benchmark/syclblas/CMakeLists.txt
```diff
@@ -28,6 +34,7 @@ foreach(syclblas_bench ${SYCLBLAS_BENCH_SRCS})
   set_target_properties(bench_${bench_exec} PROPERTIES CXX_STANDARD 11)
   add_dependencies(bench_${bench_exec} benchmark)
   target_link_libraries(bench_${bench_exec} PUBLIC benchmark ${BLAS_LIBRARIES} sycl_blas ComputeCpp::ComputeCpp)
+  target_include_directories(bench_${bench_exec} PUBLIC ${ComputeCpp_INCLUDE_DIRS} ${SYCLBLAS_INCLUDE} ${CLARA_INCLUDE} ${BLAS_INCLUDE_DIRS})
```
mehdi-goli: Use The THIRD_PARTY_INCLUDE

Nyrio:
@przemektmalon you're right, I've fixed that.
@RossBrunton sure, there are probably better ways, I'd say we can do that later (there's a lot of refactoring that could be done in the benchmarks but it's not a priority)

benchmark/clblast/blas3/gemm_batched.cpp
```diff
@@ -0,0 +1,221 @@
...
+  double k_d = static_cast<double>(k);
+  double batch_size_d = static_cast<double>(batch_size);
+
+  state.counters["m"] = m_d;
+  state.counters["k"] = k_d;
+  state.counters["n"] = n_d;
+  state.counters["batch_size"] = batch_size_d;
+
+  {
+    double nflops_AtimesB = (2 * k_d - 1) * m_d * n_d;
```
AdamHarries: How does the FLOP calculation work with batches? Are K/M/N the overall size of the buffer, or the size of each individual matrix?

przemektmalon: I think they are sizes of each, not overall. And the Flops are multiplied by batch size on line 71, I think it's correct.

Nyrio: Yes it's multiplied later, I thought it was clearer and avoids making mistakes when modifying this code to do the multiplication only once.

Nyrio:
We mentioned offline that a possibility would be to create a common function for non-batched and batched benchmarks and the non-batched benchmark cold call it with 1, but given that we need two executables anyway, I don't feel like creating a a hpp file with the common code, it would break the structure just to avoid a bit of duplication.
As said before, the benchmarks are not structured in an optimal way but their code is fairly easy to maintain.

PR #146, opened 2019-05-16
More accurate calculation of n_fl_ops

Nyrio:


benchmark/clblast/blas3/gemm.cpp
```diff
@@ -58,7 +58,7 @@ void run(benchmark::State& state, ExecutorType* executorPtr, int t1, int t2,
   double n_d = static_cast<double>(n);
   double k_d = static_cast<double>(k);
 
-  state.counters["n_fl_ops"] = 2 * (m_d * n_d * k_d) + 3 * (m_d * n_d);
+  state.counters["n_fl_ops"] = (2 * k_d - 1) * m_d * n_d + 3 * m_d * n_d;
```
mehdi-goli: Make it modular and then add each module when it is required

Nyrio: Yes, I think I'll make it another MR

PR #147, opened 2019-05-20
GEMM benchmark with ARM Compute Library

Nyrio:


benchmark/acl/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
+cmake_minimum_required(VERSION 3.2.2)
```
mehdi-goli: why 3.2.2?

benchmark/acl/blas3/gemm.cpp
```diff
@@ -0,0 +1,163 @@
...
+  auto blas_method_def = [&]() -> std::vector<void*> {
+    arm_gemm.run();
+#ifdef ACL_BACKEND_OPENCL
+    arm_compute::CLScheduler::get().sync();
+#endif
+    return {nullptr};
+  };
+
+  // Warm up to avoid benchmarking data transfer
+  // blas_benchmark::utils::warmup(blas_method_def);
```
mehdi-goli: Is this deliberate to be commented out?

Nyrio: It was because the function was moved to the `common_utils` in another MR, now that it's updated on top of everything I enabled the warmup

benchmark/acl/CMakeLists.txt
Nyrio: Because we have that in the other benchmark cmake files as well. I don't know which version I should require so unless a cmake pro (e.g @mmha) suggests something else, I will use the same version than in the other similar files for consistency.

Nyrio:
I've added the verification to the ACL benchmarks as well.
Also I've added a condition to show an error (a red line like verification failures) when the user attempts to use transposed matrices with this library. These benchmarks only support `n,n` mode.

benchmark/acl/CMakeLists.txt
mehdi-goli: 3.4.2

Nyrio: Ok, fixed for all the benchmarks

PR #151, opened 2019-05-22
Added Axpy Copy expression fusion test

przemektmalon:
- Added AXPY COPY expression fusion test.
- Added CMake "ENABLE_EXPRESSION_TESTS" flag that enables all tests in exprtest folder.
- Minor exprtest/CMakeLists.txt changes.

### DIFF
--- a/test/exprtest/CMakeLists.txt
+++ b/test/exprtest/CMakeLists.txt
@@ -31,6 +31,7 @@ set(SYCLBLAS_EXPRTEST ${CMAKE_CURRENT_SOURCE_DIR})
 # compiling tests
 set(SYCL_EXPRTEST_SRCS
   ${SYCLBLAS_EXPRTEST}/blas1_scal_asum_test.cpp
+  ${SYCLBLAS_EXPRTEST}/blas1_axpy_copy_test.cpp
 )
 
 foreach(blas_test ${SYCL_EXPRTEST_SRCS})
--- a/test/exprtest/blas1_axpy_copy_test.cpp
+++ b/test/exprtest/blas1_axpy_copy_test.cpp
@@ -1,0 +1,129 @@
+/***************************************************************************
+ *
+ *  @license
+ *  Copyright (C) Codeplay Software Limited
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  For your convenience, a copy of the License has been included in this
+ *  repository.
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *  SYCL-BLAS: BLAS implementation using SYCL
+ *
+ *  @filename blas1_axpy_copy_test.cpp
+ *
+ **************************************************************************/
+#include "blas_test.hpp"
+#include "sycl_blas.hpp"
+
+// inputs combination
+template <typename scalar_t>
+using combination_t = std::tuple<int, scalar_t, int, int>;
+
+template <typename scalar_t>
+void run_test(const combination_t<scalar_t> combi) {
+  int size;
+  scalar_t alpha;
+  int incX;
+  int incY;
+  std::tie(size, alpha, incX, incY) = combi;
+
+  // Dimensions of input vectors x and y
+  int x_dim = size * incX;
+  int y_dim = size * incY;
+
+  // Input vectors x and y (y is also output of AXPY)
+  std::vector<scalar_t> v_x(x_dim);
+  std::vector<scalar_t> v_y(y_dim);
+  fill_random(v_x);
+  fill_random(v_y);
+
+  // Copy of vector x
+  std::vector<scalar_t> v_xcopy(x_dim);
+
+  // Copy of vector y
+  // Used to store output on host for later comparison with SYCL result
+  std::vector<scalar_t> v_cpu_y(y_dim);
+
+  // Reference BLAS implementation
+  reference_blas::copy(size, v_x.data(), incX, v_xcopy.data(), incX);
+  reference_blas::copy(y_dim, v_y.data(), 1, v_cpu_y.data(), 1);
+  reference_blas::axpy(size, alpha, v_xcopy.data(), incX, v_cpu_y.data(), incY);
+
+  // SYCL-BLAS implementation
+  auto q = make_queue();
+  test_executor_t ex(q);
+
+  // Iterators
+  auto gpu_x_v = blas::make_sycl_iterator_buffer<scalar_t>(x_dim);
+  auto gpu_y_v = blas::make_sycl_iterator_buffer<scalar_t>(y_dim);
+  auto gpu_xcopy_v = blas::make_sycl_iterator_buffer<scalar_t>(x_dim);
+  auto gpu_ycopy_v = blas::make_sycl_iterator_buffer<scalar_t>(y_dim);
+
+  // Copy input data from host to device
+  auto xcp_ev =
+      ex.get_policy_handler().copy_to_device(v_x.data(), gpu_x_v, x_dim);
+  ex.get_policy_handler().wait(xcp_ev);
+  auto ycp_ev =
+      ex.get_policy_handler().copy_to_device(v_y.data(), gpu_y_v, y_dim);
+  ex.get_policy_handler().wait(ycp_ev);
+
+  // Dimensions of vector view for AXPY operations
+  int view_x_dim = (x_dim + incX - 1) / incX;
+  int view_y_dim = (y_dim + incY - 1) / incY;
+
+  // Views
+  // - for axpy operation where arbitrary stride is used
+  auto view_x_incX = make_vector_view(ex, gpu_x_v, incX, view_x_dim);
+  auto view_y_incY = make_vector_view(ex, gpu_y_v, incY, view_y_dim);
+
+  // - for copy operations where we want stride = 1
+  auto view_xcopy_inc1 = make_vector_view(ex, gpu_xcopy_v, 1, x_dim);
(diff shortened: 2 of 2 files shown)


test/exprtest/CMakeLists.txt
```diff
@@ -24,15 +24,15 @@
...
 
 set(SYCLBLAS_EXPRTEST ${CMAKE_CURRENT_SOURCE_DIR})
 
-include_directories(${SYCLBLAS_TEST})
+include_directories(${SYCLBLAS_TEST} ${BLAS_INCLUDE_DIRS})
 
 # compiling tests
 file(GLOB SYCL_EXPRTEST_SRCS
-  ${SYCLBLAS_EXPRTEST}/blas1_iface_test.cpp
+  #${SYCLBLAS_EXPRTEST}/blas1_iface_test.cpp
```
mehdi-goli: please remove the commented code

test/CMakeLists.txt
```diff
@@ -55,4 +55,4 @@ include_directories(
 )
 
 add_subdirectory(${SYCLBLAS_TEST}/unittest)
-#add_subdirectory(${SYCLBLAS_TEST}/exprtest)
+add_subdirectory(${SYCLBLAS_TEST}/exprtest)
```
mehdi-goli: Please Wrap this on EXPRESSION_TEST_ENABLED flag

test/exprtest/CMakeLists.txt
```diff
@@ -54,6 +54,36 @@ foreach(blas_test ${SYCL_EXPRTEST_SRCS})
...
+
+foreach(blas_test ${SYCL_HEADER_ONLY_EXPRTEST_SRCS})
+  include_directories(${SYCLBLAS_SRC})
+  get_filename_component(test_exec ${blas_test} NAME_WE)
+  set(test_exec, ${blas_test})
+  add_executable(${test_exec} main.cpp ${blas_test})
+  add_sycl_to_target(TARGET ${test_exec} SOURCES ${blas_test} main.cpp)
+  set_property(TARGET ${test_exec} PROPERTY CXX_STANDARD 11)
+  add_dependencies(${test_exec} gtest_main)
+  target_link_libraries(${test_exec} PUBLIC gtest_main ${BLAS_LIBRARIES} sycl_blas ComputeCpp::ComputeCpp)
```
mehdi-goli: you dont need to link it against `sycl_blas`. There is no need to build a `sycl_blas` library when testing the expression folder. Also , there is no need to link it with `BLAS_LIBRARIES` as you dont compare the result with this lib

test/exprtest/blas1_axpy_copy_test.cpp
```diff
@@ -0,0 +1,133 @@
...
+ *  SYCL-BLAS: BLAS implementation using SYCL
+ *
+ *  @filename blas1_axpy_copy_test.cpp
+ *
+ **************************************************************************/
+#include "blas_test.hpp"
+#include "sycl_blas.hpp"
+typedef ::testing::Types<blas_test_float<>, blas_test_double<> > BlasTypes;
+
+TYPED_TEST_CASE(BLAS_Test, BlasTypes);
```
mehdi-goli: The test style you use is old. Please use the new interface test so that the code can be coherent. you can look at the blas1 test interface

test/exprtest/blas1_axpy_copy_test.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+  TestClass::set_rand(vX, size);
+  TestClass::set_rand(vZ, size);
+
+  // vector A will store the result of AXPY for comparison later
+  std::vector<scalar_t> vA(size, 0);
+
+
+  // First copy X->Y, and Z->W
+  for (int i = 0; i < size; ++i) {
```
mehdi-goli: Is the stride used for the copy? if so this will be incorrect

test/exprtest/blas1_axpy_copy_test.cpp
```diff
@@ -0,0 +1,133 @@
...
+    vY[i] = vX[i];
+    vW[i] = vZ[i];
+  }
+
+  // Now copute the AXPY using the copied vectors Y and W, and store it in vector A
+  for (int i = 0; i < size; ++i) {
+    if (i % strd == 0) {
+      vA[i] = alpha * vY[i] + vW[i];
```
mehdi-goli: Has the stride been used for axpy, if so this is incorrect

test/exprtest/main.cpp
```diff
@@ -25,11 +25,45 @@
...
+
+struct Args args{};
+
+namespace {
+/**
+ * @fn parse_args
+ * @brief Returns a structure containing the data extracted from the
+ * command-line arguments.
+ */
+void parse_args(int argc, char** argv) {
```
mehdi-goli: If you need this, this can be moved to a common folder so no need for duplication

test/exprtest/CMakeLists.txt
przemektmalon: Currently linking against ${BLAS_LIBRARIES}, as otherwise the CBLAS reference implementation does not link. Is there another way ?

test/exprtest/blas1_axpy_copy_test.cpp
przemektmalon: Updated

test/exprtest/blas1_axpy_copy_test.cpp
przemektmalon: New test code should be correct.

test/exprtest/blas1_axpy_copy_test.cpp
przemektmalon: New test code should be correct.

test/exprtest/main.cpp
przemektmalon: I created a blas_test::utils namespace in blas_test.hpp. It houses both "parse_args" and "fill_random". Without the namespace I was getting duplicate function declaration errors. Because of this change, all the unittest files had to be updated to use "blas_test::utils::fill_random".

test/exprtest/blas1_axpy_copy_test.cpp
```diff
@@ -0,0 +1,144 @@
...
+  // Copy of vector x
+  std::vector<scalar_t> v_xcopy(x_dim);
+
+	// Copy of vector y
+  // Used to store output on host for later comparison with SYCL result
+	std::vector<scalar_t> v_cpu_y(y_dim);
+  
+  // Reference BLAS implementation
+  reference_blas::copy(size, v_x.data(), incX, v_xcopy.data(), incX);
+	reference_blas::copy(y_dim, v_y.data(), 1, v_cpu_y.data(), 1);
```
Nyrio: As a convention, we use *Clang Format* for this project, with the configuration defined in `.clang-format`. Can you run `clang-format -i <your-files>`? I spotted a few formatting issues (e.g indent)

test/exprtest/CMakeLists.txt
```diff
@@ -19,41 +19,37 @@
...
-include_directories(${SYCLBLAS_TEST})
-
 # compiling tests
-file(GLOB SYCL_EXPRTEST_SRCS
-  ${SYCLBLAS_EXPRTEST}/blas1_iface_test.cpp
+set(SYCL_EXPRTEST_SRCS
+  ${SYCLBLAS_EXPRTEST}/blas1_axpy_copy_test.cpp
 )
 
+include_directories(${SYCLBLAS_TEST} ${SYCLBLAS_SRC} ${SYCLBLAS_INCLUDE} ${THIRD_PARTIES_INCLUDE})
```
Nyrio: I think `target_include_directories` is preferred as we use CMake > 3. It's scope is restricted to a target instead of being applied to all targets created from this directory. @mmha, any opinion on this?

test/exprtest/blas1_axpy_copy_test.cpp
przemektmalon: I don't see the bad indents in my editor. I think it might be a tab size thing in VSCode. I'll run the format

test/exprtest/CMakeLists.txt
przemektmalon: I think all targets created from this directory will have the same include_directories. But I can change this.

test/CMakeLists.txt
```diff
@@ -55,4 +55,6 @@ include_directories(
 )
 
 add_subdirectory(${SYCLBLAS_TEST}/unittest)
-#add_subdirectory(${SYCLBLAS_TEST}/exprtest)
+if(ENABLE_EXPRESSION_TESTS)
```
mehdi-goli: Please use option and set the default value for `ENABLE_EXPRESSION_TESTS`

przemektmalon: Added to the main CMakeLists

test/exprtest/blas1_axpy_copy_test.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+  // Copy the result back to host memory
+  auto getResultEv =
+      ex.get_policy_handler().copy_to_host(gpu_y_v, v_y.data(), y_dim);
+  ex.get_policy_handler().wait(getResultEv);
+
+  ASSERT_TRUE(utils::compare_vectors(v_cpu_y, v_y));
+}
+
+#ifdef STRESS_TESTING
```
mehdi-goli: The stress testing is used for application like gemm which takes too long to execute. If the stress testing for you takes less that two second you don't need to create stress testing

test/exprtest/blas1_axpy_copy_test.cpp
```diff
@@ -0,0 +1,129 @@
...
+  auto yCopyOp = make_op<Assign>(view_ycopy_inc1, view_y_incY);
+
+  // AXPY expressions
+  auto axpy_scal_op = make_op<ScalarOp, ProductOperator>(alpha, xCopyOp);
+  auto axpy_add_op = make_op<BinaryOp, AddOperator>(yCopyOp, axpy_scal_op);
+  auto copy_axpy_op_tree = make_op<Assign>(view_y_incY, axpy_add_op);
+
+  // Execute the COPY+AXPY tree
+  auto axpy_event = ex.execute(copy_axpy_op_tree);
+  ex.get_policy_handler().wait(axpy_event);
```
RossBrunton: @mehdi-goli is this working at the moment? Wait with an event?

przemektmalon: Adding wait with an event was fixing the intermittent `deadlock avoided` issues for me so I think it works.

PR #154, opened 2019-05-27
Added Scal Asum expression fusion test

przemektmalon:
- Added a Scal + Asum expression tree fusion test

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -55,4 +55,6 @@ include_directories(
 )
 
 add_subdirectory(${SYCLBLAS_TEST}/unittest)
-#add_subdirectory(${SYCLBLAS_TEST}/exprtest)
+if(ENABLE_EXPRESSION_TESTS)
+  add_subdirectory(${SYCLBLAS_TEST}/exprtest)
+endif(ENABLE_EXPRESSION_TESTS)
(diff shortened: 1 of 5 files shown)


test/CMakeLists.txt
```diff
@@ -55,4 +55,6 @@ include_directories(
 )
 
 add_subdirectory(${SYCLBLAS_TEST}/unittest)
-#add_subdirectory(${SYCLBLAS_TEST}/exprtest)
+if(ENABLE_EXPRESSION_TESTS)
```
mehdi-goli: Please use option and set the default value for ENABLE_EXPRESSION_TESTS

test/unittest/CMakeLists.txt
```diff
@@ -28,7 +28,7 @@ project(syclblas_unittest)
 
 set(SYCLBLAS_UNITTEST ${CMAKE_CURRENT_SOURCE_DIR})
 
-include_directories(${SYCLBLAS_TEST} ${BLAS_INCLUDE_DIRS})
+include_directories(${SYCLBLAS_TEST})
```
mehdi-goli: Is this related to this PR?

test/CMakeLists.txt
przemektmalon: Added to main CMakeLists

test/unittest/CMakeLists.txt
przemektmalon: Not strictly. Just noticed that this include was not necessary, and that ${BLAS_INCLUDE_DIRS} was always empty for me, so I removed it. Should I add it back ?

mehdi-goli: yes, it is used when you install the system blas in a custom place

PR #155, opened 2019-05-27
Update tests to not use BLAS_Test and updated buffer_test

RossBrunton:
This is mostly a cleanup that allows deletion of a huge amount of
unused stuff in blas_test.hpp.

Also, buffer_test was updated to match the format of the rest of
the tests.


test/exprtest/blas1_iface_test.cpp
```diff
@@ -25,6 +25,8 @@
 
 #include "blas_test.hpp"
 
+// This test is not build or executed, and will not compile
```
Nyrio: (buil**t**)
What is this test for then? Should this stay?

RossBrunton: I was updating this test, and saw that it wasn't being built, so figured I'd add a note to future travellers. Since Przemek seems to have removed it, I'll remove my change for a cleaner merge.

PR #156, opened 2019-06-03
Added benchmark csv config files (dnn sizes + powers of 2)

Nyrio:


benchmark/README.md
```diff
@@ -95,6 +95,10 @@ n,t,64,128,64,0.5,0.5
 t,n,13,3,7,0,0.7
 ```
 
+The folder `config_csv` provides a few files corresponding to sizes that are
```
RossBrunton: Are the filenames intuitive for people who work with neural networks? Also, I personally would like subfolders for each blas level, eg `config_csv/blas1/`.

Nyrio: I've reorganized the folder (it was like that because I have a script to sync it with a database but I'll edit it).
I guess the names are intuitive for people who work with neural networks, I didn't come up with these names myself :)

PR #158, opened 2019-06-03
View sycl opt

mehdi-goli:
This PR remove the unnecessary if statement check inside view_sycl for every single element on GPU.


Nyrio:
Here are some benchmark results of this branch against master.

**ARM Mali G71 MP8**

No significant difference for most blas 1 functions but a real improvement on `scal`. No significant difference for gemv, seems a wee bit better for non-transposed matrices and a wee bit worse on transposed ones.

|scal|gemv|
|----|----|
| viewsyclopt_mali_scal | viewsyclopt_mali_gemv |

**Intel HD Graphics 530**

Performance regression on pretty much all the blas 1 functions especially for big sizes. No significant difference for gemv, seems a wee bit worse for transposed matrices.

|asum|axpy|dot|iamax|scal|
|----|----|---|-----|----|
| viewsyclopt_intel_asum | viewsyclopt_intel_axpy | viewsyclopt_intel_dot | viewsyclopt_intel_iamax | viewsyclopt_intel_scal |

src/views/view_sycl.hpp
```diff
@@ -57,98 +57,58 @@ struct VectorView<
...
+  const increment_t strd_;  // never size_t, because it could be negative
   container_t data_;
-  index_t size_data_;
-  index_t size_;
-  index_t disp_;
-  increment_t strd_;  // never size_t, because it could negative
 
   using value_t = scalar_t;
 
+  static SYCL_BLAS_INLINE const index_t calc_size(container_t &data,
```
AdamHarries: What is this function for? Could it please have a documentation comment.

przemektmalon: I'm also confused by the negative stride. Isn't increment_t supposed to be always >0 ?

mehdi-goli: @adam added. @przemektmalon  no it can be negative that is the reason every index and increment parameter in BLAS is signed.

mehdi-goli:
All the comment here has been addressed in PR 161

src/views/view_sycl.hpp
przemektmalon: Ah ok, I misinterpreted the comment at line 62

PR #161, opened 2019-06-05
optimising view sycl to reduce pointer arithmetic inside the kernel.

mehdi-goli:
This PR reduces the pointer arithmetic operation inside the kernel. Since view_sycl.hpp runs inside the kernel this  will reduce the number of  arithmetic operation requires to load/store the data


include/blas_meta.h
```diff
@@ -25,18 +25,56 @@
 
 #ifndef SYCL_BLAS_META_H
 #define SYCL_BLAS_META_H
+#include <iostream>
```
RossBrunton: Is this required?

include/views/view.h
```diff
@@ -81,10 +80,20 @@ struct VectorView {
   container_t &get_data();
 
   /*!
-   * @brief Returns the displacement
+   * @brief Returns a reference to the container
```
RossBrunton: Doesn't return a reference.

include/views/view.h
```diff
@@ -253,18 +212,18 @@ struct VectorViewTypeFactory {
                  index_t, increment_t>;
 };
 
-template <typename policy_t, typename element_t, typename index_t>
+template <typename policy_t, typename element_t, typename index_t, typename acc>
```
RossBrunton: Can you use a clearer name than "acc"?

include/views/view.h
```diff
@@ -81,10 +80,20 @@ struct VectorView {
...
+
+  /*! get_access_displacement.
+   * @brief get displacement from the origin.
    */
   index_t get_access_displacement();
 
+  /*! set_access_displacement.
+   * @brief sett displacement from the origin.
+   */
+  void set_access_displacement();
```
RossBrunton: This doesn't take a parameter? It looks like it should take a parameter. If it isn't supposed to, maybe enable/disable_access_displacement?

test/unittest/blas2/blas2_gemv_test.cpp
```diff
@@ -26,7 +26,7 @@
 #include "blas_test.hpp"
 
 template <typename T>
-using combination_t = std::tuple<int, int, bool, T, T, int, int, int>;
+using combination_t = std::tuple<int, int, T, T, bool, int, int, int>;
```
RossBrunton: Is there a reason argument order was changed?

include/views/view.h
mehdi-goli: No it does not take parameter. It will add the accessor displacement to the accessor. But it will be applied at the beginning of the matrix. it will add the displacement from by adding disp_ to the ptr_ at the begining of each kernel so that the kernel wont repeat this opetration at every eval call

test/unittest/blas2/blas2_gemv_test.cpp
mehdi-goli: yes because the order was wrong

include/views/view.h
AdamHarries: Could this be commented and explained please. It’s not clear from reading the name of the method.

mehdi-goli: added

include/views/view.h
mehdi-goli: Fixed

include/views/view.h
mehdi-goli: Fixed

include/blas_meta.h
mehdi-goli: removed

PR #162, opened 2019-06-05
Temporarily fix intermittent failures in asum and dot tests

Nyrio:


Nyrio:
@RossBrunton was there a reason for not using `ASSERT_TRUE` in `dot` or was it forgotten? Just want to make sure

test/unittest/blas1/blas1_asum_test.cpp
```diff
@@ -59,6 +59,8 @@ void run_test(const combination_t combi) {
 
   // Validate the result
   ASSERT_TRUE(utils::almost_equal(out_s[0], out_cpu_s));
+
+  ex.get_policy_handler().get_queue().wait();
```
RossBrunton: Why is there a wait here?

mehdi-goli: This is temporary work around for travis issue. @RossBrunton

PR #163, opened 2019-06-07
Separating gemm implementations (ref, local mem, no local mem)

Nyrio:
This will be needed later on as we're going to keep adding implementations (e.g tall/skinny) and it shouldn't all go in a single `blas_trees.hpp` file.


mehdi-goli:
Looks good to me! did you check if it is passed in integration testing with other in-house library?

PR #164, opened 2019-06-13
Tall & Skinny GEMM

Nyrio:
This pull request adds an optimized algorithm for the multiplication of tall and skinny matrices, e.g the case where the common dimension of A and B is big and the external dimensions are small. The algorithm is tiled and uses the same tile configuration system as the other Gemm algorithms (but generally different tile dimensions).

Suggested order of reading: first the specializations in `executor_sycl.hpp` to understand the general structure of the algorithm, then `gemm_partial_local.hpp` when you can allocate some time for it. And then all the other small changes in many different files.

I will probably produce some documentation of the algorithm soon, at least for maintainability.

At the moment it has been optimized only for the Intel GPU backend, others will follow in separate pull requests.


src/operations/blas3/gemm_no_local.hpp
```diff
@@ -56,11 +56,12 @@ template <typename input_t, typename output_t, bool DoubleBuffer, bool NbcA,
...
 class Gemm<input_t, output_t, DoubleBuffer, NbcA, NbcB, ClSize, tile_type,
            TransA, TransB, element_t, is_beta_zero,
-           static_cast<int>(Gemm_t::no_local_memory)> {
+           static_cast<int>(Gemm_memory_t::no_local_memory),
+           static_cast<int>(Gemm_shape_t::classic)> {
  public:
   using value_t = element_t;
   using index_t = typename std::make_signed<typename input_t::index_t>::type;
-  static constexpr int type = static_cast<int>(Gemm_t::no_local_memory);
+  // static constexpr int memory_type = static_cast<int>(Gemm_memory_t::no_local_memory);
```
mehdi-goli: remove the commented code

src/operations/blas3/gemm_partial_local.hpp
```diff
@@ -0,0 +1,454 @@
...
+#include "gemm_common.hpp"
+
+namespace blas {
+
+template <typename input_t, typename output_t, bool DoubleBuffer, bool NbcA,
+          bool NbcB, int ClSize, typename tile_type, bool TransA, bool TransB,
+          typename element_t>
+class GemmPartial<input_t, output_t, DoubleBuffer, NbcA, NbcB, ClSize,
+                  tile_type, TransA, TransB, element_t,
+                  static_cast<int>(Gemm_memory_t::local_memory)> {
```
mehdi-goli: gemm_memory_t{ local, no_local}

src/operations/blas3/gemm_partial_local.hpp
```diff
@@ -0,0 +1,454 @@
...
+    const cl::sycl::range<1> wgs(local_thread_size);
+    return cl::sycl::nd_range<1>(nwg * wgs, wgs);
+  }
+
+  template <typename local_memory_t>
+  SYCL_BLAS_INLINE void eval(local_memory_t scratch,
+                             cl::sycl::nd_item<1> id) noexcept {
+    /* Pointers to the scratch memory (lhs and rhs) */
+    auto scratch_ptr = scratch.localAcc.get_pointer().get();
+    auto rhs_scratch_ptr = scratch_ptr + (2 * ld_lhs_tile * tile_size_dim_k);
```
mehdi-goli: just make double buffering optional. (DoubleBuffer +1)*ld_lhs_tile*tile_size_dim_k

src/operations/blas3/gemm_partial_local.hpp
```diff
@@ -0,0 +1,454 @@
...
+      index_t global_col_offset, index_t global_rows, index_t global_cols) {
+    index_t local_linear_id = local_id;
+    const index_t global_tile_col_offset =
+        global_col_offset + block_cols * tile_idx;
+    const index_t local_thread_size = local_thread_size_n * local_thread_size_m;
+
+    // Double buffering
+    constexpr index_t block_size = ld_tile * block_cols;
+    const index_t local_mem_offset = (tile_idx & 1) * block_size;
+
```
mehdi-goli: #pragma unroll

src/operations/blas3/gemm_partial_local.hpp
Nyrio: Done, tested and benchmarked (spoiler: it's less performant if you disable double buffering)

src/operations/blas3/gemm_partial_local.hpp
Nyrio: Added, I don't see any performance gain but there is no regression either

test/unittest/blas3/blas3_gemm_tall_skinny_test.cpp
```diff
@@ -0,0 +1,105 @@
...
+ *
+ **************************************************************************/
+
+#include "blas_test.hpp"
+
+template <typename scalar_t>
+using combination_t =
+    std::tuple<int, int, int, char, char, scalar_t, scalar_t, int, int, int>;
+
+const auto combi = ::testing::Combine(::testing::Values(7, 65),      // m
```
AdamHarries: Do these test values hit all the possible code paths in the TSGEMM?

src/operations/blas3/gemm_partial_local.hpp
```diff
@@ -0,0 +1,465 @@
...
+          const element_t privateRhs = rhs_scratch_ptr[rhs_index + rhs_offset];
+
+          index_t lhs_index = 0;
+#pragma unroll
+          for (index_t wLPTM = 0; wLPTM < work_per_thread_m; wLPTM++) {
+            // load a LHS element from the scratch buffer
+            const element_t privateLhs = scratch_ptr[lhs_index + lhs_offset];
+
+            // Perform a manual MAD.
+            private_res[wLPTM + idx] += (privateLhs * privateRhs);
```
mehdi-goli: replace this with `private_res[wLPTM + id] = cl::sycl::mad(privateLhs, privateRhs, private_res[wLPTM + idx]);`

src/operations/blas3/gemm_partial_local.hpp
```diff
@@ -0,0 +1,465 @@
...
+    index_t private_index_offset = 0;
+
+#pragma unroll
+    for (index_t wLPTN = 0; wLPTN < work_per_thread_n; wLPTN++) {
+      index_t private_index = private_index_offset;
+
+      index_t slice_row = slice_row_offset;
+#pragma unroll
+      for (index_t wLPTM = 0; wLPTM < work_per_thread_m; wLPTM++) {
+        if (slice_row < m_ && slice_col < n_) {
```
mehdi-goli: in every write, you have to perform the check which reduces the performance, the same rule that you use for computing the internal block will apply here for the store as well. So none of the internal blocks should go through the check and only the edges should go there. so the store function should be wrap in a method and based on whether or not it is internal you need to go through a true or false path

src/operations/extension/reduction_partial_rows.hpp
```diff
@@ -0,0 +1,237 @@
...
+    for (index_t wpr = 0; wpr < rows_per_item; wpr++) {
+      accumulators[wpr] = init_val;
+    }
+
+    /* Sequential reduction level:
+     * Load multiple elements from the global memory, reduce them together and
+     * store them in the local memory */
+    {
+      index_t elem_col = global_col;
+      while (elem_col < cols_) {
```
mehdi-goli: Did you get better performance when you created a block per thread for the rows. I don't think it will give better performance than allocating one non-reducing element per thread and creates multiple workgroup and allow the gpu schedule them. Because in this case you have nested for loop and nested for loop means nested boundary checks. If your experiment shows this is the better approach, then use the internal_block technique used in gemm to avoid the check for the majurity of blocks

src/operations/extension/reduction_partial_rows.hpp
```diff
@@ -0,0 +1,237 @@
...
+        local_memory_idx += work_group_rows;
+      }
+    }
+
+    /* Parallel-reduction level:
+     * Tree-based reduction in local memory */
+    {
+      const index_t local_memory_lhs = local_col * local_memory_rows;
+#pragma unroll
+      for (index_t stride = work_group_cols / 2; stride > 0; stride /= 2) {
```
mehdi-goli: similarly because you are using the tile this means that you need to have 2 nested for loop here. on devices which has enough registers, this might give better performance as you unroll the for loop and creating more tasks per thread, however, in devices with limited number of register, using too much register may cause register spill and can cause performance drop. If you wanted to use this technique at least provide a macro  or  compile time template parameters to disable it for devices with small number of registers

test/exprtest/CMakeLists.txt
```diff
@@ -26,13 +26,16 @@ cmake_minimum_required(VERSION 3.2.2)
...
+add_definitions(-D${TARGET})
+
 set(SYCLBLAS_EXPRTEST ${CMAKE_CURRENT_SOURCE_DIR})
 
 # compiling tests
 set(SYCL_EXPRTEST_SRCS
   ${SYCLBLAS_EXPRTEST}/blas1_scal_asum_test.cpp
   ${SYCLBLAS_EXPRTEST}/blas1_axpy_copy_test.cpp
   ${SYCLBLAS_EXPRTEST}/collapse_nested_tuple.cpp
+  ${SYCLBLAS_EXPRTEST}/extension_reduction_partial_rows_test.cpp
```
mehdi-goli: wrap this in BLAS_EXTENSION macro as well

src/operations/blas3/gemm_common.hpp
```diff
@@ -50,54 +50,12 @@ ENABLE_TYPE_STRING(double)
...
- *                 block-level tile
- * @tparam WgCols  the number of item-level tiles within each row of
- *                 block-level tile
- * @tparam TlRows  the number of block-level tiles within each column of
- *                 top-level tile
- * @tparam TlCols  the number of block-level tiles within each row of
- *                 top-level tile
- *
- * @see GemmFactory
+ * @note See the struct definition in include/operations/blas3_trees.h for more
```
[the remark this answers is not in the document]
Nyrio: This specific line is just avoiding duplication: the exact same documentation was in two files and it's harder to maintain.

Concerning documentation of the new changes in general, I have written a pdf with nice drawings and explanations of all the optimizations. I can add some comments here but in general it might be good to have a folder in this project for the non-textual documentation such as what Goran and I wrote. It is easier to read a proper pdf or markdown with images than in-code docs

cmake/CmakeFunctionHelper.cmake
```diff
@@ -39,42 +39,54 @@ endif()
...
-  set(gemm_configuration_1 32 "false" "false" "false" 128 8 4 4 8 1 1 "local_memory")
+  set(gemm_configuration_0 32 "false" "false" "false" 128 4 8 8 4 1 1 "local" "classic")
+  set(gemm_configuration_1 32 "false" "false" "false" 128 8 4 4 8 1 1 "local" "classic")
+
   list(APPEND gemm_configuration_lists gemm_configuration_0 gemm_configuration_1)
 elseif(${TARGET} STREQUAL "ARM_GPU")
-  set(gemm_configuration_0 64 "false" "false" "false" 64 4 4 8 8 1 1 "no_local_memory")
-  set(gemm_configuration_1 128 "false" "false" "false" 64 4 8 16 8 1 1 "no_local_memory")
-  set(gemm_configuration_2 32 "false" "false" "false" 64 8 4 4 8 1 1 "no_local_memory")
+  set(gemm_configuration_0 64 "false" "false" "false" 64 4 4 8 8 1 1 "no_local" "classic")
```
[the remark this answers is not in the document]
Nyrio: Yes. I can call it `legacy` if you prefer, or something else if you have ideas

src/operations/blas3/gemm_common.hpp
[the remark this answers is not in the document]
Nyrio: (my opinion is that the long explanation should be out of the code, and the comments in the code should only highlight the structure and what specific blocks do)

test/unittest/blas3/blas3_gemm_tall_skinny_test.cpp
Nyrio: Not exactly, I will explain why, and why it's not a problem:

- heuristics are used independently on each platform to use different tile dimensions. Having the tests hit all the code paths, and all the possible parameterizations for each platform would require a number of tests that we cannot afford if we want to keep the testing reasonably short. I have hit all these parameterizations with verified benchmarks when optimizing the heuristics, though.
- Also, the tests only use a ldc value different than m, which is handled in a different branch in the executor. In the benchmarks the ldc value is always m so this branch is verified by the benchmarks. I realize this can be seen as an issue and I can change the test cases, but it will double their number.

There's a trade-off here, between exhaustivity and testing time.

cmake/CmakeFunctionHelper.cmake
```diff
@@ -39,42 +39,54 @@ endif()
...
   list(APPEND gemm_configuration_lists gemm_configuration_0 gemm_configuration_1)
 else() # default cpu backend
-  set(gemm_type "no_local_memory" )
-  if(NAIVE_GEMM)
-    set(gemm_type "naive")
-  endif()
-  set(gemm_configuration_0 64 "false" "false" "false" 64 8 8 8 8 1 1 "${gemm_type}")
-  set(gemm_configuration_lists "")
-  list(APPEND gemm_configuration_lists gemm_configuration_0)
+  set(gemm_configuration_0 64 "false" "false" "false" 64 8 8 8 8 1 1 "no_local" "naive")
```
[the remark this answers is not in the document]
Nyrio: Yes, my mistake. I thought that naive could be selected with heuristics but it's actually a `ifdef`

src/executors/executor_sycl.hpp
```diff
@@ -206,28 +211,209 @@ Executor<PolicyHandler<codeplay_policy>>::execute(
...
+      reduction_step(in, out, group_count_cols);
+  auto step_range = reduction_step.get_nd_range(num_compute_units);
+  return execute_tree<using_local_memory::enabled>(
+      queue, reduction_step, step_range.get_local_range()[0],
+      step_range.get_global_range()[0], local_memory_size);
+}
+
+// ReductionPartialRows
+template <>
+template <typename operator_t, typename input_t, typename output_t, int ClSize,
```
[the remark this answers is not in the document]
Nyrio: It is used for tsgemm which is a part of BLAS. But what I can do is disable it by default and add code that forces it to true if tsgemm is enabled. Is that a good idea @mehdi-goli ?

src/operations/blas3/gemm_partial_local.hpp
```diff
@@ -0,0 +1,465 @@
...
+  using value_t = element_t;
+
+  input_t a_;
+  input_t b_;
+  output_t cube_;
+
+  element_t alpha;
+
+  /* Matrix dimensions */
+  const index_t m_;
```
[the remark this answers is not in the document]
Nyrio: They are read from the views. Is there any reason to pass them as an argument instead of using the matrix views? It seems better if the gemm partial isn't given the wrapper

src/executors/executor_sycl.hpp
```diff
@@ -206,28 +211,209 @@ Executor<PolicyHandler<codeplay_policy>>::execute(
...
+          typename element_t, bool is_beta_zero, int GemmMemoryType>
+inline typename codeplay_policy::event_t
+Executor<PolicyHandler<codeplay_policy>>::execute(
+    Gemm<input_t, output_t, DoubleBuffer, NbcA, NbcB, ClSize, tile_type, TransA,
+         TransB, element_t, is_beta_zero, GemmMemoryType,
+         static_cast<int>(gemm_algorithm_t::tall_skinny)>
+        gemm_wrapper) {
+  using index_t = typename std::make_signed<typename input_t::index_t>::type;
+
+  const index_t rows = gemm_wrapper.m_;
```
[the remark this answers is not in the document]
Nyrio: The variable is used a lot of times in this function so I found it more readable like that (4 characters instead of 15)

src/operations/extension/reduction_partial_rows.hpp
Nyrio: You're right, removing that gives best performance. I made the change.

src/operations/extension/reduction_partial_rows.hpp
Nyrio: Solved

src/executors/executor_sycl.hpp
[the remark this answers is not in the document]
Nyrio: I suggest this goes into another MR: it is related to the reduction which is not part of this MR (this branch is based on the `reduction-partial-rows` which is supposed to go in before), but the change needs to take into account whether or not tsgemm is used.

Also the size of the library will only be significantly increased if the reduction is used, and in that case it means that we want it anyway.

src/operations/blas3/gemm_partial_local.hpp
Nyrio: This change will go with the other optimizations that I included in #178 (unless we want to move these optimizations here, but I feel like we should merge this MR as soon as possible and add optimizations in another MR)

src/operations/blas3/gemm_partial_local.hpp
Nyrio: Done

src/operations/blas3/gemm_partial_local.hpp
```diff
@@ -0,0 +1,465 @@
...
+  /* Number of private summation registers */
+  static constexpr index_t private_res_size =
+      work_per_thread_m * work_per_thread_n;
+
+  /* Transpose mode for matrices A and B */
+  static constexpr bool trans_a = TransA;
+  static constexpr bool trans_b = TransB;
+
+  /* Work groups per dimension m, n, k */
+  const index_t group_count_m;
```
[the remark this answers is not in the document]
Nyrio: The dimensions are not passed to the constructor but read from the matrix views. The constructor takes A, B, the output buffer, alpha and the number of work groups on the k dimension. The formers are also passed to the other gemm kernels so passing them like that is a matter of consistency. The latter is specific to this kernel and therefore necessary.

test/exprtest/CMakeLists.txt
Nyrio: See the answer above. A good change but will go in another MR

src/executors/executor_sycl.hpp
```diff
@@ -206,28 +211,209 @@ Executor<PolicyHandler<codeplay_policy>>::execute(
...
+
+  /* First step: partial gemm */
+  /* Create the cube buffer that will hold the output of the partial gemm */
+  auto cube_buffer = make_sycl_iterator_buffer<element_t>(rows * cols * depth);
+  auto cube = make_matrix_view<col_major>(*this, cube_buffer, rows * cols,
+                                          depth, rows * cols);
+  /* Execute the partial gemm operation */
+  GemmPartial<input_t, output_t, DoubleBuffer, NbcA, NbcB, ClSize, tile_type,
+              TransA, TransB, element_t, GemmMemoryType>
+      gemm_partial(gemm_wrapper.a_, gemm_wrapper.b_, cube, gemm_wrapper.alpha_,
```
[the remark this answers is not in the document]
Nyrio: See answer below

src/operations/blas3/gemm_partial_local.hpp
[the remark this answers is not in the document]
Nyrio: See answer below

src/operations/blas3/gemm_partial_local.hpp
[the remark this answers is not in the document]
Nyrio: See related answer below

test/unittest/blas3/blas3_gemm_tall_skinny_test.cpp
Nyrio: I added the case where `ldc_mul` = 1 to have more exhaustive testing as the test time is still reasonable (17s).

PR #165, opened 2019-06-24
Updated tests for iamax/iamin

RossBrunton:
It now has four "profiles" which determine the generation of the
data. Also added a common header file so that the code for these
profiles isn't duplicated.

We also don't write the reference blas output into a float/double,
because that was wrong.

### DIFF
--- a/test/unittest/blas1/blas1_iaminmax_common.hpp
+++ b/test/unittest/blas1/blas1_iaminmax_common.hpp
@@ -1,0 +1,79 @@
+/***************************************************************************
+ *
+ *  @license
+ *  Copyright (C) Codeplay Software Limited
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


test/unittest/blas1/blas1_iaminmax_common.hpp
```diff
@@ -0,0 +1,79 @@
...
+
+template <typename scalar_t>
+void populate_data(generation_mode_t mode, scalar_t limit,
+                   std::vector<scalar_t> &vec) {
+  switch (mode) {
+    case generation_mode_t::Random:
+      fill_random(vec);
+      break;
+    case generation_mode_t::Limit:
+      std::fill(vec.begin(), vec.end(), limit);
```
mehdi-goli: What will the value of fill will be here?

RossBrunton: Left a comment on `generation_mode_t::Limit`.

Nyrio: `limit`, which is passed by the iamin or iamax test

PR #166, opened 2019-06-25
Fix constants so that the value matches the enum name

Nyrio:


mehdi-goli:
Thanks for the PR. This has been addressed in https://github.com/codeplaysoftware/sycl-blas/pull/169

PR #167, opened 2019-06-25
Added partial reduction on the rows

Nyrio:
This PR add a `Reduction` class that is a wrapper around different reduction methods, and `ReductionPartialRows` which holds the kernel for the reduction of the rows of a matrix. The executor contains a specialization for executing the former, using the latter.

The code has support for 1-step or 2-step but at the moment 1-step gives better results in most cases so it is preferred.


src/executors/executor_sycl.hpp
```diff
@@ -230,6 +230,88 @@ Executor<PolicyHandler<codeplay_policy>>::execute(
...
+              element_t, static_cast<int>(Reduction_t::partial_rows)>
+        reduction_wrapper) {
+  using index_t = typename std::make_signed<typename input_t::index_t>::type;
+  using params_t = blas::ReductionRows_Params<index_t, element_t, ClSize,
+                                              WgSize, WorkPerItem>;
+
+  /* Extract data from the reduction wrapper */
+  const index_t rows_ = reduction_wrapper.rows_,
+                cols_ = reduction_wrapper.cols_;
+  input_t &in_ = reduction_wrapper.in_, out_ = reduction_wrapper.out_;
```
RossBrunton: Can this be split into two lines?

src/executors/executor_sycl.hpp
```diff
@@ -230,6 +230,88 @@ Executor<PolicyHandler<codeplay_policy>>::execute(
...
+
+  const index_t num_compute_units = policy_handler_.get_num_compute_units();
+
+  /* Choose at run-time whether to do a one-step or two-step reduction */
+  const bool do_first_step =
+      cols_ > params_t::work_group_cols * params_t::work_group_cols;
+  // TODO: find out in which cases it is better to use 2-step reduction
```
RossBrunton: Obligatory "Does this need a Jira?"

src/executors/executor_sycl.hpp
```diff
@@ -230,6 +230,88 @@ Executor<PolicyHandler<codeplay_policy>>::execute(
...
+  /* Create an empty event vector */
+  typename codeplay_policy::event_t reduction_event;
+
+  /* 2-step reduction */
+  if (do_first_step) {
+    static constexpr index_t group_count_cols = params_t::work_group_cols;
+
+    /* Create a temporary buffer */
+    std::vector<element_t> temp_vector(rows_ * group_count_cols);
+    auto temp_buffer = make_sycl_iterator_buffer<element_t>(
```
RossBrunton: make_sycl_iterator_buffer has a constructor that automatically allocates, so you don't need temp_buffer.

include/operations/extension_trees.h
```diff
@@ -0,0 +1,103 @@
...
+ * @brief Wrapper around the reduction.
+ *
+ * The executor will specialize the execution for every reduction type and use
+ * the specific reduction classes
+ */
+template <typename operator_t, typename input_t, typename output_t, int ClSize,
+          int WgSize, int WorkPerItem, typename element_t, int Reduction_type>
+class Reduction {
+ public:
+  using index_t = typename std::make_signed<typename input_t::index_t>::type;
```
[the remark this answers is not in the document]
Nyrio: It is something that I saw in gemm implementation, a signed index type is always used. I assume it is to avoid problems with all the `int` to `index_t` conversions (the template parameters are `int`)
@mehdi-goli any opinion on this?

src/executors/executor_sycl.hpp
Nyrio: It will be part of the optimization of tall and skinny gemm, I guess. Ideally there should be a benchmark for the reduction but the current benchmarks only measure the performance of the API and not the underlying kernels.

src/executors/executor_sycl.hpp
Nyrio: That's right, thank you.

src/executors/executor_sycl.hpp
Nyrio: Done

include/operations/extension_trees.h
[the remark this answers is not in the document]
mehdi-goli: It has been changed in GEMM to signed due to the way that GEMM algorithm has been implemented. In Gemm if the index goes negative, it does not load the data. If you have such a concept in your kernel then apply the make_signed, if not, it is not necessary to be here.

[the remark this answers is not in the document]
Nyrio: Ok, I don't need it so I removed the `make_signed`

src/executors/executor_sycl.hpp
Nyrio: I've created a JIRA issue.

Nyrio:
I noticed that the constructor of the partial reduction was taking the numbers of rows and columns as parameters but reading them from the matrix views anyway. I fixed that. Also merged master on top of this one so it's now passing the test (thanks to the fix that Adam and I did in the constants).

include/operations/extension_trees.h
```diff
@@ -0,0 +1,103 @@
...
+
+#include <CL/sycl.hpp>
+
+namespace blas {
+
+/*!
+ * @brief Determines which type of reduction to perform
+ */
+enum class Reduction_t : int {
+  full = 0,  // Not implemented yet
```
AdamHarries: Is this PR waiting on other PRs before it can be merged and working?

Nyrio: The ones for the full reduction and the reduction of the columns will come later if ever. They're not used in any blas algorithm yet, just a nice-to-have.

Do you think these lines in the enum should be commented out?

Side note: the PR is waiting on #174 though, but #174 is not that big of a change

Nyrio: #174 merged. This PR doesn't have any more dependencies.

src/views/view_sycl.hpp
```diff
@@ -160,7 +160,7 @@ struct VectorView<
     return eval(ndItem.get_global_id(0));
   }
 
-  SYCL_BLAS_INLINE scalar_t eval(cl::sycl::nd_item<1> ndItem) const {
+  SYCL_BLAS_INLINE const scalar_t eval(cl::sycl::nd_item<1> ndItem) const {
```
RossBrunton: Why is this const?

src/operations/extension/reduction_partial_rows.hpp
```diff
@@ -0,0 +1,237 @@
...
+  }
+  /*!
+   * @brief get_num_workgroup_cluster. This function is used to extend the
+   * number of work group clusters, in order to make sure that at least 4
+   * operations are available per work group. The number 4 is used based on
+   * empirical research.
+   */
+  SYCL_BLAS_INLINE index_t
+  get_num_workgroup_cluster(index_t compute_units) noexcept {
+    return ((4 * compute_units - 1) / get_workgroup_cluster() + 1);
```
RossBrunton: Put 4 in a constant somewhere.

src/views/view_sycl.hpp
CodeplayBen: I think this was supposed to be addressed in #174 but it seems to have just been missed.

Nyrio: It was indeed fixed in the other branch, but I didn't catch it when merging `master` into this branch because there was no conflict for this one (this branch had the changes of the other before the change). Thanks for catching that

src/operations/extension/reduction_partial_rows.hpp
Nyrio: This function is exactly copied from the legacy Gemm algorithm (and the constant verified empirically as the most performant).
I agree that the constant could be put elsewhere to be controlled (or even made a template parameter), but where would you put it? Unless there's a common location that makes sense, the best place for readability is right here

RossBrunton: At the moment at least, I'd just put it at the top of the file.

Nyrio: In that case, I'd rather let it here, when you're reading the code you don't have to jump somewhere else to get the value. Moreover it's not meant to be changed often, or even to be changed at all.

RossBrunton: Why do you need to know the value?

Nyrio: Because it gives you an idea of the number of work groups. And I think it's better to let this value here at the moment, unless we want to create a parameter to control this (e.g to use different values on different architectures).

Nyrio: (this function can be seen as a heuristics to select the number of work groups, which is why I think it's fine to have the constant here; and the constant is not meant to be used anywhere else)

RossBrunton: Ah, so in a way the whole "function" could be seen as the constant then? That's fair then.

src/executors/executor_sycl.hpp
```diff
@@ -230,6 +230,91 @@ Executor<PolicyHandler<codeplay_policy>>::execute(
...
+  /* Extract data from the reduction wrapper */
+  const index_t rows_ = reduction_wrapper.rows_,
+                cols_ = reduction_wrapper.cols_;
+  input_t& in_ = reduction_wrapper.in_;
+  output_t& out_ = reduction_wrapper.out_;
+
+  const index_t num_compute_units = policy_handler_.get_num_compute_units();
+
+  /* Choose at run-time whether to do a one-step or two-step reduction */
+  constexpr bool do_first_step = false;
```
AdamHarries: Could this be renamed? The comment makes it clear what the purpose is, but `do_first_step` is more than a little confusing, as it sounds like it's an optional thing, wheras sometimes it's required.

src/operations/extension/reduction_partial_rows.hpp
```diff
@@ -0,0 +1,237 @@
...
+    for (index_t wpr = 0; wpr < rows_per_item; wpr++) {
+      accumulators[wpr] = init_val;
+    }
+
+    /* Sequential reduction level:
+     * Load multiple elements from the global memory, reduce them together and
+     * store them in the local memory */
+    {
+      index_t elem_col = global_col;
+      while (elem_col < cols_) {
```
AdamHarries: Have you benchmarked replacing this with a for loop? You indicated that other loops got good performance by replacing while with for.

src/executors/executor_sycl.hpp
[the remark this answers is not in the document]
Nyrio: It is never required. It is absolutely optional, in contrary to algorithms that only use a parallel/tree reduction. In this reduction algorithm, the first part loops until the end of each row with a stride that depends on the size of work groups and the number of work groups. Doing the algorithm in only one step means that there is only one group on the columns dimension, the algorithm handles it well.
And after some experimentation doing a two-step reduction doesn't seem worth the overhead of launching two kernels. We could use heuristics to check whether the number of rows is very very small (it is not a use case that we have at the moment)

src/operations/extension/reduction_partial_rows.hpp
```diff
@@ -0,0 +1,237 @@
...
+   * number of work groups required to execute each partial reduction step.
+   */
+  SYCL_BLAS_INLINE index_t get_workgroup_cluster() noexcept {
+    return ((rows_ - 1) / local_memory_rows + 1) * group_count_cols_;
+  }
+  /*!
+   * @brief get_num_workgroup_cluster. This function is used to extend the
+   * number of work group clusters, in order to make sure that at least 4
+   * operations are available per work group. The number 4 is used based on
+   * empirical research.
```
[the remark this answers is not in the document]
Nyrio: Same question as Ross: where would you put it? In this function, at the top of the file, in a common location somewhere, or even as a template parameter for per-platform control?

src/operations/extension/reduction_partial_rows.hpp
Nyrio: It makes absolutely no difference (this kernel is very memory-bound anyway) but I will make the change because a `for` loop is more suited here

src/executors/executor_sycl.hpp
```diff
@@ -230,6 +230,91 @@ Executor<PolicyHandler<codeplay_policy>>::execute(
...
+  // At the moment the one-step proved to be better in almost every case
```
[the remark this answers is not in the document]
Nyrio: The minimum number of columns for 2-step to be worth it is >2048. The number of rows doesn't really matter. I've updated with heuristics instead of a hard-coded value (our current use case, tsgemm, will always use 1-step though).

src/operations/extension/reduction_partial_rows.hpp
```diff
@@ -0,0 +1,237 @@
...
+                                 get_num_workgroup_cluster(compute_units));
+    const cl::sycl::range<1> wgs(work_group_size);
+    return cl::sycl::nd_range<1>(nwg * wgs, wgs);
+  }
+
+  template <typename local_memory_t>
+  SYCL_BLAS_INLINE void eval(local_memory_t scratch,
+                             cl::sycl::nd_item<1> id) noexcept {
+    /* reference to the scratch memory */
+    auto scratch_ptr = scratch.localAcc.get_pointer().get();
```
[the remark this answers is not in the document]
Nyrio: I've updated the line where the pointer is obtained.
Concerning bank conflicts, the only case where it can be an issue is for a very small tile at the last steps of the parallel reduction. Even in that rare case it would not affect the performance significantly

Nyrio:
I removed the multiple workload per thread, and got good speedups (it was a bad idea and I didn't do it right).

src/operations/extension/reduction_partial_rows.hpp
Nyrio: I have addressed the issue with a constexpr variable at the moment. We can later think about making it a template parameter if we want per-platform control.

src/executors/executor_sycl.hpp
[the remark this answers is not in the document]
Nyrio: I have renamed the variable `two_step_reduction`, it's a bit more explicit

src/operations/extension/reduction_partial_rows.hpp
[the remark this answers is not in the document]
Nyrio: Done (cf comment on the other discussion above)

PR #169, opened 2019-06-26
Constant refactoring

AdamHarries:
This PR significantly refactors the binary/unary operation definitions, along with theconstant generation machinery. Additionally, it also refactors the `IndexValueTuple` type so that template types are passed in the same order as in the the constructor.

### DIFF
--- a/benchmark/syclblas/blas1/iamax.cpp
+++ b/benchmark/syclblas/blas1/iamax.cpp
@@ -46,21 +46,21 @@ void run(benchmark::State& state, ExecutorType* executorPtr, index_t size,
 
   // Create data
   std::vector<scalar_t> v1 = blas_benchmark::utils::random_data<scalar_t>(size);
-  blas::IndexValueTuple<scalar_t, index_t> out(-1, -1);
+  blas::IndexValueTuple<index_t, scalar_t> out(-1, 0);
 
   auto inx = blas::make_sycl_iterator_buffer<scalar_t>(v1, size);
   auto outI =
-      blas::make_sycl_iterator_buffer<blas::IndexValueTuple<scalar_t, index_t>>(
+      blas::make_sycl_iterator_buffer<blas::IndexValueTuple<index_t, scalar_t>>(
           &out, 1);
 
 #ifdef BLAS_VERIFY_BENCHMARK
   // Run a first time with a verification of the results
   index_t idx_ref =
       static_cast<index_t>(reference_blas::iamax(size, v1.data(), 1));
-  blas::IndexValueTuple<scalar_t, index_t> idx_temp(-1, -1);
+  blas::IndexValueTuple<index_t, scalar_t> idx_temp(-1, 0);
   {
     auto idx_temp_gpu =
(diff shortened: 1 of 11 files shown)


include/operations/blas_constants.h
```diff
@@ -58,71 +60,66 @@ enum class const_val : int {
...
-template <typename value_t, typename index_t>
-struct constant<IndexValueTuple<value_t, index_t>, const_val::imin> {
-  constexpr static SYCL_BLAS_INLINE IndexValueTuple<value_t, index_t> value() {
-    return IndexValueTuple<value_t, index_t>(
-        std::numeric_limits<index_t>::max(),
-        std::numeric_limits<value_t>::max());
+template <typename value_t>
+struct constant<value_t, const_val::abs_max> {
+  constexpr static SYCL_BLAS_INLINE value_t value() {
+    return std::numeric_limits<value_t>::max();
```
Nyrio: As discussed offline, it might make sense to combine that with the specialization for `max`, which I think could be done with `enable_if`, if it doesn't make the code less readable.

benchmark/syclblas/blas1/iamax.cpp
```diff
@@ -46,21 +46,21 @@ void run(benchmark::State& state, ExecutorType* executorPtr, index_t size,
 
   // Create data
   std::vector<scalar_t> v1 = blas_benchmark::utils::random_data<scalar_t>(size);
-  blas::IndexValueTuple<scalar_t, index_t> out(-1, -1);
+  blas::IndexValueTuple<index_t, scalar_t> out(-1, 0);
```
Nyrio: Is there a way to create this from the constants without including the operators hpp file? The following is a bit verbose:
```
auto out = constant_pair<IndexValueTuple<index_t, scalar_t>, const_val::m_one, const_val::zero>::value();
```

I thought we would also have something simple like:
```
auto out = constant<IndexValueTuple<index_t, scalar_t>, const_val::amin>::value();
```

mehdi-goli: No no need. The constant and operator is pure internal code. So no need to use it in user code.

src/policy/sycl_policy_handler.cpp
```diff
@@ -128,12 +128,12 @@ INSTANTIATE_TEMPLATE_METHODS(double)
...
 
-INSTANTIATE_TEMPLATE_METHODS_SPECIAL(float, int)
-INSTANTIATE_TEMPLATE_METHODS_SPECIAL(float, long)
-INSTANTIATE_TEMPLATE_METHODS_SPECIAL(float, long long)
-INSTANTIATE_TEMPLATE_METHODS_SPECIAL(double, int)
-INSTANTIATE_TEMPLATE_METHODS_SPECIAL(double, long)
-INSTANTIATE_TEMPLATE_METHODS_SPECIAL(double, long long)
+INSTANTIATE_TEMPLATE_METHODS_SPECIAL(int, float)
+INSTANTIATE_TEMPLATE_METHODS_SPECIAL(long, float)
+INSTANTIATE_TEMPLATE_METHODS_SPECIAL(long long, float)
```
mehdi-goli: can you remove this line

PR #171, opened 2019-07-02
Added 2D versions of the operations: `Scalar2DOp`, `Unary2DOp`, `Binary2DOp`, `Assign2D`

Nyrio:
I haven't created expression tests for the individual operations but used them successfully in tall & skinny gemm. Let me know if you think they should have expression tests.

These operations work very similarly to their 1D counterparts but add a notion of leading dimension and two functions for the size of rows and columns. Basically the scalar and unary operations are quite transparent and return whatever the RHS returns, except that `eval` performs the scalar or unary operations. The binary and assign operations are a bit more complex because the lhs and rhs can have different leading dimensions, so it does the index calculation and the leading dimension of the operator will be the number of rows.

Note: calling `eval` on the scalar or unary operations must take into account the leading dim, because they behave like the matrix view 1-parameter `eval` function. Technically all the operations do but the binary and assign operations have leading dim = nb rows. I could change that behavior to always return a leading dimension equal to the number of rows and remap that manually in the `eval` functions, but it would have a greater performance cost.

### DIFF
--- a/include/operations/blas2_trees.h
+++ b/include/operations/blas2_trees.h
@@ -293,7 +293,7 @@ struct Unary2DOp {
 };
 
 /*! Binary2DOp.
- * @brief Implements a Binary Operation (x OP z) with x and z vectors.
+ * @brief Implements a Binary Operation (A OP B) with A and B matrices.
  */
 template <typename operator_t, typename lhs_t, typename rhs_t>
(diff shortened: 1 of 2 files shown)


Nyrio:
Closing as this is useless after #174

PR #173, opened 2019-07-05
Reduction benchmark

Nyrio:


benchmark/syclblas/expression/reduction_rows.cpp
```diff
@@ -0,0 +1,152 @@
...
+      reduction(buffer_in, buffer_out, rows, cols);
+  return ex.execute(reduction);
+}
+
+template <typename scalar_t>
+void run(benchmark::State& state, ExecutorType* executorPtr, index_t rows,
+         index_t cols, bool* success) {
+  // The counters are double. We convert m, n and k to double to avoid integer
+  // overflows for n_fl_ops and bytes_processed
+  double rows_d = static_cast<double>(rows);
```
CodeplayBen: Is it possible to use an unsigned integer type of suitable width here, for example uint64_t? Double seems an odd choice when floating point is not required.

Edit: Also I don't know how relevant it is but perhaps useful on platforms which don't support doubles? Had an error about this on rcar boards today which brought it to mind but again not sure how relevant a problem that is, or if platforms that don't support doubles are likely to support uint64_t.

src/executors/executor_sycl.hpp
```diff
@@ -26,6 +26,8 @@
 #ifndef SYCL_BLAS_EXECUTOR_SYCL_HPP
 #define SYCL_BLAS_EXECUTOR_SYCL_HPP
 
+#include <algorithm>
```
CodeplayBen: Is it worth including the algorithm header just for the single usage of std::min? I was under the impression that it is quite heavyweight.

Nyrio: You're right that it might affect a bit the compilation time, I fix that in the reduction PR since this code belongs there (this one is only the benchmark, as said in the description the changes from #167 are only here for convenience and will disappear once the other is merged).

benchmark/syclblas/expression/reduction_rows.cpp
Nyrio: I know this is odd, but the Google Benchmark framework only supports `double` for the counters. I could have used a static cast to a `uint64_t` and then when assigning the counters there would be an implicit cast to `double`, but doing it that way seems better to me because it explicitly shows that the counters are double.

Concerning the platforms that don't support doubles, it's just that doubles have the size of a float. Using floats for the calculation of the counters is fine, the precision is good enough.

Nyrio: And also, just wondering: on RCar does the CPU not support doubles? I think it's only the accelerator.

CodeplayBen: Ah OK, that makes sense then! I'm not sure about the RCar CPU as I'd just seen the error in some of my own work, but you are probably right it is just the accelerator rather than the CPU. I'll resolve this now then :)

Nyrio:
@AdamHarries sorry for that. I have merged the other branch into this one as rebasing is destructive.

PR #174, opened 2019-07-11
Updating views to make matrices usable in expression trees without 2d operators

Nyrio:


src/views/view_sycl.hpp
```diff
@@ -143,22 +143,39 @@ struct VectorView<
...
+  template <bool use_as_ptr = false>
+  SYCL_BLAS_INLINE typename std::enable_if<!use_as_ptr, scalar_t &>::type eval(
+      index_t i) {
     return (strd_ == 1) ? *(ptr_ + i) : *(ptr_ + i * strd_);
   }
-  /**** EVALUATING ****/
-  SYCL_BLAS_INLINE scalar_t &eval(index_t i) {
+
+  template <bool use_as_ptr = false>
+  SYCL_BLAS_INLINE typename std::enable_if<!use_as_ptr, const scalar_t>::type
```
CodeplayBen: Does this return type need to be const? It's a copy either way so the const seems superfluous here unless required.

Nyrio: It probably doesn't, I did that for consistency with the existing code. @mehdi-goli any opinion on this?

mehdi-goli: As it is a copy const qualifier has no effect here, however, the function labeled as const, if it does not give you an error when removing the const, then remove it

Nyrio: Fixed, I've removed the useless `const` qualifiers in the return type. The function **has to** be labeled as `const` though because 1) otherwise it doesn't compile as there would be two overloads that differ only in their return type ; 2) it is good to label a method as `const` if it doesn't modify the object state.

src/views/view_sycl.hpp
```diff
@@ -232,32 +249,60 @@ struct MatrixView<
...
     return eval(ndItem.get_global_id(0));
   }
 
-  SYCL_BLAS_INLINE const scalar_t eval(cl::sycl::nd_item<1> ndItem) const
+  SYCL_BLAS_INLINE scalar_t eval(cl::sycl::nd_item<1> ndItem) const
       noexcept {
     return eval(ndItem.get_global_id(0));
   }
 
+  template <bool use_as_ptr = false>
```
RossBrunton: Should these not be moved above so they are below their non-enable_if versions?

Nyrio: The non `enable_if` versions are a different overload. The one with `enable_if` has `enable_if` both for the `true` and `false` cases

PR #176, opened 2019-07-16
Optimize the AMD GPU Gemm parameterization backend

Nyrio:


CodeplayBen:
Does this need rebased on top of master or the commits squashed? That seems like a lot of commits to add to the history.

Nyrio:
@CodeplayBen I have merged master on it, but not rebased (rebasing would be very painful here). The commits come from the tsgemm branch, squashing should be fine, like we always do

PR #178, opened 2019-07-22
Make the tsgemm kernel usable as Gemm and select the best Gemm algorithm based on empirical heuristics

Nyrio:
Only enabled for Intel GPU at the moment.
Note: this branch contains some huge optimizations that will also benefit tsgemm for tall and skinny matrices


include/operations/blas3_trees.h
```diff
@@ -42,7 +42,12 @@ enum class gemm_memory_t : int { local = 0, no_local = 1 };
  * It can be either naive to use a naive algorithm, classic for the default
  * algorithms, or tall_skinny for tall and skinny matrices
  */
-enum class gemm_algorithm_t : int { naive = 0, classic = 1, tall_skinny = 2 };
+enum class gemm_algorithm_t : int {
+  naive = 0,
+  classic = 1,
+  tall_skinny = 2,
+  alternative = 3
```
mehdi-goli: what does alternative means?
do you call normal gemm (non-skinny) classic? classic gemm algorithm is the name chosen to call the 3 nested for loop doing the matmul. when you call the optimize gemm for non-skinny classic, it can be confusing

src/operations/blas3/gemm_partial_local.hpp
```diff
@@ -177,8 +223,8 @@ class GemmPartial<input_t, output_t, DoubleBuffer, NbcA, NbcB, ClSize,
   SYCL_BLAS_INLINE void eval(local_memory_t scratch,
                              cl::sycl::nd_item<1> id) noexcept {
     /* Pointers to the scratch memory (lhs and rhs) */
-    auto scratch_ptr = scratch.localAcc.get_pointer().get();
-    auto rhs_scratch_ptr = scratch_ptr + rhs_scratch_offset;
+    value_t* scratch_ptr = scratch.localAcc.get_pointer().get();
```
mehdi-goli: you don't need .get() at the end anymore. ComputeCpp implicitly convert multi_pointer to the pointer * since you remove the auto

src/operations/blas3/gemm_partial_local.hpp
```diff
@@ -285,160 +339,122 @@ class GemmPartial<input_t, output_t, DoubleBuffer, NbcA, NbcB, ClSize,
...
-                                    index_t global_m_offset,
-                                    index_t global_n_offset,
-                                    index_t global_k_offset) {
+  SYCL_BLAS_INLINE void load_blocks(
+      const index_t& lhs_row, const index_t& lhs_col, const index_t& rhs_row,
+      const index_t& rhs_col, const index_t& tile_idx, local_ptr_t scratch_ptr,
+      local_ptr_t rhs_scratch_ptr, const index_t& global_m_offset,
+      const index_t& global_n_offset, const index_t& global_k_offset) {
     // LHS tile
     if (TransA) {
```
mehdi-goli: if you move `if(TransA )` and `if(TransB)` to the load_block it reduces the size of code and make the code more readable. So TransA or TransB will be passed as a template parameter to loadblock and loadblock will check if(trans)

include/operations/blas3_trees.h
```diff
@@ -42,7 +42,12 @@ enum class gemm_memory_t : int { local = 0, no_local = 1 };
...
+};
```
RossBrunton: "classic", "alternative"? Could we have clearer names?

Nyrio: Sure. Do you have something in mind?

include/operations/blas3_trees.h
Nyrio: The 3 nested loops is the `naive` algorithm. Anyway, as answered to Ross I can change that but I can't think of better names, so do you have something in mind?

src/interface/blas3/backend/intel_gpu.hpp
```diff
@@ -92,32 +92,31 @@ typename executor_t::policy_t::event_t _gemm(
...
+  } else if (_M <= 128 && _N <= 128) {
     return blas::Gemm_Launcher<
         64, true, false, false, 64, Tile<4, 4, 8, 8>, _t_a, _t_b,
         static_cast<int>(gemm_memory_t::local),
         static_cast<int>(gemm_algorithm_t::classic),
         is_beta_zero>::template _select_gemm(ex, _M, _N, _K, _alpha, _a, _lda,
                                              _b, _ldb, _beta, _c, _ldc,
                                              batch_size);
-  } else if (!_t_a) {
+  } else if (_t_b && !_t_a) {
```
CodeplayBen: Could this be changed to _t_b != _t_a or is the performance significantly different when b is not transposed and a is?

Nyrio: It really can't. The algorithm without local memory is better than the algorithm with local memory only in the case when both of the matrices are naturally accessed in a coalesced pattern. With column-major matrices this is when B is transposed and A isn't.

Changing that to `_t_b != _t_a` would give awful performance in the case where A is transposed and B isn't (non-coalesced access in both).

CodeplayBen: OK no worries, thanks for clarifying that!

include/operations/blas3_trees.h
mehdi-goli: why do you need separate name. you tsgem is a gemm with numkgroup==1. and you should use the same gemm. you dont need to name it separately

Nyrio: As discussed offline, `alternative` has been integrated in `tall_skinny` with simple heuristics to make it final. `classic` has been renamed to `standard`

include/operations/blas3_trees.h
Nyrio: Discussed offline with Mehdi and Adam. `alternative` was swallowed by `tall_skinny` and `classic` renamed to `standard` to avoid confusion with the `naive` algorithm which is sometimes called classic in the literature

src/operations/blas3/gemm_partial_local.hpp
Nyrio: Done.

PR #183, opened 2019-08-08
Update documentation and add code samples

Nyrio:


samples/gemv.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+  /* Execute the GEMM operation */
+  std::cout << "---\nExecuting Y = " << alpha << "*A*X + " << beta << "*Y\n";
+  {
+    auto a_gpu = blas::make_sycl_iterator_buffer<float>(A, lda * n);
+    auto x_gpu = blas::make_sycl_iterator_buffer<float>(X, lx);
+    auto y_gpu = blas::make_sycl_iterator_buffer<float>(Y, ly);
+    auto event = blas::_gemv(executor, 'n', m, n, alpha, a_gpu, lda, x_gpu,
+                             incx, beta, y_gpu, incy);
+    policy_handler.wait(event);
```
mehdi-goli: you don't need this wait if you are going out of scope

Nyrio: That's right. Fixed

PR #186, opened 2019-08-12
New gemv implementation

przemektmalon:
This PR implements a new, faster, and clearer, GEMV kernel. The GEMV class has a `work_per_thread` template parameter which is not used, but is there for future implementation.

The `blas2_gemv_test.h` was also changed to make it clearer and work properly with the new implementation.

### DIFF
--- a/include/blas_meta.h
+++ b/include/blas_meta.h
@@ -120,6 +120,11 @@ static inline index_t get_power_of_two(index_t wGSize, bool rounUp) {
 
 #define SYCL_BLAS_INLINE inline __attribute__((always_inline))
 
+template<typename index_t>
+static  SYCL_BLAS_INLINE index_t roundUp(index_t x, index_t y) {
+  return ((x + y - 1) / y) * y;
+}
+
 template <typename index_t, typename vector_t>
 index_t vec_total_size(index_t &vector_size, vector_t &&current_vector) {
   vector_size += static_cast<index_t>(current_vector.size());
--- a/include/operations/blas2_trees.h
+++ b/include/operations/blas2_trees.h
@@ -26,37 +26,91 @@
 #ifndef SYCL_BLAS_BLAS2_TREES_H
 #define SYCL_BLAS_BLAS2_TREES_H
 namespace blas {
-template <typename output_t, typename matrix_t, typename vector_t>
+
+/*!
+ * @brief Determines the memory type of the GEMV kernel.
+ * It can either use local memory or not
+ */
+enum class gemv_memory_t : int { local = 0, no_local = 1 };
+
+/*!
+ * @brief Gemv is a templated class whose instantiations provide different
+ * implementations of the the GEMV kernel function.
+ *
+ * The class is constructed using the make_Gemv function below.
+ *
+ * @tparam local_range  specifies the number of threads per work group used by
+ *                      the kernel
+ * @tparam is_transposed  specifies whether the input matrix should be
+ * transposed
+ * @tparam cache_line_size  specifies the size in bytes of the cache line. This
+ *                          value will determine the dimensions of tiles loaded
+ *                          into local memory in the transposed local memory
+ *                          version of the kernel
+ * @tparam memory_type  specifies whether the kernel should use local shared
+ *                      memory or not
+ * @tparam work_per_thread  (not implemented) would specify the multiplier of
+ *                          work done per each work item
+ * @param lhs_        the output buffer of the kernel
+ * @param matrix_a_   the input matrix a
+ * @param vector_x_   the input vector x
+ * @param wgs_per_nc  the number of work groups per non-contracting dimension
+ * @param wgs_per_c   the number of work groups per contracting dimension
+ *
+ */
+template <typename lhs_t, typename matrix_t, typename vector_t,
+          uint32_t local_range, bool is_transposed, int cache_line_size,
+          int work_per_thread>
 struct Gemv {
   using value_t = typename vector_t::value_t;
   using index_t = typename vector_t::index_t;
-  output_t lhs_;
-  matrix_t matrix_;
-  vector_t vector_;
+  lhs_t lhs_;
+  matrix_t matrix_a_;
+  vector_t vector_x_;
+  index_t wgs_per_nc_;
+  index_t wgs_per_c_;
 
-  Gemv(output_t &_l, matrix_t &_matrix, vector_t &_vector);
-  index_t get_size() const;
+  Gemv(lhs_t &_l, matrix_t &_matrix, vector_t &_vector, index_t &_wgs_per_nc,
+       index_t &_wgs_per_c);
   bool valid_thread(cl::sycl::nd_item<1> ndItem) const;
-  value_t eval(index_t i);
   value_t eval(cl::sycl::nd_item<1> ndItem);
+  template <typename local_memory_t>
+  value_t eval(local_memory_t local_mem, cl::sycl::nd_item<1> ndItem);
   void bind(cl::sycl::handler &h);
   void adjust_access_displacement();
+
+ private:
+  template <typename ScratchPointerType>
+  void extract_input_block(ScratchPointerType scratch, const index_t &local_id,
+                           const index_t &group_id, const index_t &lda,
+                           index_t mat_tile_id);
 };
 
-template <typename output_t, typename matrix_t, typename vector_t>
-Gemv<output_t, matrix_t, vector_t> make_gemv(output_t &lhs_, matrix_t &matrix_,
-                                             vector_t &vector_) {
-  return Gemv<output_t, matrix_t, vector_t>(lhs_, matrix_, vector_);
+/*!
+ * @brief Contructs an instance of the Gemv class used to launch Gemv operation
+ * kernels
+ */
+template <uint32_t local_range, bool is_transposed, int cache_line_size,
+          int work_per_thread, typename lhs_t, typename matrix_t,
+          typename vector_t>
+Gemv<lhs_t, matrix_t, vector_t, local_range, is_transposed, cache_line_size,
+     work_per_thread>
+make_Gemv(lhs_t &lhs_, matrix_t &matrix_, vector_t &vector_,
+          typename vector_t::index_t wgs_per_nc_,
+          typename vector_t::index_t wgs_per_c_) {
+  return Gemv<lhs_t, matrix_t, vector_t, local_range, is_transposed,
+              cache_line_size, work_per_thread>(lhs_, matrix_, vector_,
+                                                wgs_per_nc_, wgs_per_c_);
 }
 
 template <typename rhs_t>
-struct AddSetColumns {
+struct SumMatrixColumns {
   using value_t = typename rhs_t::value_t;
   using index_t = typename rhs_t::index_t;
 
   rhs_t rhs_;
 
-  AddSetColumns(rhs_t &_r);
+  SumMatrixColumns(rhs_t &_r);
 
   index_t get_size() const;
 
@@ -70,12 +124,12 @@ struct AddSetColumns {
 };
 
 /**
- * @fn make_addSetColumns
- * @brief Constructs an AddSetColumns structure.
+ * @fn make_sumMatrixColumns
+ * @brief Constructs a SumMatrixColumns structure.
  */
 template <typename rhs_t>
-AddSetColumns<rhs_t> make_addSetColumns(rhs_t &rhs_) {
-  return AddSetColumns<rhs_t>(rhs_);
+SumMatrixColumns<rhs_t> make_sumMatrixColumns(rhs_t &rhs_) {
+  return SumMatrixColumns<rhs_t>(rhs_);
 }
 
 /**
--- a/src/interface/blas2_interface.hpp
+++ b/src/interface/blas2_interface.hpp
@@ -29,6 +29,7 @@
 #include "blas_meta.h"
 #include "container/sycl_iterator.h"
 #include "executors/executor.h"
+#include "interface/blas2/backend/backend.hpp"
 #include "interface/blas2_interface.h"
 #include "operations/blas2_trees.h"
 #include "operations/blas_constants.h"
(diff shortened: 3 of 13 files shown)


src/operations/blas2/gemv.hpp
```diff
@@ -80,6 +80,244 @@ SYCL_BLAS_INLINE void AddSetColumns<rhs_t>::adjust_access_displacement() {
...
+SYCL_BLAS_INLINE
+void
+Gemv<lhs_t, matrix_t, vector_t, is_transposed, cache_line_size, work_per_thread>::extract_input_block(ScratchPointerType matrix_scratch, cl::sycl::nd_item<1> id, index_t c_tile_id) {
+
+  constexpr int cache_capacity = cache_line_size / sizeof(value_t);
+
+  const index_t local_id = id.get_local_id(0);
+  const index_t group_id = id.get_group(0);
+  const index_t local_range = id.get_local_range(0);
+  const index_t lda = matrix_.getSizeL();
```
CodeplayBen: Seems like these could be passed in as parameters as the values are already obtained inside eval() where this function is called.

src/operations/blas2/gemv.hpp
```diff
@@ -80,6 +80,244 @@ SYCL_BLAS_INLINE void AddSetColumns<rhs_t>::adjust_access_displacement() {
...
+
+  // Number of iterations to load all tiles
+  // local_range / tile_dim_nc == cache_capacity
+  constexpr int tile_nc_loops = cache_capacity;
+
+  #pragma unroll
+  for (index_t nc_tile_id = 0; nc_tile_id < tile_nc_loops; ++nc_tile_id) {
+
+    // Bank offset adds 1 word of padding every 'cache_capacity' elements
+    const index_t bank_offset = tile_c_index;
```
CodeplayBen: This could also be moved out of the for loop, or just use the tile_c_index directly.

CodeplayBen:
I've also just noticed your fork is a little bit behind on master, I think it would be worth updating your fork and making sure that nothing is broken before merging.

src/interface/blas2_interface.hpp
```diff
@@ -55,67 +55,85 @@ typename Executor::policy_t::event_t _gemv_impl(
...
   typename Executor::policy_t::event_t ret{};
 
-  index_t M = (trn == transpose_type::Normal) ? _M : _N;
-  index_t N = (trn == transpose_type::Normal) ? _N : _M;
+  constexpr int cache_line_size = 64;
+  constexpr int cache_capacity = cache_line_size / sizeof(element_t);
+  constexpr bool is_transposed = trn != transpose_type::Normal;
+
+  index_t local_size = (_localSize == 0)
+                                ? (!is_transposed
```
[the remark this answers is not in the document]
przemektmalon: Non-transposed is a default in my mind so I usually put the code for that first in ternaries. I don't think it impacts the readability at all to be honest but I can change it if people disagree.

src/interface/blas2_interface.hpp
```diff
@@ -55,67 +55,85 @@ typename Executor::policy_t::event_t _gemv_impl(
...
+                                : _localSize;
 
-  static constexpr auto data_layout_access =
-      Choose<trn == transpose_type::Normal, access_layout,
-             access_layout::col_major, access_layout::row_major>::type;
-  using data_layout_t = typename Layout<data_layout_access>::type;
-  auto mA = make_matrix_view<data_layout_t>(ex, _mA, M, N, _lda);
-  auto vx = make_vector_view(ex, _vx, _incx, N);
-  auto vy = make_vector_view(ex, _vy, _incy, M);
+  const auto x_vector_size = !is_transposed ? _N : _M;
```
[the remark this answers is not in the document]
przemektmalon: Again, non-transposed feels default for me so hence the negation and first value of ternary being the default. Negation doesn't cost anything since its a `constexpr`. Readability is not affected in my opinion, but can change if it bothers people that much haha

src/interface/blas2_interface.hpp
```diff
@@ -55,67 +55,85 @@ typename Executor::policy_t::event_t _gemv_impl(
...
-        ret, ex.execute(gemvC, localSize, globalSize, scratchPadSize));
-  } else {
-    auto gemvR = make_Gemv_Row<interLoop>(mat1, mA, vx, nWGPerRow, nWGPerCol,
-                                          scratchPadSize);
-    ret = concatenate_vectors(
-        ret, ex.execute(gemvR, localSize, globalSize, scratchPadSize));
-  }
+  // Leading dimension for partial dot products matrix
+  auto ld = !is_transposed ? _M : _N;
+  auto dot_products_size = ld * output_scratch_size;
```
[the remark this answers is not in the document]
przemektmalon: Made const. And negation discussion as in previous comments

src/interface/blas2_interface.hpp
```diff
@@ -55,67 +55,85 @@ typename Executor::policy_t::event_t _gemv_impl(
...
+  auto dot_products_buffer = 
+        blas::make_sycl_iterator_buffer<element_t>(dot_products_size);
+  auto dot_products_matrix = 
+        make_matrix_view<col_major>(ex, dot_products_buffer, ld, output_scratch_size, ld);
+
+  // Global size for shared-mem-kernel
+  const index_t global_size = local_size * WGs_per_C * WGs_per_NC;
+
+  // Global size for no-shared-mem kernel
+  //const index_t global_size = roundUp(dot_products_size, local_size);
```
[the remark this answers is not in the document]
przemektmalon: Commented code is for calling the non-shared memory version of the kernel. Since the interface function provides no way to the user to chose between shared and non-shared memory kernels, I left the code in there to show how to set the global size for the non-shared memory kernel correctly, until the interface to the function changes to allow calling non-shared-memory kernel, and the code can be used.

I can remove it. But if at any point in the future we want to be able to call the non-shared-memory kernel version, I'm just making it easier for the person writing the code to set the global size.

src/operations/blas2/gemv.hpp
```diff
@@ -80,6 +80,244 @@ SYCL_BLAS_INLINE void AddSetColumns<rhs_t>::adjust_access_displacement() {
...
+  const index_t group_id = ndItem.get_group(0);
+  const index_t group_range = ndItem.get_group_range(0);
+  const index_t lda = matrix_.getSizeL();
+  const index_t nc_dim = is_transposed ? matrix_.get_size_col() : matrix_.get_size_row();
+  const index_t c_dim = is_transposed ? matrix_.get_size_row() : matrix_.get_size_col();
+
+  // ID of the group in the non-contracting dimension of the global grid of WGs
+  const index_t nc_group_id = group_id / WGs_per_C_;
+  // ID of the group in the contracting dimension in the global grid of WGs
+  const index_t c_group_id = group_id % WGs_per_C_;
```
Nyrio: Generally try to avoid the modulo in kernel code (except for compile-time expressions). This can be replaced with:
```
c_group_id = group_id - nc_group_id * WGs_per_C_
```

src/operations/blas2/gemv.hpp
```diff
@@ -80,6 +80,244 @@ SYCL_BLAS_INLINE void AddSetColumns<rhs_t>::adjust_access_displacement() {
...
+template <typename lhs_t, typename matrix_t, typename vector_t, bool is_transposed, int cache_line_size, int work_per_thread>
+SYCL_BLAS_INLINE
+Gemv<lhs_t, matrix_t, vector_t, is_transposed, cache_line_size, work_per_thread>::
+      Gemv(lhs_t &_l, matrix_t &_mat, vector_t &_vec,
+            typename vector_t::index_t &_WGs_per_NC,
+            typename vector_t::index_t &_WGs_per_C,
+            typename vector_t::index_t &_local_memory_size)
+      : lhs_(_l), // Result is stored in this
+        matrix_(_mat), // Input matrix
+        vector_(_vec), // Input vector
```
Nyrio: Given that the name `x` still appear e.g in index variable names, maybe it would make sense to call the vector `x_` or `x_vec`, and same for the matrix `A_`? For `y` it's not really why since it's an intermediate buffer so `lhs_` works

src/operations/blas2/gemv.hpp
```diff
@@ -80,6 +80,244 @@ SYCL_BLAS_INLINE void AddSetColumns<rhs_t>::adjust_access_displacement() {
...
+
+    for (index_t c_tile_id = 0; c_tile_id < tile_c_loops; ++c_tile_id) {
+
+      // Extract a matrix block from global memory
+      extract_input_block(matrix_scratch, ndItem, c_tile_id);
+
+      // Ensure memory synchronization within work group
+      ndItem.barrier(cl::sycl::access::fence_space::local_space);
+
+      index_t mat_index = local_id;
```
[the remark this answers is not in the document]
przemektmalon: The declaration could be moved out, but the assignment cant, `mat_index` has to be reset on each iteration of the outer loop since it is in local memory, which gets updated on each iteration of the outer loop.

src/operations/blas2/gemv.hpp
```diff
@@ -80,6 +80,244 @@ SYCL_BLAS_INLINE void AddSetColumns<rhs_t>::adjust_access_displacement() {
...
+    }
+
+    // Calculate the matrix index
+    index_t mat_index = nc_dim_index + c_group_id * local_range * lda;
+    
+    const index_t last_c_dim_id = std::min(c_dim - c_group_id * local_range, local_range);
+
+    // Computes the partial dot product for a row
+    for (index_t c_dim_id = 0; c_dim_id < last_c_dim_id; ++c_dim_id) {
+      sum += ProductOperator::eval(matrix_.template eval<true>(mat_index), local_mem[c_dim_id]);
```
Nyrio: Replace that with a `cl::sycl::mad`

src/operations/blas2/gemv.hpp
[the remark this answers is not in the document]
CodeplayBen: Ah that is my mistake, I missed it being modified in the inner loop!

src/interface/blas2_interface.hpp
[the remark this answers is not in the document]
Nyrio: I agree with Ben here. I have to revert it in my head to read it and I get confused

src/operations/blas2/gemv.hpp
```diff
@@ -80,6 +80,244 @@ SYCL_BLAS_INLINE void AddSetColumns<rhs_t>::adjust_access_displacement() {
...
+
+      // Ensure memory synchronization within work group
+      ndItem.barrier(cl::sycl::access::fence_space::local_space);
+
+      index_t mat_index = local_id;
+
+      #pragma unroll
+      for (index_t c_dim_id = 0; c_dim_id < cache_capacity; ++c_dim_id) {
+
+        sum += ProductOperator::eval(matrix_scratch[mat_index], * vector_scratch++);
```
Nyrio: `cl::sycl::mad`

src/operations/blas2/gemv.hpp
przemektmalon: Done for `local_id`, `group_id`, and `local_range`. Don't think it's necessary for `lda` since `matrix_` is a member and `getSizeL()` is inlined ?

include/operations/blas2_trees.h
```diff
@@ -26,27 +26,41 @@
...
-Gemv<output_t, matrix_t, vector_t> make_gemv(output_t &lhs_, matrix_t &matrix_,
-                                             vector_t &vector_) {
-  return Gemv<output_t, matrix_t, vector_t>(lhs_, matrix_, vector_);
+template <bool is_transposed, int cache_line_size, int work_per_thread, typename lhs_t, typename matrix_t, typename vector_t>
+Gemv<lhs_t, matrix_t, vector_t, is_transposed, cache_line_size, work_per_thread> make_Gemv(lhs_t &lhs_, matrix_t &matrix_,
+    vector_t &vector_, 
+    typename vector_t::index_t WGs_per_NC_, 
+    typename vector_t::index_t WGs_per_C_,
+    typename vector_t::index_t local_memory_size_) {
+  return Gemv<lhs_t, matrix_t, vector_t, is_transposed, cache_line_size, work_per_thread>(lhs_, matrix_, vector_, WGs_per_NC_, WGs_per_C_, local_memory_size_);
```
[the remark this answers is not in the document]
przemektmalon: I always forget. I'll format it.

src/operations/blas2/gemv.hpp
CodeplayBen: That is true, though I would probably replace it anyway just to avoid the duplication, I'll leave it up to you though!

src/operations/blas2/gemv.hpp
przemektmalon: Updated

src/interface/blas2_interface.hpp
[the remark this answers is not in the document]
przemektmalon: Ok, I'll change all the uses of `is_transposed` to be non-negated.

src/interface/blas2_interface.hpp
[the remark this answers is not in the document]
RossBrunton: I'd preface the comment with `// If you want to call the non-shared-memory version, do this`

src/operations/blas2/gemv.hpp
przemektmalon: I've made it a parameter. Though I'm still unsure of which option would result in more performant code. Either way the difference should be miniscule.

include/blas_meta.h
```diff
@@ -120,6 +120,11 @@ static inline index_t get_power_of_two(index_t wGSize, bool rounUp) {
 
 #define SYCL_BLAS_INLINE inline __attribute__((always_inline))
 
+template<typename index_t>
+static  SYCL_BLAS_INLINE index_t roundUp(index_t x, index_t y) {
+  return ((((x) + (y)-1) / (y)) * (y));
```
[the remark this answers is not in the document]
przemektmalon: Mehdi copied this code from somewhere, either another project or an older version of this one.

It's used in calculating the global size for the non-shared-memory version of the kernel. But since the user cannot specify which version to use, I made shared-memory the default and left the other global size calculation commented out. I don't know if I should remove it, but if we want to add the possibility of using the non-shared-memory kernel in the future then this function would be used and the commented out code in the interface file.

src/interface/blas2_interface.hpp
```diff
@@ -55,67 +55,85 @@ typename Executor::policy_t::event_t _gemv_impl(
...
-  index_t N = (trn == transpose_type::Normal) ? _N : _M;
+  constexpr int cache_line_size = 64;
+  constexpr int cache_capacity = cache_line_size / sizeof(element_t);
+  constexpr bool is_transposed = trn != transpose_type::Normal;
+
+  index_t local_size = (_localSize == 0)
+                                ? (!is_transposed 
+                                    ? ex.get_policy_handler().get_work_group_size()
+                                    : ex.get_policy_handler().get_work_group_size() / 2)
+                                : _localSize;
```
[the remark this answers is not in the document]
przemektmalon: It's based on memory limitations of the CPU I'm working on for loading tiles with one dimension being the maximum cache capacity (16 for float). So I'm loading 128 x 16 tiles then using 128 threads to compute the 16 dot products, and repeat until I have all of the them computed.

src/operations/blas2/gemv.hpp
```diff
@@ -80,6 +80,244 @@ SYCL_BLAS_INLINE void AddSetColumns<rhs_t>::adjust_access_displacement() {
...
+
+    // Populate the scratch memory with matrix values, ensuring we don't fetch beyond bounds
+    matrix_scratch[scratch_index] = mat_index < matrix_.get_data_size() ? matrix_.template eval<true>(mat_index) : 0;
+
+    // Move to loading the next tile
+    mat_index += tile_dim_nc * lda;
+    local_nc_index += tile_dim_nc;
+  }
+
+  return;
```
RossBrunton: Bare return at end of function?

src/interface/blas2_interface.hpp
```diff
@@ -55,67 +55,85 @@ typename Executor::policy_t::event_t _gemv_impl(
     index_t _scratchPadSize = 0, index_t _nRowsWG = 0, index_t _nColsWG = 0) {
   typename Executor::policy_t::event_t ret{};
 
-  index_t M = (trn == transpose_type::Normal) ? _M : _N;
-  index_t N = (trn == transpose_type::Normal) ? _N : _M;
+  constexpr int cache_line_size = 64;
+  constexpr int cache_capacity = cache_line_size / sizeof(element_t);
```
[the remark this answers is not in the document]
przemektmalon: I'll rename

src/interface/blas2_interface.hpp
```diff
@@ -55,67 +55,85 @@ typename Executor::policy_t::event_t _gemv_impl(
...
-                                : _localSize;
-  const index_t nRowsWG =
-      (_nRowsWG == 0) ? ((data_layout_t::is_col_major()) ? localSize : 1)
-                      : std::min(M, _nRowsWG);
-  const index_t nColsWG =
-      (_nColsWG == 0) ? ((data_layout_t::is_col_major()) ? localSize : N)
-                      : std::min(N, _nColsWG);
+  auto mA = make_matrix_view<col_major>(ex, _mA, _M, _N, _lda);
+  auto vx = make_vector_view(ex, _vx, _incx, x_vector_size);
+  auto vy = make_vector_view(ex, _vy, _incy, _M);
```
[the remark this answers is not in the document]
przemektmalon: That seems to be right. Never caused a problem in the tests which is strange. I will change it though.

src/interface/blas2_interface.hpp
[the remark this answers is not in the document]
przemektmalon: Yeah pretty much. Basically right now the user cannot choose whether the shared-memory or non-shared-memory kernel runs. I made shared-memory the default, but the calculation for `global_size` is different for the non-shared memory kernel, so I left this in here in case someone was adding this functionality in the future. It would require quite a lot of changes to the interface though I think.

src/interface/blas2_interface.hpp
```diff
@@ -55,67 +55,85 @@ typename Executor::policy_t::event_t _gemv_impl(
...
+                                    WGs_per_C,
+                                    local_size);
+
+  // Execute the GEMV kernel that calculate the partial dot products of rows
+  auto gemvEvent = concatenate_vectors(ret, ex.execute(gemv, local_size,
+                                    global_size,
+                                    kernel_scratch_size));
+
+  // Sum the partial dot products results from the GEMV kernel
+  auto sumColsOp = make_addSetColumns(dot_products_matrix);
```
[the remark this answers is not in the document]
przemektmalon: I thought about using your reduction. But at the moment this is ok for performance. It can be left for a future task however, then we can have proper reduction and not just a simple row sum.

src/operations/blas2/gemv.hpp
```diff
@@ -80,6 +80,244 @@ SYCL_BLAS_INLINE void AddSetColumns<rhs_t>::adjust_access_displacement() {
...
+  const index_t local_id = ndItem.get_local_id(0);
+  const index_t local_range = ndItem.get_local_range(0);
+  const index_t group_id = ndItem.get_group(0);
+  const index_t group_range = ndItem.get_group_range(0);
+  const index_t contract_dim = is_transposed ? matrix_.get_size_row() : matrix_.get_size_col();
+  const index_t non_contract_dim = is_transposed ? matrix_.get_size_col() : matrix_.get_size_row();
+  const index_t group_stride = group_range * local_range;
+  const index_t lda = matrix_.getSizeL();
+
+  const index_t work_load_per_thread = contract_dim;
```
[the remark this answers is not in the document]
przemektmalon: Removed

src/operations/blas2/gemv.hpp
```diff
@@ -80,6 +80,244 @@ SYCL_BLAS_INLINE void AddSetColumns<rhs_t>::adjust_access_displacement() {
...
+
+    index_t non_contract_dim_index = (thread_id + row_id) * non_contract_local_thread_stride;
+
+    sum = 0;
+    for(index_t col_id = 0; col_id < contract_dim; ++col_id) { 
+      sum += ProductOperator::eval(matrix_. template eval<true>(non_contract_dim_index), vector_.eval(col_id));
+      non_contract_dim_index += contract_stride;
+    }
+    
+    lhs_.eval(thread_id + row_id) = sum;
```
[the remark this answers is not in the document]
przemektmalon: Yes. As I said in another comment, the user cannot select which kernel is used, so shared-memory is always used. This function was tested, but less vigorously than the shared memory version.

src/operations/blas2/gemv.hpp
przemektmalon: Changed

src/operations/blas2/gemv.hpp
```diff
@@ -80,6 +80,244 @@ SYCL_BLAS_INLINE void AddSetColumns<rhs_t>::adjust_access_displacement() {
...
+        vector_(_vec), // Input vector
+        WGs_per_NC_(_WGs_per_NC), // number of work groups per non-contracting dimension
+        WGs_per_C_(_WGs_per_C), // number of work groups per contracting dimension
+        local_memory_size_(_local_memory_size) // Local memory shared between a work group
+        {}
+
+template <typename lhs_t, typename matrix_t, typename vector_t, bool is_transposed, int cache_line_size, int work_per_thread>
+SYCL_BLAS_INLINE
+typename Gemv<lhs_t, matrix_t, vector_t, is_transposed, cache_line_size, work_per_thread>::index_t
+Gemv<lhs_t, matrix_t, vector_t, is_transposed, cache_line_size, work_per_thread>::get_size() const { return matrix_.get_size(); }
```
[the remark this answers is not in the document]
przemektmalon: I've removed the function.

src/interface/blas2_interface.hpp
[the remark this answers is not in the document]
przemektmalon: Other lines would have to change to call the non shared memory version though. I think I will just remove it. The interface doesn't allow the user to choose, so shared-memory is all they will get for now.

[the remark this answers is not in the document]
przemektmalon: I've deleted it. The non-shared-mem kernel will remain, but there is no way to call it currently without bigger changes to the interface.

src/operations/blas2/gemv.hpp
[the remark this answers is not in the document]
Nyrio: The question here is more: if it was used, would it be used followed by a reduction or not? And also there was the issue of the `if` statement at the wrong place but this has been solved. The question of the availability through the interface is for another PR anyway.

include/blas_meta.h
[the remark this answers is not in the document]
Nyrio: Yes this code looks very much like it has been copied from a macro because parentheses are useful in that case, because of operator priorities. Here it is not necessary.

Concerning the fact that the function is not used yet, I would comment it out but some people don't like to comment out code so I'm not sure what to do about it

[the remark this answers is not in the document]
mehdi-goli: yes it was copied from macro, since you are using the function, there is no need for prantheses

src/operations/blas2/gemv.hpp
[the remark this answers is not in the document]
przemektmalon: At the moment, each thread computes the dot product for an entire row, there is no reduction necessary. The function can be further optimized, but I mostly focused on the shared-memory version.

src/operations/blas2/gemv.hpp
przemektmalon: Renamed

przemektmalon:
> I have left a few comments. Additionally one general comment: why not making the tile/group dimensions a template parameter, instead of using the id ranges? A lot of expressions could become constexpr and some loops would be easier for the compiler to optimize

Yes that should be done, but in another task I think. The current performance is better than the old implementation, satisfying the task requirements. I couldn't do it from the start as I was still on the learning curve. There should be a task created to add these optimizations, and another to change the interface to allow the user to choose the shared/non-shared kernel.

src/operations/blas2/gemv.hpp
```diff
@@ -80,6 +80,244 @@ SYCL_BLAS_INLINE void AddSetColumns<rhs_t>::adjust_access_displacement() {
...
+    return 0;
+  }
+
+  const index_t non_contract_local_thread_stride = is_transposed ? lda : 1;
+  const index_t contract_stride = is_transposed ?  1 : lda;
+  value_t sum = 0;
+  
+  for (index_t row_id = 0; row_id < non_contract_dim; row_id += group_stride) {
+
+    index_t non_contract_dim_index = (thread_id + row_id) * non_contract_local_thread_stride;
```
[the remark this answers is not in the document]
przemektmalon: Fixed

test/unittest/blas2/blas2_gemv_test.cpp
```diff
@@ -42,40 +42,41 @@ void run_test(const combination_t<scalar_t> combi) {
...
 
   // Input matrix
-  std::vector<scalar_t> a_m(lda * n);
+  std::vector<scalar_t> a_m(a_size, 1.0);
   // Input Vector
-  std::vector<scalar_t> b_v(x * incX);
+  std::vector<scalar_t> x_v(x_size, 1.0);
   // output Vector
-  std::vector<scalar_t> c_v_gpu_result(y * incY, scalar_t(10.0));
+  std::vector<scalar_t> y_v_gpu_result(y_size, scalar_t(0.0));
```
Nyrio: Why do you change that to 0? I think that this is intended on purpose because when using a value different than 0 you'd notice e.g if your algorithm writes a 0 in a location that shouldn't be written (it can happen in that kind of algorithms that read zeroes when out of range, if the check before writing is wrong).

Nyrio: Actually `y_v_gpu_result` could be random and `y_v_cpu` constructed by copy.

src/operations/blas2/gemv.hpp
```diff
@@ -80,6 +80,261 @@ SYCL_BLAS_INLINE void AddSetColumns<rhs_t>::adjust_access_displacement() {
...
+  // Matrix index calculated based on global grid indices
+  index_t mat_index = grid_c_index + grid_nc_index * lda;
+
+  // Number of iterations to load all tiles
+  // local_range / tile_dim_nc == cl_elems
+  constexpr int tile_nc_loops = cl_elems;
+
+  // Transposed index in scratch-space with bank offset which adds 1 word of
+  // padding every 'tile_dim_c' elements
+  const index_t scratch_index = tile_c_index * local_range + tile_c_index;
```
Nyrio: Can you write that `scratch_index = (tile_c_index + 1) * local_range`?

test/unittest/blas2/blas2_gemv_test.cpp
przemektmalon: Reverted

src/operations/blas2/gemv.hpp
przemektmalon: Yup. `tile_c_index` had an alias before for the bank offset, hence the mistake here. (compiler would/should optimize the old code either way).

Nyrio:
Approved. Somebody should follow up to add the following optimizations:

- per-platform selection of the shared-mem or no-shared-mem version
- tile dimensions as a template parameter, also configured per-platform
- double buffering
- make the padding for bank conflicts optional
- compare the performance of no-shared-mem followed by reduction or full operation
- compare the naive reduction against the sophisticated one

src/operations/blas2/gemv.hpp
Nyrio: Actually it's `scratch_index = tile_c_index * (local_range + 1)` sorry

src/operations/blas2/gemv.hpp
```diff
@@ -80,6 +80,261 @@ SYCL_BLAS_INLINE void AddSetColumns<rhs_t>::adjust_access_displacement() {
...
+  // Transposed index in scratch-space with bank offset which adds 1 word of
+  // padding every 'tile_dim_c' elements
+  const index_t scratch_index = (local_range + 1) * tile_c_index;
+
+#pragma unroll
+  for (index_t nc_tile_id = 0; nc_tile_id < tile_nc_loops; ++nc_tile_id) {
+    // Populate the scratch memory with matrix values, ensuring we don't fetch
+    // beyond bounds
+    matrix_scratch[scratch_index + local_nc_index] =
+        mat_index < matrix_a_.get_data_size()
```
Nyrio: The bug is probably here. When lda is greater than m some values that are outside the matrix but with an index smaller than the container size will be read.

Change that to a comparison of the row and column indices to the height and width of the matrix

przemektmalon: Done. Thanks for the fix!

src/operations/blas2/gemv.hpp
```diff
@@ -80,6 +80,266 @@ SYCL_BLAS_INLINE void AddSetColumns<rhs_t>::adjust_access_displacement() {
...
+  // Transposed index in scratch-space with bank offset which adds 1 word of
+  // padding every 'tile_dim_c' elements
+  const index_t scratch_index = (local_range + 1) * tile_c_index;
+
+#pragma unroll
+  for (index_t nc_tile_id = 0; nc_tile_id < tile_nc_loops; ++nc_tile_id) {
+    // Populate the scratch memory with matrix values, ensuring we don't fetch
+    // beyond bounds
+    matrix_scratch[scratch_index + local_nc_index] =
+        grid_nc_index < nc_dim && grid_c_index < c_dim
```
Nyrio: This is still wrong. `grid_c_index` and `grid_nc_index` are constants but the actual nc index is changing with `mat_index += tile_dim_nc * lda`.

przemektmalon: Ah you're right. I will fix it soon.

include/operations/blas2_trees.h
```diff
@@ -26,27 +26,44 @@
...
 struct Gemv {
   using value_t = typename vector_t::value_t;
   using index_t = typename vector_t::index_t;
-  output_t lhs_;
-  matrix_t matrix_;
-  vector_t vector_;
+  lhs_t lhs_;
+  matrix_t matrix_a_;
+  vector_t vector_x_;
+  index_t WGs_per_NC_;
```
mehdi-goli: can you follow the same name convention as other files (AFAIK we don't use capital naming for class members)

mehdi-goli:
Overall looks good! I just left one minor comment.

include/operations/blas2_trees.h
przemektmalon: Updated

mehdi-goli:
> The tests are failing sometimes with `-nan` values, please investigate that

@przemektmalon  is this still the case?

include/operations/blas2_trees.h
```diff
@@ -26,27 +26,44 @@
...
 
-template <typename output_t, typename matrix_t, typename vector_t>
-Gemv<output_t, matrix_t, vector_t> make_gemv(output_t &lhs_, matrix_t &matrix_,
-                                             vector_t &vector_) {
-  return Gemv<output_t, matrix_t, vector_t>(lhs_, matrix_, vector_);
+template <bool is_transposed, int cache_line_size, int work_per_thread,
+          typename lhs_t, typename matrix_t, typename vector_t>
+Gemv<lhs_t, matrix_t, vector_t, is_transposed, cache_line_size, work_per_thread>
+make_Gemv(lhs_t &lhs_, matrix_t &matrix_, vector_t &vector_,
+          typename vector_t::index_t WGs_per_NC_,
```
mehdi-goli: I think we do not use a capital letter for naming variable. Can you fix that @przemektmalon

przemektmalon:
> > The tests are failing sometimes with `-nan` values, please investigate that
> 
> @przemektmalon is this still the case?

No. Has been fixed with the help of @Nyrio

include/operations/blas2_trees.h
przemektmalon: @mehdi-goli No capital letters at all, or just the first letter ? Like is `wGs_per_NC_` ok ?

przemektmalon: @mehdi-goli I've gone with `wgs_per_NC_` and same format for others, looks a bit cleaner I think.

mehdi-goli: thanks, change the NC to nc as well

mehdi-goli: and also if there is _C, change it to c. Thanks!

mehdi-goli:
Thanks for your hard work! @przemektmalon . We are waiting for some internal performance test results before merging this PR!

PR #187, opened 2019-08-14
Change the tests and benchmarks' random distributions to avoid the precision failures

Nyrio:
Most of the failures of our benchmarks with a large k are due to the distribution averaging around 0 so a lot of values in the result are very close to 0 and they eventually break the verification test.

Changing the random distribution greatly reduces these failures while still testing accurately the exactitude of the algorithm.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,7 +22,7 @@
 # *  @filename CMakeLists.txt
 # *
 # **************************************************************************/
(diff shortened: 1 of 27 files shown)


Nyrio:
@RossBrunton This won't impact the tests, only the benchmarks.

We don't hide a bug, it's simply impossible to expect an acceptable relative error when the sum of 400k `float` falls around 0, but this precision error is not relevant for us in the benchmarks.

PR #190, opened 2019-08-15
Replace LICENSE file with full Apache license

jwlawson:
As required by SYCLE-590.


Nyrio:
This full version doesn't include at all the year(s) and name of the company?
Aren't we supposed to include a header with the information `Copyright 2016-2019 Codeplay Software Ltd.` before the full details of the license?

Nyrio:
Thanks for adding the notice. According to the appendix in the license the notice should mention the license, e.g:

```
SYCL-BLAS
Copyright 2016-2019 Codeplay Software Limited.

   Licensed under the Apache License, Version 2.0 (the "License");
   you may not use this file except in compliance with the License.
   A copy of the license is attached (see LICENSE)
```

jwlawson:
I don't think the License header is not required to be included in the NOTICE file, see [the Apache Software Foundation advice](http://www.apache.org/dev/apply-license.html#new):

> Each original source document (code and documentation, but excluding the LICENSE and NOTICE files) SHOULD include a short license header at the top.

and the [example NOTICE file](http://www.apache.org/legal/src-headers.html#notice) the ASF provide.

Nyrio:
Ok. What about the year? All the examples include a year or a range xxxx-xxxx. I think it is needed to determine when the copyright expires.

PR #195, opened 2019-09-24
Removes CMake always inline option, modify always inline define

CodeplayBen:
CMake option SYCL_BLAS_ALWAYS_INLINE is never used so has been removed. SYCL_BLAS_INLINE has been modified to allow the define SYCL_BLAS_ALWAYS_INLINE to be used with lambdas.


jwlawson:
This looks good!  I like an MR that removes things.

I think that the force inlining should only really be done when compiling for the device, but that can be a separate change.

include/blas_meta.h
```diff
@@ -118,10 +118,12 @@ static inline index_t get_power_of_two(index_t wGSize, bool rounUp) {
   return ((!rounUp) ? (wGSize - (wGSize >> 1)) : ++wGSize);
 }
 
-#define SYCL_BLAS_INLINE inline __attribute__((always_inline))
+#define SYCL_BLAS_ALWAYS_INLINE \
```
AdamHarries: I'm a bit intrigued as to why this is broken out into a separate preprocessor variable. Why not just `#define SYCL_BLAS_INLINE __attribute__((flatten)) __attribute__((always_inline)) inline` ?

mehdi-goli: Because we are no using each of them separately in different cases

mehdi-goli:
> This looks good! I like an MR that removes things.
> 
> I think that the force inlining should only really be done when compiling for the device, but that can be a separate change.

Yes @jwlawson, The SYCL_INLINE flag has only been used for functions using inside the kernel, not for the host side. Maybe we can have a #if def and set it to inline when it is not used on the device. So in this case, if a user uses this flag in host functions, it should just be just normal inline

PR #196, opened 2019-10-15
Improved performance when multiplying by beta in GEMM

CodeplayBen:
Changes the way that beta is applied in the GEMM computation to improve performance, vastly reducing or eliminating the performance drop that usually comes with beta values > 0. The beta and exisiting C matrix value are now pre-multiplied and loaded into the result register, the actual beta value divided by the alpha and then the final register result multiplied by alpha to produce the correct result in fewer computations.


src/operations/blas3/gemm_local.hpp
```diff
@@ -326,6 +326,32 @@ class Gemm<input_t, output_t, DoubleBuffer, NbcA, NbcB, ClSize, TileType,
   }
 
  private:
+  /* @brief If beta is not zero then this function will load in values from C,
```
ProGTX: If you want this to be Doxygen you need to start with `/**`

src/operations/blas3/gemm_local.hpp
```diff
@@ -326,6 +326,32 @@ class Gemm<input_t, output_t, DoubleBuffer, NbcA, NbcB, ClSize, TileType,
...
+  beta is zero then this function does nothing. */
+  template <bool check_m_limit, bool check_n_limit, typename InputPointerType>
+  static SYCL_BLAS_INLINE void preload_result(
+      element_t (&reg_res)[item_rows][item_cols], InputPointerType C,
+      const index_t &ldc, const index_t &mc, const index_t &nc,
+      const element_t &beta, bool out_of_range) {
+    if (out_of_range) {
+      return;
+    }
+    if (!is_beta_zero) {
```
ProGTX: You could combine this condition with the previous one, and have an early return, allowing you to reduce indentation

src/operations/blas3/gemm_local.hpp
```diff
@@ -326,6 +326,32 @@ class Gemm<input_t, output_t, DoubleBuffer, NbcA, NbcB, ClSize, TileType,
...
+#pragma unroll
+      for (index_t i = 0; i < item_cols; ++i) {
+#pragma unroll
+        for (index_t j = 0; j < item_rows; ++j) {
+          const bool in_range = do_check<check_m_limit>(j * wg_rows < mc) &&
+                                do_check<check_n_limit>(i < nc);
+          if (in_range) {
+            reg_res[j][i] = beta * C[j * wg_rows];
```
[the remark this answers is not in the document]
mehdi-goli:
```c++
#include<new>
....
template <bool is_beta_zero, bool check_m_limit, bool check_n_limit, typename InputPointerType>
  static SYCL_BLAS_INLINE typename std::enable_if<!is_beta_zero>::type preload_result(
      element_t (&reg_res)[item_rows][item_cols], InputPointerType C,
      const index_t &ldc, const index_t &mc, const index_t &nc,
      const element_t &beta, bool out_of_range) {
    if (out_of_range) {
      return;
    }

#pragma unroll
      for (index_t i = 0; i < item_cols; ++i) {
#pragma unroll
        for (index_t j = 0; j < item_rows; ++j) {
          const bool in_range = do_check<check_m_limit>(j * wg_rows < mc) &&
                                do_check<check_n_limit>(i < nc);
          if (in_range) {
            reg_res[j][i] = beta * C[j * wg_rows];
}
        }
        C = C + ldc;
      }
  }
template <bool is_beta_zero, bool check_m_limit, bool check_n_limit, typename InputPointerType>
  static SYCL_BLAS_INLINE typename std::enable_if<!is_beta_zero>::type preload_result(
      element_t *reg_reg, InputPointerType ,
      const index_t &, const index_t &, const index_t &,
      const element_t &beta, bool out_of_range) {
       // just a simple for loop to set it to zero; then the compiler will capture it and set it to memset as currently you cannot call memset inside the sycl
}

```

src/operations/blas3/gemm_local.hpp
CodeplayBen: Fixed!

src/operations/blas3/gemm_local.hpp
```diff
@@ -357,7 +393,9 @@ class Gemm<input_t, output_t, DoubleBuffer, NbcA, NbcB, ClSize, TileType,
       auto A = orig_A;
       auto B = orig_B;
       auto C = orig_C;
-      element_t reg_res[item_rows][item_cols] = {};
+      element_t reg_res[item_rows][item_cols];
+      preload_result<check_m_limit, check_n_limit>(reg_res[0], C, ldc, mc, nc,
```
mehdi-goli: you can call it `scaling_c` instead of preload_result

CodeplayBen: Changed!

PR #197, opened 2019-10-16
Removing 2D array usage in Gemm Kernels

CodeplayBen:
Removes 2D array usage in Gemm kernels as it offers no performance improvements, can be confusing/cause problems (due to the row major nature of C++ 2D arrays), and does not decay nicely to a single pointer like 1 dimensional arrays.

### DIFF
--- a/src/operations/blas3/gemm_local.hpp
+++ b/src/operations/blas3/gemm_local.hpp
@@ -385,7 +385,7 @@ class Gemm<input_t, output_t, DoubleBuffer, NbcA, NbcB, ClSize, TileType,
       index_t lda, InputPointerType orig_B, index_t ldb, element_t beta,
       OutputPointerType orig_C, index_t ldc, ScratchPointerType s1,
       ScratchPointerType s2, ScratchPointerType s3, ScratchPointerType s4,
-      element_t (&reg_a)[item_rows], element_t &reg_b, const bool out_of_range,
+      element_t *reg_a, element_t &reg_b, const bool out_of_range,
       const index_t batch_stride, const index_t wg_batch_id,
       index_t batch_size) noexcept {
     index_t ofs = 1;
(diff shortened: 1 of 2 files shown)


src/operations/blas3/gemm_local.hpp
```diff
@@ -577,9 +576,11 @@ class Gemm<input_t, output_t, DoubleBuffer, NbcA, NbcB, ClSize, TileType,
    * @param reg_res  2D register array used to store the result C
    */
   template <typename InputPointerType>
-  static SYCL_BLAS_INLINE void compute_block_gemm(
-      InputPointerType B, InputPointerType A, element_t (&reg_a)[item_rows],
-      element_t &reg_b, element_t (&reg_res)[item_rows][item_cols]) noexcept {
+  static SYCL_BLAS_INLINE void compute_block_gemm(InputPointerType B,
+                                                  InputPointerType A,
+                                                  element_t (&reg_a)[item_rows],
```
mehdi-goli: you can just use a pointer here.

src/operations/blas3/gemm_no_local.hpp
```diff
@@ -298,7 +298,7 @@ class Gemm<input_t, output_t, DoubleBuffer, NbcA, NbcB, ClSize, tile_type,
       for (index_t i = 0; i < item_rows; ++i) {
         if (do_check<need_check_boundary>(check_boundary(
                 dim_m_c_start + i * wg_rows, dim_n_c_start + j * wg_cols))) {
-          reg_res[i * item_rows + j] = beta * C[i * wg_rows];
+          reg_res[j * item_rows + i] = beta * C[i * wg_rows];
```
mehdi-goli: can you adjust the index everywhere to be either  i + j*xxx or j+ i* xxx

src/operations/blas3/gemm_local.hpp
CodeplayBen: Fixed!

src/operations/blas3/gemm_no_local.hpp
CodeplayBen: Fixed!

PR #207, opened 2019-11-29
Add items/bytes per second field to GEMM benchmark

jwlawson:
Google benchmark provides a field for items per second. We can use this
to give the number of flops per second achieved by the benchmark, to
allow users to easily compare benchmark results and avoid having to
compute this value themselves.


mehdi-goli:
> Google benchmark provides a field for items per second. We can use this
> to give the number of flops per second achieved by the benchmark, to
> allow users to easily compare benchmark results and avoid having to
> compute this value themselves.

That is a good idea! Also, google bench has a field called byte_per_second which we can use to compute the memory bandwidth. I think that would be useful as well.

PR #208, opened 2019-11-29
Fix compile problems with GEMM auto tuners

jwlawson:
Recent changes to the build system and GEMM kernels caused problems with
the auto tuners. Some small fixes are needed to make them compile.

For now we use a hardcoded vector size of 1, but we should look into
extending the tuners to try different vector sizes to get optimal
performance.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -126,7 +126,7 @@ endif()
 
 option(BLAS_ENABLE_AUTO_TUNERS "Whether to enable building GEMM auto tuners" OFF)
 if(${BUILD_AUTO_TUNERS})
-  # Note that the auto tuners are very slow to compiler, so we avoid adding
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -123,3 +123,10 @@ option(BUILD_EXPRESSION_BENCHMARKS "Whether to build the expression benchmarks"
 if(${BLAS_ENABLE_BENCHMARK})
   add_subdirectory(benchmark)
 endif()
+
+option(BLAS_ENABLE_AUTO_TUNERS "Whether to enable building GEMM auto tuners" OFF)
+if(${BUILD_AUTO_TUNERS})
+  # Note that the auto tuners are very slow to compiler, so we avoid adding
```
DuncanMcBain:
```suggestion
  # Note that the auto tuners are very slow to compile, so we avoid adding
```

PR #211, opened 2019-12-02
Call _scal when alpha is zero in GEMM

jwlawson:
When alpha is zero in GEMM, the operation is just `C = beta * C`, so we
can just call `_scal` to scale the output matrix without computing the
matrix multiply.


test/unittest/blas3/gemm_alpha_zero.cpp
```diff
@@ -0,0 +1,202 @@
+/***************************************************************************
```
mehdi-goli: For the test, it would be good to use the combinatorial test to match the test with what we already have in SYCL-BLAS. So in this case, all the test will be the same. you should be able to pass alpha=0 as one of the parameter in the test.

jwlawson: I could do that, but I am reluctant for a few reasons:
* That would rely on the reference BLAS version handling the alpha=0 case well, and we have seen some implementations do not.
* The alpha=0 case does not depend on many of the parameters to GEMM, so adding it to the combinations will double the number of tests (and so double the test time) with very little gain.
* Doing that loses the ability to easily see whether the alpha=0 case is broken, rather than any of the other cases. The combinations test just shows
    ```
    ...
    [ RUN      ] gemm/GemmFloat.test/60
    Device vendor: Advanced Micro Devices, Inc.
    Device name: Fiji
    Device type: gpu
    [       OK ] gemm/GemmFloat.test/60 (86 ms)
    ...
    ```
    where it is incredibly hard to match up the test case with the set of parameters used.
* The standard combination (i.e. not stress test) actually tests fewer corner cases than the hardcoded test I wrote. In particular it only tests the case where LDC is greater than the size of the matrix C.

If you're happy with these considerations then let me know and I'll switch over to the combinations test.

jwlawson: I'll remove the separate tests, and once !213 is merged will add new test cases using the refactored GEMM tests in that framework.

src/interface/blas3_interface.hpp
```diff
@@ -83,6 +84,22 @@ typename executor_t::policy_t::event_t _gemm_backend(
     index_t _K, element_t _alpha, container_0_t a_, index_t _lda,
     container_1_t b_, index_t _ldb, element_t _beta, container_2_t _C,
     index_t _ldc, index_t batch_size) {
+  if (_alpha == element_t{0}) {
+    // When alpha = 0, GEMM is equivalent to C = beta * C.
+    if (_ldc == _M) {
+      // When LDC is M, we can scale the full matrix at once.
+      return ::blas::_scal(ex, _N * _M, _beta, _C, index_t{1});
```
mehdi-goli: I just realized one thing. When alpha is zero and beta is zero and C is NaN the result according to the netlib gemm should be zero, but in this case, I think the result will be NaN. maybe we need to add another check saying that if beta is zero as well, just sycl.memset(0);

test/unittest/blas3/gemm_alpha_zero.cpp
CodeplayBen: >     * Doing that loses the ability to easily see whether the alpha=0 case is broken, rather than any of the other cases. The combinations test just shows
When a test fails it will show the parameters used both at the point where the test itself fails and when displaying all failed tests at the end of the test run. For example:
```
[ RUN      ] gemm/GemmFloat.test/31
Device vendor: Intel(R) Corporation
Device name: Intel(R) Gen9 HD Graphics NEO
Device type: gpu
Value mismatch at index 0: 3.82841; expected 35.2439
../test/unittest/blas3/blas3_gemm_test.cpp:119: Failure
Value of: utils::compare_vectors(c_m_gpu, c_m_cpu)
  Actual: false
Expected: true
[  FAILED  ] gemm/GemmFloat.test/31, where GetParam() = (5, 512, 49, 21, 't' (116, 0x74), 't' (116, 0x74), 1.5, 1.5, 2, 3, 2) (118 ms)
[----------] 32 tests from gemm/GemmFloat (3264 ms total)

[----------] Global test environment tear-down
[==========] 32 tests from 1 test suite ran. (3264 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 32 tests, listed below:
[  FAILED  ] gemm/GemmFloat.test/0, where GetParam() = (5, 11, 14, 21, 'n' (110, 0x6E), 'n' (110, 0x6E), 1.5, 0, 2, 3, 2)
[  FAILED  ] gemm/GemmFloat.test/1, where GetParam() = (5, 11, 14, 21, 'n' (110, 0x6E), 'n' (110, 0x6E), 1.5, 1.5, 2, 3, 2)
[  FAILED  ] gemm/GemmFloat.test/2, where GetParam() = (5, 11, 14, 21, 'n' (110, 0x6E), 't' (116, 0x74), 1.5, 0, 2, 3, 2)
```
Certainly not the most readable it could be but you can easily see at a glance that tests are failing when alpha = 0 for example.

jwlawson: I've changed to add a new test case to the old test, so I'm marking this resolved.

src/interface/blas3_interface.hpp
jwlawson: I added this change to `_scal`. If you would prefer this behaviour only for `_gemm` then I can move the change there.

mehdi-goli: Thanks! I don't mind where it is added. Inside the scal is good.

PR #213, opened 2019-12-04
Refactor GEMM tests to use single function

jwlawson:
Rather than having a duplicated verification function in each source
file, we can provide a single GEMM verification function in a header,
and then unify how the tests are registered in GTest with a single
macro. This allows different kernel parameters and test cases to be
easily added and amended.

Provide a cached SYCL queue to avoid recompiling kernels in each test
case. This can significantly improve test runtime.

### DIFF
--- a/test/unittest/blas3/blas3_gemm_tall_skinny_test.cpp
+++ b/test/unittest/blas3/blas3_gemm_tall_skinny_test.cpp
@@ -23,83 +23,22 @@
  *
  **************************************************************************/
 
+#include "blas3_gemm_common.hpp"
 #include "blas_test.hpp"
 
-template <typename scalar_t>
-using combination_t =
-    std::tuple<int, int, int, char, char, scalar_t, scalar_t, int, int, int>;
+const auto BatchOneSkinny =
+    ::testing::Combine(::testing::Values(0, 10),     // offset
+                       ::testing::Values(1),         // batch
+                       ::testing::Values(7, 65),     // m
+                       ::testing::Values(9, 126),    // n
+                       ::testing::Values(5678),      // k
+                       ::testing::Values('n', 't'),  // transa
+                       ::testing::Values('n', 't'),  // transb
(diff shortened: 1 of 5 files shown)


test/unittest/blas3/blas3_gemm_tall_skinny_test.cpp
```diff
@@ -23,83 +23,22 @@
...
+const auto BatchOneSkinny =
+    ::testing::Combine(::testing::Values(0, 10),     // offset
+                       ::testing::Values(1),         // batch
+                       ::testing::Values(7, 65),     // m
+                       ::testing::Values(9, 126),    // n
+                       ::testing::Values(5678),      // k
+                       ::testing::Values('n', 't'),  // transa
+                       ::testing::Values('n', 't'),  // transb
+                       ::testing::Values(1.5),       // alpha
+                       ::testing::Values(0.0, 0.5),  // beta
```
mehdi-goli: Here maybe you can separate the beta

test/unittest/blas3/blas3_gemm_tall_skinny_test.cpp
```diff
@@ -23,83 +23,22 @@
...
+                       ::testing::Values(3),         // lda_mul
+                       ::testing::Values(2),         // ldb_mul
+                       ::testing::Values(1, 3)       // ldc_mul
```
mehdi-goli: also maybe this one cab be separated as well

test/unittest/blas3/blas3_gemm_tall_skinny_test.cpp
jwlawson: Sure. I just copied the config for the tall skinny test as it was, but I'll break it up for you.

mehdi-goli: Thanks!

PR #214, opened 2019-12-06
Vectorizing gemm no_local memory kernel

CodeplayBen:
Vectorizes the no_local memory gemm kernel. Currently it is slightly limited in that A is only vectorized when not transposed and B is only vectorized when it is transposed. This still provides good performance on Intel GPU. Default CPU backend vector size has been set to 1 as the likely increased register pressure hurts performance.

### DIFF
--- a/cmake/CmakeFunctionHelper.cmake
+++ b/cmake/CmakeFunctionHelper.cmake
@@ -156,7 +156,7 @@ elseif(${TARGET} STREQUAL "AMD_GPU")  # need investigation
   endif()
 else() # default cpu backend
   set(gemm_configuration_0 "float"  64 "false" "false" "false" 64 8 8 8 8 1 1 "no_local" "naive" 1)
-  set(gemm_configuration_1 "float"  64 "false" "false" "false" 64 8 8 8 8 1 1 "local" "standard" 1)
+  set(gemm_configuration_1 "float"  64 "false" "false" "false" 64 8 8 8 8 1 1 "no_local" "standard" 1)
   set(gemm_configuration_2 "double" 64 "false" "false" "false" 64 8 8 8 8 1 1 "no_local" "naive" 1)
   set(gemm_configuration_3 "double" 64 "false" "false" "false" 64 8 8 8 8 1 1 "no_local" "standard" 1)
 
--- a/cmake/CmakeFunctionHelper.cmake
+++ b/cmake/CmakeFunctionHelper.cmake
@@ -155,10 +155,10 @@ elseif(${TARGET} STREQUAL "AMD_GPU")  # need investigation
     endif()
   endif()
 else() # default cpu backend
-  set(gemm_configuration_0 "float"  64 "false" "false" "false" 64 8 8 8 8 1 1 "no_local" "naive" 8)
-  set(gemm_configuration_1 "float"  64 "false" "false" "false" 64 8 8 8 8 1 1 "no_local" "standard" 8)
-  set(gemm_configuration_2 "double" 64 "false" "false" "false" 64 8 8 8 8 1 1 "no_local" "naive" 8)
-  set(gemm_configuration_3 "double" 64 "false" "false" "false" 64 8 8 8 8 1 1 "no_local" "standard" 8)
+  set(gemm_configuration_0 "float"  64 "false" "false" "false" 64 8 8 8 8 1 1 "no_local" "naive" 1)
+  set(gemm_configuration_1 "float"  64 "false" "false" "false" 64 8 8 8 8 1 1 "local" "standard" 1)
+  set(gemm_configuration_2 "double" 64 "false" "false" "false" 64 8 8 8 8 1 1 "no_local" "naive" 1)
+  set(gemm_configuration_3 "double" 64 "false" "false" "false" 64 8 8 8 8 1 1 "no_local" "standard" 1)
 
   if(NAIVE_GEMM)
     list(APPEND gemm_configuration_lists gemm_configuration_0)
--- a/src/operations/blas3/gemm_load_store.hpp
+++ b/src/operations/blas3/gemm_load_store.hpp
@@ -39,7 +39,9 @@ used no matter what value is passed here.
 supported).
 * @tparam index_t The type of the matrix indices.
 */
-template <bool aligned, size_t vector_size, typename value_t, typename index_t>
+template <bool aligned, size_t vector_size, typename value_t, typename index_t,
+          cl::sycl::access::address_space src_space,
+          cl::sycl::access::address_space dst_space>
 struct Packetize {
 #ifdef GEMM_VECTORIZATION_SUPPORT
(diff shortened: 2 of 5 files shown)


src/operations/blas3/gemm_no_local.hpp
```diff
@@ -416,19 +444,28 @@ class Gemm<input_t, output_t, DoubleBuffer, NbcA, NbcB, ClSize, tile_type,
...
     if (out_of_range) {
       return;
     }
 #pragma unroll
-    for (int i = 0; i < item_size; i++) {
-      reg[i] =
-          do_check<check_block>(chk_boundary(index)) ? ptr[0] : element_t(0);
+    for (int i = 0; i < item_size / work_per_load; i++) {
+      bool in_range =
+          do_check<check_block>(chk_boundary(index + (work_per_load - 1)));
```
jwlawson: Would it be good to add a comment or something explaining what you are dong here?

src/operations/blas3/gemm_no_local.hpp
```diff
@@ -476,21 +533,27 @@ class Gemm<input_t, output_t, DoubleBuffer, NbcA, NbcB, ClSize, tile_type,
...
         }
       }
-      C = C + (wg_cols * ldc);
+      C += ((i + 1) % b_packet_size == 0
+                ? ((wg_cols * b_packet_size - (b_packet_size - 1)) * ldc)
+                : ldc);
     }
   }
-};  // end class No Local GemmFactory
+};  // namespace blas
```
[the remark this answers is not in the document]
CodeplayBen: Thanks, fixed!

src/operations/blas3/gemm_no_local.hpp
```diff
@@ -476,21 +533,27 @@ class Gemm<input_t, output_t, DoubleBuffer, NbcA, NbcB, ClSize, tile_type,
       return;
     }
 #pragma unroll
-    for (int i = 0; i < item_cols; i++) {
+    for (int i = 0; i < item_cols; ++i) {
```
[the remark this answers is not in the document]
CodeplayBen: Fixed to be consistent with the rest of the file!

src/operations/blas3/gemm_no_local.hpp
CodeplayBen: Good idea, added some comments to this function!

src/operations/blas3/gemm_load_store.hpp
```diff
@@ -128,24 +127,20 @@ struct Packetize {
...
-           DestPointerType dest, const index_t &dest_offset,
+      load(const bool in_range, SrcPointerType src, DestPointerType dest,
            EdgePredicate edge_in_range) {
     if (in_range) {
-      *reinterpret_cast<packet_t *>(
-          static_cast<value_t *>(dest + dest_offset)) =
-          *reinterpret_cast<packet_t *>(
-              static_cast<value_t *>(src + src_offset));
+      *reinterpret_cast<PacketType *>(static_cast<value_t *>(dest)) =
+          *reinterpret_cast<PacketType *>(static_cast<value_t *>(src));
```
ProGTX: Why are there both the reinterpret and static cast, wouldn't a single `reinterpret_cast` suffice?

cmake/CmakeFunctionHelper.cmake
```diff
@@ -155,10 +155,10 @@ elseif(${TARGET} STREQUAL "AMD_GPU")  # need investigation
...
   endif()
 else() # default cpu backend
-  set(gemm_configuration_0 "float"  64 "false" "false" "false" 64 8 8 8 8 1 1 "no_local" "naive" 8)
-  set(gemm_configuration_1 "float"  64 "false" "false" "false" 64 8 8 8 8 1 1 "no_local" "standard" 8)
-  set(gemm_configuration_2 "double" 64 "false" "false" "false" 64 8 8 8 8 1 1 "no_local" "naive" 8)
-  set(gemm_configuration_3 "double" 64 "false" "false" "false" 64 8 8 8 8 1 1 "no_local" "standard" 8)
+  set(gemm_configuration_0 "float"  64 "false" "false" "false" 64 8 8 8 8 1 1 "no_local" "naive" 1)
+  set(gemm_configuration_1 "float"  64 "false" "false" "false" 64 8 8 8 8 1 1 "no_local" "standard" 1)
+  set(gemm_configuration_2 "double" 64 "false" "false" "false" 64 8 8 8 8 1 1 "no_local" "naive" 1)
+  set(gemm_configuration_3 "double" 64 "false" "false" "false" 64 8 8 8 8 1 1 "no_local" "standard" 1)
```
ProGTX: I don't quite understand how reducing the vector size improves vectorization

src/operations/blas3/gemm_load_store.hpp
ProGTX: I guess that's from older code, but I see it in a few other places as well

src/operations/blas3/gemm_no_local.hpp
```diff
@@ -188,20 +196,36 @@ class Gemm<input_t, output_t, DoubleBuffer, NbcA, NbcB, ClSize, tile_type,
...
     const index_t wg_col = tile_id_col * block_cols;
+    /*!
+     * @brief is_internal_block_m and is_internal_block_n is used to distinguish
+     * the internal block. Therefore, work items using these blocks dont need to
+     * check for boundaries.
+     */
+    const bool is_internal_block =
+        (m - wg_row >= block_rows) && (n - wg_col >= block_cols);
+
+    /* @brief Packet sizes for A and B. A can only be vectorised if it is not
```
ProGTX: For Doxygen you need to start the comment with `/**` or `/*!`. But even then this would only apply to `a_packet_size`. Might be nicer to do `///` instead for variables.

Also `vectorised`.

src/operations/blas3/gemm_load_store.hpp
CodeplayBen: In this context dest and src are multi_ptrs to value_t, which can't be directly cast to PacketType * without first being cast to a regular value_t *. These functions are removed by #215 anyway!

cmake/CmakeFunctionHelper.cmake
CodeplayBen: The previous values of 8 for vector size were chosen somewhat arbitrarily before the no_local kernel was vectorized. However, upon testing the vectorization actually significantly reduces performance on Intel CPU (which uses these default CPU configurations) likely due to the increased register pressure so they have been set to 1 here (effectively disabling vectorization for default CPU).

src/operations/blas3/gemm_no_local.hpp
CodeplayBen: Fixed, thanks!

CodeplayBen:
Closing this to perform further work on the no_local kernel so it can be vectorised in all cases.

PR #219, opened 2019-12-16
 Generate tuner kernels in separate TUs 

jwlawson:


tools/auto_tuner/CMakeLists.txt
```diff
@@ -40,12 +76,15 @@ foreach(blas_tuner ${SYCL_AUTO_TUNNER_SRCS})
   get_filename_component(tuner_exec ${blas_tuner} NAME_WE)
   set(TARGET tuner_exec ${blas_tuner})
   add_executable(${tuner_exec} ${blas_tuner})
-  target_link_libraries(${tuner_exec} PRIVATE m blas::blas)
+  target_link_libraries(${tuner_exec} PRIVATE m blas::blas tuner_kernel_lib)
```
mehdi-goli: I know that you did not add the `m` here and it was there. Do you know what is it refer to and can we rename it to something meaningful?

jwlawson: That refers to GCC's libm that contains the implementation of various math library functions. We can't easily refer to it with a different name, but it is likely that we don't need to explicitly link it as it is already linked to `libstdc++`.

PR #221, opened 2020-01-02
Reorder cmake commands to prevent overwriting options

jwlawson:
The SYCL-BLAS cmake scripts are very sensitive to the order in which
options are declared and used. Recently this causes link failures when
cmake is run with default options, as the `GEMM_TALL_SKINNY_SUPPORT`
option is used before it has been declared to have a default of `ON`.
This means that the first time the library object targets are generated
they do not include any tall skinny GEMM kernels, however when the host
launcher object is compiled the option has been defaulted on, so the
host expects these kernels to be available.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,13 +42,6 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 set(CMAKE_POSITION_INDEPENDENT_CODE ${BUILD_SHARED_LIBS})
 
-set(CBLAS_INCLUDE ${CMAKE_CURRENT_SOURCE_DIR}/external/cblas)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -131,7 +142,7 @@ if(${BLAS_ENABLE_BENCHMARK})
 endif()
 
 option(BLAS_ENABLE_AUTO_TUNERS "Whether to enable building GEMM auto tuners" OFF)
-if(${BUILD_AUTO_TUNERS})
+if(${BUILD_ENABLE_AUTO_TUNERS})
```
CodeplayBen: Should this not be $BLAS_ENABLE_AUTO_TUNERS instead of $BUILD_ENABLE_AUTO_TUNERS? I can't see any option or variable with that name unless it is coming from somewhere else.

jwlawson: Yes, well spotted.

PR #231, opened 2020-01-17
Add fully vectorized no local memory GEMM kernel

CodeplayBen:
Adds the fully vectorized gemm kernel as a seperate kernel. This kernel is a work in progress - it currently only provides performance increases on Intel CPU at small sizes, so it is only enabled for the "default_cpu" configuration. This PR contains fixes addressing merge comments from #222


src/operations/blas3/gemm_no_local_full_vec.hpp
```diff
@@ -0,0 +1,911 @@
...
+  static constexpr index_t item_cols = tile_type::item_cols;
+  /*! @brief The number of work items in each row of work group */
+  static constexpr index_t wg_rows = tile_type::wg_rows;
+  /*! @brief The number of work items in each column of work group */
+  static constexpr index_t wg_cols = tile_type::wg_cols;
+  /*! @brief Number of rows within a work-group level tile */
+  static constexpr index_t block_rows = wg_rows * item_rows;
+  /*! @brief Number of columns within a work-group level tile */
+  static constexpr index_t block_cols = wg_cols * item_cols;
+  /*! @brief A boolean parameter represents wheather or not matrix A is
```
jwlawson:
```suggestion
  /*! @brief A boolean parameter represents whether or not matrix A is
```

src/operations/blas3/gemm_no_local_full_vec.hpp
```diff
@@ -0,0 +1,911 @@
...
+   * transposed */
+  static constexpr bool trans_a = TransA;
+  /*! @brief A boolean parameter represents wheather or not matrix B is
```
jwlawson:
```suggestion
  /*! @brief A boolean parameter represents whether or not matrix B is
```

src/operations/blas3/gemm_no_local_full_vec.hpp
```diff
@@ -0,0 +1,911 @@
...
+        b_(B),
+        c_(C),
+        alpha_(alpha),
+        beta_(beta / alpha_),
+        batch_size_(batch_size) {}
+
+  /*!
+   * @brief Get the type of this NoLocalGemmFactory as a human readable string.
+   */
+  static SYCL_BLAS_INLINE std::string get_type_string() noexcept {
```
jwlawson: I don't really know where this is used, but it doesn't look like it contains any information about the vector type/size. Is that important?

src/operations/blas3/gemm_no_local_full_vec.hpp
```diff
@@ -0,0 +1,911 @@
...
+    std::ostringstream str{};
+    str << "NoLocalGemmFactory<" << ClSize << ", "
+        << tile_type::get_type_string() << ", "
+        << type_string<value_t>::get_value() << ">";
+    return str.str();
+  }
+  /*!
+   *@brief gt_workgroup_cluster. This function is used to find the optimum
```
jwlawson:
```suggestion
   *@brief get_workgroup_cluster. This function is used to find the optimum
```

src/operations/blas3/gemm_no_local_full_vec.hpp
```diff
@@ -0,0 +1,911 @@
...
+        (item_id / wg_rows) * (trans_b ? vector_ofs : 1);
+
+    /* Exiting from any threads outside of the m and n boundary */
+    const bool out_of_range = ((local_item_id_row + wg_row >= m) ||
+                               (local_item_id_col + wg_col >= n));
+
+    m -= local_item_id_row + wg_row;
+    n -= local_item_id_col + wg_col;
+    /*
+     * The ma and nb are used to adjust the start position of each work-item for
```
jwlawson: This comment doesn't look to correspond to the code. Should `ma` actually by `dim_m_a_start`?

src/operations/blas3/gemm_no_local_full_vec.hpp
```diff
@@ -0,0 +1,911 @@
...
+#ifdef ARM_GPU
+          ,
+          id
+#endif
+      );
+    }
+  }
+  /** @brief If beta is not zero then this function will load in values from C,
+  multiply them by the beta value and store them in the results register. If
+  beta is zero then this function does nothing. */
```
jwlawson: > this function does nothing

Strictly speaking this function sets the values in the results register to zero.

src/operations/blas3/gemm_no_local_full_vec.hpp
```diff
@@ -0,0 +1,911 @@
...
+        out_of_range);
+#ifdef ARM_GPU
+    id.barrier(cl::sycl::access::fence_space::local_space);
+#endif
+
+#pragma unroll
+    for (int j = 0; j < packet_size; j++) {
+#pragma unroll
+      for (int i = 0; i < item_cols / packet_size; i++) {
+        const index_t ofs = i * B_ptr_index;
```
jwlawson: It's possibly not worth spending lots of time on this but some SYCL compilers do not do loop strength reduction, so manually converting these index calculations to repeated additions may give better performance. This transform could probably be applied elsewhere as well.

```
index_t ofs = 0;
index_t col_ofs = 0;
for (int i = 0; i < item_cols / packet_size; i++) {
  ...
  ofs += B_ptr_index;
  col_ofs += (trans_b ? packet_size : 1) * wg_cols;
}
```

src/operations/blas3/gemm_no_local_full_vec.hpp
CodeplayBen: It looks like it is only used in the auto tuner which from what I gather appends the configuration information anyway.

src/operations/blas3/gemm_no_local_full_vec.hpp
CodeplayBen: Good catch, fixed!

src/operations/blas3/gemm_no_local_full_vec.hpp
CodeplayBen: Fixed!

src/operations/blas3/gemm_no_local_full_vec.hpp
CodeplayBen: I've applied this in the instance you specified there but have left the others alone for now. I'll make a note to track this for the future though

src/operations/blas3/gemm_no_local_full_vec.hpp
CodeplayBen: I've updated this to fix some inconsistencies and add the vectorize type and size. Other GEMM kernels also need this but I don't want to clutter this PR with those changes, and it's not a huge priority I think.

jwlawson: Yeah if it's only used in the tuner then it's not particularly important

PR #239, opened 2020-05-08
WIP: Add MiniBatchSize parameter

Rbiessy:
The parameter is only used in the ARM_GPU configuration. It allows each
work item to process a batch of size MiniBatchSize at a time. This
changes the access pattern of the matrices which may work well with the
caching system.

A MiniBatchSize of 1 should compile to the same kernel than before.

### DIFF
--- a/src/operations/blas3/gemm_no_local_partial_vec.hpp
+++ b/src/operations/blas3/gemm_no_local_partial_vec.hpp
@@ -51,15 +51,19 @@ namespace blas {
  * @tparam TransA  if true, matrix A will be transposed on the fly
  * @tparam TransB  if true, matrix B will be transposed on the fly
  * @tparam element_t  type of matrix elements
+ * @tparam MiniBatchSize Parameter tuning how many batches are processed per
+ * block
  */
 template <typename input_t, typename output_t, bool DoubleBuffer, bool NbcA,
           bool NbcB, int ClSize, typename tile_type, bool TransA, bool TransB,
-          typename element_t, bool is_beta_zero, int VectorSize>
+          typename element_t, bool is_beta_zero, int VectorSize,
+          int MiniBatchSize>
 class Gemm<input_t, output_t, DoubleBuffer, NbcA, NbcB, ClSize, tile_type,
            TransA, TransB, element_t, is_beta_zero,
            static_cast<int>(gemm_memory_t::no_local),
            static_cast<int>(gemm_algorithm_t::standard),
-           static_cast<int>(gemm_vectorization_t::partial), VectorSize> {
+           static_cast<int>(gemm_vectorization_t::partial), VectorSize,
+           MiniBatchSize> {
  public:
   using value_t = element_t;
   using index_t = typename std::make_signed<typename input_t::index_t>::type;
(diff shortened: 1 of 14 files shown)


Rbiessy:
The tests are passing with -DARM_GPU. I still need to run some benchmarks.

src/operations/blas3/gemm_no_local_partial_vec.hpp
```diff
@@ -356,48 +362,58 @@ class Gemm<input_t, output_t, DoubleBuffer, NbcA, NbcB, ClSize, tile_type,
       auto B = orig_B;
       auto C = orig_C;
 
-      /* 2D register array used to store the result C*/
-      value_t reg_res[item_rows * item_cols];
-      scaling_c<need_check_boundary, a_packet_size, b_packet_size>(
-          reg_res, C, ldc, dim_m_a_start, dim_n_b_start, boundary_check_c,
-          out_of_range);
-      while (k > 0) {
```
ProGTX, line 364: What's the reason this was changed from a `while` to a `do-while`? How come it has to happen at least once now?

src/operations/blas3/gemm_no_local_partial_vec.hpp
```diff
@@ -511,8 +527,8 @@ class Gemm<input_t, output_t, DoubleBuffer, NbcA, NbcB, ClSize, tile_type,
   }
 
   /*!
-   * @brief For each work itemThe following function store the computed block of
-   * GEMM reg_res into output matrix C
+   * @brief For each work item the following function store the computed block
```
ProGTX, line 530:
```suggestion
   * @brief For each work item the following function stores the computed block
```

test/unittest/blas3/blas3_gemm_batched_test.cpp
```diff
@@ -44,7 +44,7 @@ GENERATE_GEMM_TEST(BatchGemm, BetaNonZeroLDMatch);
 
 const auto BetaNonZeroLDMultiplied =
     ::testing::Combine(::testing::Values(0),         // offset
-                       ::testing::Values(5),         // batch
+                       ::testing::Values(8),         // batch
```
ProGTX, line 47: Does it not work with `5` anymore? Is there a restriction on being a multiple of 4?

src/operations/blas3/gemm_no_local_partial_vec.hpp
Rbiessy: This is a minor optimization because k is guaranteed to be at least 1. It probably doesn't make any difference.

test/unittest/blas3/blas3_gemm_batched_test.cpp
Rbiessy: This is put to 8 to test the MiniBatchSize parameter. The kernel with MiniBatchSize=4 is used only if batch_size is divisible by 4. It still works with 5, there just wasn't any reason to keep both 5 and 8 I think.

 I'm still not sure whether we should add a specific test for MiniBatchSize, otherwise I can just add a comment explaining why we need a parameter divisible by 4.

Rbiessy:
> should the new parameter be documented somewhere?

There is a tiny bit of documentation here: https://github.com/codeplaysoftware/sycl-blas/pull/239/files#diff-6de77a53b2ed53d8bd6381f1f5f16fd0R54

I don't know what else to say about this parameter.

Rbiessy:
This does not improve performance so closing the MR. Note that the changes here break the auto_tuner.

PR #240, opened 2020-05-08
Batch GEMM with alpha==0 and batch>1

ProGTX:
When `alpha == 0`, GEMM can just scale the matrix.
However, this currently doesn't take the `batch_size` into account,
leading to wrong results.

This PR fixes this issue and adds tests for the new cases.
It also updates existing tests by changing `alpha` and `beta`
to something other than `1` -
the issue this PR is fixing doesn't present itself when `beta == 1`.

### DIFF
--- a/test/unittest/blas3/blas3_gemm_batched_test.cpp
+++ b/test/unittest/blas3/blas3_gemm_batched_test.cpp
@@ -34,8 +34,8 @@ const auto BetaNonZeroLDMatch =
                        ::testing::Values(63, 128),   // k
                        ::testing::Values('n', 't'),  // transa
                        ::testing::Values('n', 't'),  // transb
-                       ::testing::Values(1.0),       // alpha
-                       ::testing::Values(1.0),       // beta
+                       ::testing::Values(3.0),       // alpha
+                       ::testing::Values(7.0),       // beta
                        ::testing::Values(1),         // lda_mul
                        ::testing::Values(1),         // ldb_mul
                        ::testing::Values(1)          // ldc_mul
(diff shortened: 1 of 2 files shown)


test/unittest/blas3/blas3_gemm_batched_test.cpp
```diff
@@ -34,8 +34,8 @@ const auto BetaNonZeroLDMatch =
                        ::testing::Values(63, 128),   // k
                        ::testing::Values('n', 't'),  // transa
                        ::testing::Values('n', 't'),  // transb
-                       ::testing::Values(1.0),       // alpha
-                       ::testing::Values(1.0),       // beta
+                       ::testing::Values(3.0),       // alpha
+                       ::testing::Values(7.0),       // beta
```
Rbiessy, line 38: Why this change? There is no special case for alpha=1 or for beta=1 as far as I know.

test/unittest/blas3/blas3_gemm_batched_test.cpp
```diff
@@ -50,11 +50,42 @@ const auto BetaNonZeroLDMultiplied =
...
```
mehdi-goli, line 54: same here

test/unittest/blas3/blas3_gemm_batched_test.cpp
ProGTX: The issue this PR is fixing doesn't get triggered when `beta == 1`. I discovered that because I copy-pasted the original tests :) I thought it might be better in the future to have more complicated values to potentially capture more issues. I can remove it if you don't think that should be included in this PR.

Rbiessy: Right, I don't mind if it stays.

PR #245, opened 2020-07-13
Allow generic types

ProGTX:
Right now only `DOUBLE_SUPPORT` is possible,
and there is a lot of manual intervention in benchmarks and tests.

This patch generalizes support for types
by using `BLAS_DATA_TYPES` variable.
For example, instead of specifying `-DDOUBLE_SUPPORT=ON` to CMake,
the user would specify `-DBLAS_DATA_TYPES="float;double"`.
The old flag is still there for backwards compatibility.

Note however that the reference BLAS
doesn't support generic types yet.
That is a more difficult problem
because system BLAS requires either `float*` or `double*` to data.

1. `BLAS_DATA_TYPES` CMake variable
    * Passed in as a CMake list
    * `float` is a mandatory type
2. Introduced mapping of CMake string types to C++ types
    * Allows the user to pass in a simpler representation
    * Done by `to_cpp_type`
    * No special types supported at the moment,
      but it allows e.g. cl::sycl::half
3. Replaced `DOUBLE_SUPPORT` C++ macro
  with `BLAS_DATA_TYPE_DOUBLE`
4. Generic way to register benchmarks
    * `BLAS_REGISTER_BENCHMARK`
5. Generic way to register tests
    * `BLAS_REGISTER_TEST`
    * Requires `combination_t` to be templated,
      so some types had to be made into templates
6. `TypeDispatcher` replaced with `blas_system_function`
    * More generic, separates function retrieval from the call
    * Not enough to actually allow different types,
      this is only the foundation
7. Conditionally instantiate policy handler
8. Minor fixes
    * System `dot` can take `y` as a constant
    * Renamed `a` and `b` in `gemv` and `ger` to `x` and `y`
    * Some clang-format


benchmark/common_utils.hpp
```diff
@@ -566,4 +566,34 @@ static inline void calc_avg_counters(benchmark::State& state) {
 }  // namespace utils
 }  // namespace blas_benchmark
 
+/** Registers benchmark for the float data type
+ * @see BLAS_REGISTER_BENCHMARK
+ */
+#define BLAS_REGISTER_BENCHMARK_FLOAT(register_benchmark, args, exPtr, \
```
DuncanMcBain, line 572: I can't see anywhere that this is called where `register_benchmark` isn't just `register_benchmark`. It's quite weird looking!

cmake/CmakeFunctionHelper.cmake
```diff
@@ -26,18 +26,27 @@
...
 
+# Maps a user provided data type name to the C++ type
+# See BLAS_DATA_TYPES
+function(to_cpp_type output data)
+  set(${output} "${data}" PARENT_SCOPE)
+endfunction()
+
+# Strips some C++ symbols from the string to make it suitable to be used
+# as part of a filename
+function(cpp_type_to_name output cpp_type)
```
DuncanMcBain, line 45: You might be interested in https://cmake.org/cmake/help/latest/command/string.html#make-c-identifier

cmake/CmakeFunctionHelper.cmake
```diff
@@ -26,18 +26,27 @@
...
-if(DOUBLE_SUPPORT)
-  set(data_list "float" "double")
-endif()
 
 ## represent the list of bolean options
 set(boolean_list "true" "false")
 
+# Maps a user provided data type name to the C++ type
+# See BLAS_DATA_TYPES
+function(to_cpp_type output data)
```
DuncanMcBain, line 39: I don't really understand why this is a function, or what it does. Can't you just inline the `set` call where you use it? Then you don't need the `PARENT_SCOPE`

cmake/CmakeFunctionHelper.cmake
ProGTX: Not quite what I wanted because it leads to longer names, but I agree it's better to use a standard function: https://github.com/codeplaysoftware/sycl-blas/pull/245/commits/da4befce878c6de889783bfb5e38056d8977289e

benchmark/common_utils.hpp
ProGTX: TBH I was expecting feedback to go into the other direction, to do the same for the test macro :) Removed the parameter: https://github.com/codeplaysoftware/sycl-blas/pull/245/commits/ae90906486cd0ad960fcc538bc1edc6ad87e2552

cmake/CmakeFunctionHelper.cmake
ProGTX: In this PR the function doesn't really do anything, but it's important for new data types. For exaple, the user might pass in `_DBLAS_DATA_TYPES="float;half"`, but the actual C++ type would be `cl::sycl::half`.

benchmark/common_utils.hpp
```diff
@@ -569,31 +569,29 @@ static inline void calc_avg_counters(benchmark::State& state) {
...
   register_benchmark<double>(args, exPtr, success)
 #else
-#define BLAS_REGISTER_BENCHMARK_DOUBLE(register_benchmark, args, exPtr, success)
+#define BLAS_REGISTER_BENCHMARK_DOUBLE(args, exPtr, success)
 #endif  // BLAS_DATA_TYPE_DOUBLE
 
-/** Registers benchmark for all supported data types
- * @param register_benchmark Name of the function to call to register benchmark
+/** Registers benchmark for all supported data types.
+ *  Expects register_benchmark<scalar_t> to exists.
```
DuncanMcBain, line 586:
```suggestion
 *  Expects register_benchmark<scalar_t> to exist.
```
or
```suggestion
 *  Expects register_benchmark<scalar_t> exists.
```
or
```suggestion
```
(I don't really think it's a necessary comment)

cmake/CmakeFunctionHelper.cmake
DuncanMcBain: Oh, I see. Maybe it would be easier to have the old style flags (enable double, enable half etc.) and use the presence of those flags to construct the list internally, such that it happens once and you know its contents are correct?

ProGTX: I don't know, I prefer to have a list of supported types instead of potentially many different flags. If you misspell the flag CMake will just warn of unused variables, whereas with wrong types in a list you'll get configuration or compilation errors. I think it's also easier to search for `BLAS_DATA_TYPE` (without the S) in the repo which will give you all the places where any type related shenaningans are happening, which really shouldn't be that many places.

DuncanMcBain: I don't really agree, but I don't have strong feelings. We could still move the processing of the list such that each element in it is guaranteed to be the correct SYCL type by this point in the code, then you don't need to jump out to this function each time. In theory, can I have `DATA_TYPES=half` without `float` and it will work?

ProGTX: In theory, skipping `float` could work, but I didn't invest the necessary effort into it because I'm focusing on `cl::sycl::half`. I think it would probably be relatively easy to do, but there might be some hidden issues.

I removed the function for now and instead I just sanitize the filename: https://github.com/codeplaysoftware/sycl-blas/pull/245/commits/f83012489997567492d2508b6a3f0af99597604a . Note that this functionality has been there before, I've just made it into a function and extended the filtering to also cover `:`. `MAKE_C_IDENTIFIER` isn't suitable anymore because it also converts the `.` of the file extension.

PR #248, opened 2020-07-27
WIP: System BLAS conversion

ProGTX:
As mentioned in https://github.com/codeplaysoftware/sycl-blas/pull/245, system BLAS (used for reference) is the only thing preventing generic types from working. The main problem is that system BLAS functions can only take `float*` or `double*`, not any other array type, so there needs to be a conversion at some point.

Even though this patch works with generic data types, it's not necessarily meant to be merged. Instead it's an RFC to help me figure out what could be improved and if there are better alternatives.

This patch introduces helper classes for converting an array or a scalar of `T` to an array or scalar of `float` or `double`. I believe this part is essential to any form generic type support. There is some complexity in the classes that allows for the operation to be no-op if `T` is already `float` or `double`, which isn't absolutely necessary, just a slight optimization.

The question is then when does the conversion take place. In this patch it happens when `reference_blas::func` is called, which involves some complicated calculations for the data sizes.

An alternative to this is to perform the conversion at the point of calling reference BLAS. This is done at 4 places for each function (I marked `asum` as an example with the comment `TODO(Peter): reference_blas called here`). This would reduce the complexity of the reference BLAS, but is a much larger code change.

### DIFF
--- a/include/utils/system_reference_blas.hpp
+++ b/include/utils/system_reference_blas.hpp
@@ -29,6 +29,11 @@
 #include "cblas.h"
 #include <cmath>
 #include <iostream>
+#include <vector>
+
+#ifdef BLAS_DATA_TYPE_HALF
+#include <CL/sycl.hpp>
+#endif  // BLAS_DATA_TYPE_HALF
 
 namespace {
 CBLAS_TRANSPOSE c_trans(char x) {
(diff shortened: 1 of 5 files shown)


include/utils/system_reference_blas.hpp
```diff
@@ -29,6 +29,11 @@
 #include "cblas.h"
 #include <cmath>
 #include <iostream>
+#include <vector>
+
+#ifdef BLAS_DATA_TYPE_HALF
+#include <CL/sycl.hpp>
+#endif  // BLAS_DATA_TYPE_HALF
```
ProGTX, line 36: Please ignore this

ProGTX:
Closing in favor of a quantization kernel to be used at the call site.

PR #251, opened 2020-08-17
Support for cl::sycl::half

ProGTX:
1. Added specializations for `cl::sycl::half`
    * `AbsoluteValue`, `DetectScalar`, `PolicyHandler`
2. Defined new macros
    * `BLAS_DATA_TYPE_HALF`, `BLAS_REGISTER_TEST_HALF`,
      `BLAS_REGISTER_BENCHMARK_HALF`
3. Some changes required because `cl::sycl::half` is not a regular type
    * Defined `operator<<`
    * Treat half as floating point
    * Specializations for `blas::constant`
    * Removed constexpr from `init``
    * Specialized some builtins
4. Some static casts or explicit construction required
    * Align types in ternary operator
    * Construct constants as `value_t` or `scalar_t`
5. Enabled `cl::sycl::half` GEMM configurations
    * AMD GPU, ARM GPU, Power VR

### DIFF
--- a/cmake/Modules/ConfigureSYCLBLAS.cmake
+++ b/cmake/Modules/ConfigureSYCLBLAS.cmake
@@ -44,7 +44,7 @@ if("double" IN_LIST BLAS_DATA_TYPES)
   add_definitions(-DBLAS_DATA_TYPE_DOUBLE)
 endif()
 
-if("cl::sycl::half" IN_LIST BLAS_DATA_TYPES)
+if("half" IN_LIST BLAS_DATA_TYPES)
   add_definitions(-DBLAS_DATA_TYPE_HALF)
 endif()
 
--- a/cmake/Modules/ConfigureSYCLBLAS.cmake
+++ b/cmake/Modules/ConfigureSYCLBLAS.cmake
@@ -44,6 +44,10 @@ if("double" IN_LIST BLAS_DATA_TYPES)
   add_definitions(-DBLAS_DATA_TYPE_DOUBLE)
 endif()
 
+if("cl::sycl::half" IN_LIST BLAS_DATA_TYPES)
+  add_definitions(-DBLAS_DATA_TYPE_HALF)
+endif()
+
 # If the user has specified a specific workgroup size for tests, pass that on to the compiler
 if(WG_SIZE)
   add_definitions(-DWG_SIZE=${WG_SIZE})
--- a/src/operations/blas2/gemv.hpp
+++ b/src/operations/blas2/gemv.hpp
@@ -201,8 +201,9 @@ SYCL_BLAS_INLINE
 
   // Threads pre-fetch portions of X into local group-shared memory
   const index_t x_vec_index = local_id + c_group_id * local_range;
-  vector_scratch[local_id] =
-      x_vec_index < vector_x_.get_size() ? vector_x_.eval(x_vec_index) : 0;
+  vector_scratch[local_id] = x_vec_index < vector_x_.get_size()
+                                 ? vector_x_.eval(x_vec_index)
+                                 : value_t{0};
 
   // Barrier to ensure whole portion of vector X is in local memory
   ndItem.barrier(cl::sycl::access::fence_space::local_space);
@@ -337,7 +338,7 @@ Gemv<lhs_t, matrix_t, vector_t, local_range, is_transposed, cache_line_size,
     matrix_scratch[scratch_index + local_nc_index] =
         in_c_range && grid_nc_index < nc_dim
             ? matrix_a_.template eval<true>(mat_index)
-            : 0;
+            : value_t{0};
 
     // Move to loading the next tile
     mat_index += tile_dim_nc * lda;
@@ -580,7 +581,8 @@ GemvRow<interLoop, Lower, Diag, Upper, Unit, lhs_t, matrix_t, vector_t>::eval(
 
   index_t frs_col = idWFC * dimWFC + interLoop * localid;
   index_t lst_col = std::min(dimC, frs_col + dimWFC);
-  static constexpr value_t init_val = AddOperator::template init<vector_t>();
+  // TODO(Peter): Does it hurt performance if this is not constexpr?
+  static const value_t init_val = AddOperator::template init<vector_t>();
   // PROBLEM IF ONLY SOME THREADS OF A WORKGROUP ARE CANCELED
   // TO SOLVE IT, USE GLOBAL VALUES OF frs_col AND lst_col
   if ((!Upper && (((idWFC * dimWFC) + ((!Diag) ? 1 : 0)) > (lst_row - 1))) ||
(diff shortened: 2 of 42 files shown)


cmake/Modules/ConfigureSYCLBLAS.cmake
```diff
@@ -44,6 +44,10 @@ if("double" IN_LIST BLAS_DATA_TYPES)
   add_definitions(-DBLAS_DATA_TYPE_DOUBLE)
 endif()
 
+if("cl::sycl::half" IN_LIST BLAS_DATA_TYPES)
```
mcleary, line 47: Are you sure this is a good name? Shouldn't this be just `half`?

src/operations/extension/reduction_partial_rows.hpp
```diff
@@ -44,7 +44,7 @@ class ReductionPartialRows {
   using params_t = ReductionRows_Params<index_t, element_t, ClSize, WgSize>;
 
   /* Neutral value for this reduction operator */
-  static constexpr value_t init_val = operator_t::template init<output_t>();
+  static const value_t init_val;
```
mcleary, line 47: I'm not following why you did this

cmake/Modules/ConfigureSYCLBLAS.cmake
ProGTX: This is what I initially wanted to do: https://github.com/codeplaysoftware/sycl-blas/pull/245#discussion_r454552435. It's imporntant to have both a user facing name and the C++ type it maps to. There are multiple options to do it:
1. Have a conversion function
2. Convert user provided list into an internal one
3. Keep both lists

Maybe there's another option that I'm not seeing. I personally prefer 1. 2 would be pretty easy to do, but this line (and others) would stay the same. I explored 3 a little bit, but the connected between the lists seems very fragile: there are a few places in CMake where it's OK to use the user provided name, but also places where you must use the C++ type.

src/operations/extension/reduction_partial_rows.hpp
ProGTX: This worries me a little bit - whether it has any impact on performance - but the reason is that `cl::sycl::half` is not `constexpr`.

CodeplayBen (approved):
Looks good to me!

src/operations/extension/reduction_partial_rows.hpp
mehdi-goli: This can affect the performance on other types.

ProGTX: I can take another look, but I'm not sure how best to proceed. I guess I should measure and if it doesn't have an impact it can go in as-is?

CodeplayBen: @ProGTX Have you had a chance to check the performance impact of this change?

ProGTX:
Blocked by https://github.com/codeplaysoftware/sycl-blas/pull/262

cmake/Modules/ConfigureSYCLBLAS.cmake
mehdi-goli: having :: and namespace in the name does not seem convenient. Can you make it float16?

src/operations/blas2/gemv.hpp
```diff
@@ -580,7 +581,8 @@ GemvRow<interLoop, Lower, Diag, Upper, Unit, lhs_t, matrix_t, vector_t>::eval(
 
   index_t frs_col = idWFC * dimWFC + interLoop * localid;
   index_t lst_col = std::min(dimC, frs_col + dimWFC);
-  static constexpr value_t init_val = AddOperator::template init<vector_t>();
+  // TODO(Peter): Does it hurt performance if this is not constexpr?
```
mehdi-goli, line 584: Can you test that and if it does revert it, If not you can remove the TODO. It is important that we do not introduce performance regression.

src/operations/extension/reduction_partial_rows.hpp
mehdi-goli: It is very important that the init value to be known at compile time. In this case, the compiler will be able to flatten the value, and later on when it is used in a kernel or for loop, it can unroll the for loop and promote the array element to a register. Can you please revert that change?

mehdi-goli (changes requested):

src/operations/blas2/gemv.hpp
ProGTX: Wrt. name: originally it was `half` for what the user passes in, and at some point it has to get translated to the C++ type. I've added the translation function here: https://github.com/codeplaysoftware/sycl-blas/pull/251/commits/d6b68d572c46a7302efd412782fa82fd414a80dd

Wrt. performance, we still need to test that.

src/operations/extension/reduction_partial_rows.hpp
ProGTX: The problem is with the lack of `constexpr`, so this cannot be reverted for now. @mcleary is getting performance numbers, but so far it doesn't seem like there's a performance regression.

cmake/Modules/ConfigureSYCLBLAS.cmake
ProGTX: Updated with a conversion function, the user would just use `half`: https://github.com/codeplaysoftware/sycl-blas/pull/251/commits/c49578f8c5a9a2de31374da18e08d0d16b6e34b9

src/operations/extension/reduction_partial_rows.hpp
mehdi-goli: Yes, I understand, Since the issue will be fixed in SYCL-2020 (Hopefully), 
Can add a comment explaining the issue followed by TODO/FIXME saying the issue will be fixed  (comment out the Constexpr line heare) once SYCL-2020 is feature is added to computeCpp.

mehdi-goli: Also, the `int_val` variable should have an initial value  `operator_t::template init<output_t>();` and the initial value is not necessarily zero. Where do you set that?

mehdi-goli (approved):
Overall looks good to me! Thanks! Just add [this](https://github.com/codeplaysoftware/sycl-blas/pull/251#discussion_r552561835) so that we won't forget why we did this conversion and we can revert it back once the issue is fixed

src/operations/extension/reduction_partial_rows.hpp
ProGTX: I've added TODOs: https://github.com/codeplaysoftware/sycl-blas/pull/251/commits/3cfa6b09ac2b423e048f8fe8a6ea54ec1b79f381

The reason for this particular line, why it's not initialized at this point, is because of C++ rules on initializing static variables. In this case it can't be initialized inline anymore because it's not a constant initialization anymore, so it had to be moved out of the class.

mehdi-goli: Thanks! As long as it is initialized somewhere before using, it is ok. It will be fixed once we have the `constexpr` for half.

PR #252, opened 2020-08-17
Quantize all tests and benchmarks

ProGTX:
* Quantize data when using reference BLAS
* Quantize data in a few leftover cases


mcleary:
As far as I can see there is no actual quantization happening just yet right? All tests uses double or float.

benchmark/syclblas/blas1/asum.cpp
```diff
@@ -44,24 +44,26 @@ void run(benchmark::State& state, ExecutorType* executorPtr, index_t size,
...
 
+  using data_t = utils::data_storage_t<scalar_t>;
+
   // Create data
-  std::vector<scalar_t> v1 = blas_benchmark::utils::random_data<scalar_t>(size);
-  scalar_t vr;
+  std::vector<data_t> v1 = blas_benchmark::utils::random_data<data_t>(size);
+  data_t vr;
 
-  auto inx = blas::make_sycl_iterator_buffer<scalar_t>(v1, size);
```
mehdi-goli, line 51: A buffer iterator can create a subset of the buffer/data and the data passed as an input is not necessarily going to be a vector. Actually, it is a raw pointer, for most cases. Removing this can break the c-style support of BLAS, which should be traditionally maintained. So the size must be passed as an input.

test/exprtest/blas1_axpy_copy_test.cpp
```diff
@@ -104,8 +99,7 @@ void run_test(const combination_t<scalar_t> combi) {
   ex.get_policy_handler().wait(axpy_event);
 
   // Copy the result back to host memory
-  auto getResultEv =
-      ex.get_policy_handler().copy_to_host(gpu_y_v, v_y.data(), y_dim);
+  auto getResultEv = utils::quantized_copy_to_host<scalar_t>(ex, gpu_y_v, v_y);
```
mehdi-goli, line 102: Since we have quantized copy to host, do we have a quantized copy to a device? Explicit data movement is crucial for getting performance since it separates the computation from the data movement. Just to mention that you don't need to add it here, since it is a test and it does not matter whether or not you explicitly copy the data.

mehdi-goli (commented):
Looks good to me, I have added some minor comments

test/exprtest/blas1_axpy_copy_test.cpp
CodeplayBen: @mehdi-goli I think the quantize to device happens as part of the call to ```utils::make_quantized_buffer```

mehdi-goli: So you mean it quantizes and copies it to the device?

CodeplayBen: Yeah you can see it [here I believe ](https://github.com/ProGTX/sycl-blas/commit/cf48142082bea1ea864fc438ed13114cb9e69d75#diff-64279e7f75aee013cccd104e34670f11R84) where this is called from within ```utils::make_quantized_buffer```.

CodeplayBen (approved):
Looks good to me apart from the already mentioned concerns about ```make_quantized_buffer```!

benchmark/syclblas/blas1/asum.cpp
ProGTX: This is just removing some calls from the tests and benchmarks, and replacing them with utility functions that calls this internally. Additionally there are still a few places where this is used.

test/exprtest/blas1_axpy_copy_test.cpp
ProGTX: Correct, these are the steps `make_quantized_buffer` takes when having to quantize:
 * 1. Constructs a buffer to hold the input data, which is float or double
 * 2. Copies input data to the buffer
 * 3. Constructs a buffer to hold data of scalar_t
 * 4. Performs quantization from input data to scalar_t

For `float` and `double` it can skip the quantization steps.

benchmark/syclblas/blas1/asum.cpp
CodeplayBen: If that's the case and you can still make a subset manually then that's fine I think.

test/exprtest/blas1_axpy_copy_test.cpp
CodeplayBen: Per the other discussion as this is just a helper function it seems fine, if a user wants more explicit control over when exactly the data is quantized/copied to the device they can just do it the regular way without the helper.

mehdi-goli (approved):

PR #253, opened 2020-08-18
Fix Windows build of sycl-blas

mcleary:
- Long filenames are problematic on Windows. This hashes long filenames to avoid build failures on Windows
- SYCL-BLAS is not prepared to generate a DLL because all symbols must be resolved at link time, so force the library to be linked statically on Windows
- Minor C++ fixes to make the code compatible with MSVC
- Update the ComputeCpp-SDK submodule to get some flags required to build sycl code on Windows

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -68,6 +68,15 @@ list(APPEND COMPUTECPP_USER_FLAGS
   -no-serial-memop
 )
 
+if (MSVC)
+  # The device compiler needs C++14 to parse the Windows headers
+  set(CMAKE_CXX_STANDARD 14)
+  set(BUILD_SHARED_LIBS FALSE CACHE BOOL
+    "Force SYCL-BLAS to be built as a static library on Windows"
+    FORCE
+  )
+endif()
+
 # By default, tall and skinny Gemm is enabled (for better performance)
 option(GEMM_TALL_SKINNY_SUPPORT "Whether to enable tall and skinny Gemm" ON)
 # By default vectorization in gemm kernels is disabled until fully implemented.
@@ -104,6 +113,14 @@ include(CmakeFunctionHelper)
 
 add_subdirectory(src)
 build_library(sycl_blas)
+
+if (WIN32)
+  # On Windows, all symbols must be resolved at link time for DLLs.
+  # The sycl_blas target is just a collection of other objects so
+  # the linked libraries are not going to be propagated to this target.
+  # This requires manual linking against ComputeCpp.lib on Windows.
+  target_link_libraries(sycl_blas PRIVATE ComputeCpp::ComputeCpp)
+endif()
 
 set_target_properties(sycl_blas PROPERTIES
   VERSION ${PROJECT_VERSION}
--- a/src/policy/sycl_policy_handler.cpp
+++ b/src/policy/sycl_policy_handler.cpp
@@ -57,7 +57,7 @@ namespace blas {
   template typename codeplay_policy::event_t                                   \
   PolicyHandler<codeplay_policy>::copy_to_device<element_t>(                   \
       const element_t *src, BufferIterator<element_t, codeplay_policy> dst,    \
-      size_t size = 0);                                                        \
+      size_t size);                                                            \
   template typename codeplay_policy::event_t                                   \
   PolicyHandler<codeplay_policy>::copy_to_host<element_t>(                     \
       element_t * src, element_t * dst, size_t size);                          \
@@ -65,7 +65,7 @@ namespace blas {
   template typename codeplay_policy::event_t                                   \
   PolicyHandler<codeplay_policy>::copy_to_host<element_t>(                     \
       BufferIterator<element_t, codeplay_policy> src, element_t * dst,         \
-      size_t size = 0);                                                        \
+      size_t size);                                                            \
   template ptrdiff_t PolicyHandler<codeplay_policy>::get_offset<element_t>(    \
       const element_t *ptr) const;                                             \
                                                                                \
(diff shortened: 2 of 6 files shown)


CMakeLists.txt
```diff
@@ -68,6 +68,13 @@ list(APPEND COMPUTECPP_USER_FLAGS
   -no-serial-memop
 )
 
+if (MSVC)
+  list(APPEND COMPUTECPP_USER_FLAGS
+    -D_ALLOW_COMPILER_AND_STL_VERSION_MISMATCH=1
```
ProGTX, line 73: I believe the ComputeCpp SDK already contains a proper fix for this in `FindComputeCpp`, so a better fix would be to update the SDK submodule. However this is probably fine for now.

CMakeLists.txt
```diff
@@ -105,6 +112,19 @@ include(CmakeFunctionHelper)
...
 
+if (WIN32)
+  # On Windows, all symbols must be resolved at link time for DLLs.
+  # The sycl_blas target is just a collection of other objects so
+  # the linked libraries are not going to be propagated to this target.
+  # This requires manual linking against ComputeCpp.lib on Windows.
+  target_link_libraries(sycl_blas PRIVATE ComputeCpp::ComputeCpp)
+
+  set(BUILD_SHARED_LIBS FALSE CACHE BOOL
+    "Force SYCL-BLAS to be built as a static library on Windows"
```
ProGTX, line 123: This seems a bit strange, the place where the variable is originally set has no description, wouldn't it make more sense to have the description there, where it's not platform dependent, and just have an empty one here?

mcleary: The description is required when forcing the value on the cache so the user can't change it. I think it is better to force the value than required users manually set this to false.

CMakeLists.txt
mcleary: You are correct. I think I will update the submodule as well

mcleary: I've updated the computecpp-sdk submodule and I changed the way to set the c++ standard using cmake `set(CMAKE_CXX_STANDARD 14)` when on Windows

src/policy/sycl_policy_handler.cpp
```diff
@@ -122,7 +122,7 @@ INSTANTIATE_TEMPLATE_METHODS(double)
   template typename codeplay_policy::event_t                                  \
   PolicyHandler<codeplay_policy>::copy_to_host<IndexValueTuple<ind, val>>(    \
       BufferIterator<IndexValueTuple<ind, val>, codeplay_policy> src,         \
-      IndexValueTuple<ind, val> * dst, size_t size = 0);                      \
```
mehdi-goli, line 125: same here

src/policy/sycl_policy_handler.cpp
```diff
@@ -57,15 +57,15 @@ namespace blas {
   template typename codeplay_policy::event_t                                   \
   PolicyHandler<codeplay_policy>::copy_to_device<element_t>(                   \
       const element_t *src, BufferIterator<element_t, codeplay_policy> dst,    \
-      size_t size = 0);                                                        \
```
mehdi-goli, line 60: The idea here was not to force the user to pass the size. When the size is not passed it gets the size from the buffer and copies it. Otherwise, it uses the specific size passed by a user.

src/policy/sycl_policy_handler.cpp
```diff
@@ -57,15 +57,15 @@ namespace blas {
...
+      size_t size);                                                            \
   template typename codeplay_policy::event_t                                   \
   PolicyHandler<codeplay_policy>::copy_to_host<element_t>(                     \
       element_t * src, element_t * dst, size_t size);                          \
                                                                                \
   template typename codeplay_policy::event_t                                   \
   PolicyHandler<codeplay_policy>::copy_to_host<element_t>(                     \
       BufferIterator<element_t, codeplay_policy> src, element_t * dst,         \
-      size_t size = 0);                                                        \
```
mehdi-goli, line 68: same here

src/policy/sycl_policy_handler.cpp
```diff
@@ -113,7 +113,7 @@ INSTANTIATE_TEMPLATE_METHODS(double)
   PolicyHandler<codeplay_policy>::copy_to_device<IndexValueTuple<ind, val>>(  \
       const IndexValueTuple<ind, val> *src,                                   \
       BufferIterator<IndexValueTuple<ind, val>, codeplay_policy> dst,         \
-      size_t size = 0);                                                       \
```
mehdi-goli, line 116: Same here

mehdi-goli (commented):
Looks good to me! I have added some minor comments

src/policy/sycl_policy_handler.cpp
mcleary: The problem is that msvc doesn't seem to accept default parameters in template definitions like that. I will take another look

CodeplayBen (approved):

src/policy/sycl_policy_handler.cpp
ProGTX: It's fine to specify a default, but it cannot be done at the specialization or explicit instantiation. The best place to do it is in a forward declaration.

CodeplayBen: @mcleary Any updates on this? If we can resolve the default parameters issue I think we are good to merge this.

mcleary: I'm looking into this. As soon as I have an update I'll let you know.

mcleary: @CodeplayBen @mehdi-goli I've added the default values for the size parameters in the actual function declaration so that MSVC can compile the code. Please take a look to see if that's ok.

I also updated the computecpp-sdk submodule because there are flags in FindComputeCpp.cmake required to build SYCL code on Windows with the latest Visual Studio.

src/policy/sycl_policy_handler.cpp
mcleary: Please look at https://github.com/codeplaysoftware/sycl-blas/pull/253#discussion_r491915014

src/policy/sycl_policy_handler.cpp
mcleary: Please look at https://github.com/codeplaysoftware/sycl-blas/pull/253#discussion_r491915014

src/policy/sycl_policy_handler.cpp
mcleary: Please look at https://github.com/codeplaysoftware/sycl-blas/pull/253#discussion_r491915014

src/policy/sycl_policy_handler.cpp
mcleary: @mehdi-goli  This patch is working in Linux as well.

mehdi-goli: Thanks @mcleary

mehdi-goli (approved):

PR #255, opened 2020-08-31
Set backend for autotuner targets

StuartDAdams:
Set the correct blas backend for the autotuner.

Removed duplicate branch for adding ARM_GPU backend compiler definitions to targets.

Added a few new status messages so that it is easier to see the backend and feature support (vectorisation + tall/skinny) when running cmake.


CodeplayBen (approved):
Looks good to me!

ProGTX (approved):

cmake/CmakeFunctionHelper.cmake
```diff
@@ -52,19 +53,19 @@ function(set_target_compile_def in_target)
     target_compile_definitions(${in_target} PUBLIC ARM_GPU=1)
   elseif(${BACKEND_DEVICE} STREQUAL "RCAR")
     target_compile_definitions(${in_target} PUBLIC RCAR=1)
-  elseif(${BACKEND_DEVICE} STREQUAL "ARM_GPU")
```
mehdi-goli, line 55: You cannot remove the ARM_GPU backend from here as it is the single point of adding the compile definition for GEMM on a different target, it used for the main library, test, and benchmark. If you want to fix the auto-tuner you need to make auto-tuner to use this function instead of having a separate function.

mehdi-goli (changes requested):

mehdi-goli (approved):

PR #257, opened 2020-09-07
Adds Resnet50 ARM Configs, minor fixes

CodeplayBen:
This PR includes optimised configurations for Resnet50 Convolution/GEMM sizes on the ARM GPU backend and includes a few changes to facilitate this without potentially harming performance of other sizes:
* A new CMake variable has been added and documented, ```BLAS_MODEL_OPTIMIZATION``` which allows selecting different sets of configurations for the ARM backend. When not set these will default to configurations from commit ```d2b0b263d372139cb488f0799e384b67298b53cd``` which is the commit currently used by SYCL-DNN. Passing ```VGG_16``` uses the current ARM configurations which were optimised for those sizes and passing ```RESNET_50``` selects the new configurations. Setting this has no effect on other backends currently.
* Fixed an issue with the auto tuner python generator in that it didn't properly respect the rules around vector size of the partial and fully vectorised no local GEMM kernels.

Some formatting was applied to the auto tuner's python generator file creating some noise in the PR but was probably needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -338,7 +338,7 @@ Some of the supported options are:
 | `BLAS_ENABLE_STATIC_LIBRARY` | `ON`/`OFF` | Build as a static library (`OFF` by default) |
 | `ENABLE_EXPRESSION_TESTS` | `ON`/`OFF` | Build additional tests that use the header-only framework (e.g to test expression trees); `OFF` by default |
 | `BLAS_VERIFY_BENCHMARK` | `ON`/`OFF` | Verify the results of the benchmarks instead of only measuring the performance. See the documentation of the benchmarks for more details. `ON` by default |
-| `MODEL` | name | Pass a model name here to use optimized GEMM configurations for specific convolution models/sizes. Currently this only affects the `ARM_GPU` platform. The supported models are: `RESNET_50`, `VGG_16` |
+| `BLAS_MODEL_OPTIMIZATION` | name | Pass a model name here to use optimized GEMM configurations for specific convolution models/sizes. Currently this only affects the `ARM_GPU` platform. The supported models are: `RESNET_50`, `VGG_16` |
 
 
 ### Cross-Compile
--- a/README.md
+++ b/README.md
@@ -338,6 +338,7 @@ Some of the supported options are:
 | `BLAS_ENABLE_STATIC_LIBRARY` | `ON`/`OFF` | Build as a static library (`OFF` by default) |
 | `ENABLE_EXPRESSION_TESTS` | `ON`/`OFF` | Build additional tests that use the header-only framework (e.g to test expression trees); `OFF` by default |
 | `BLAS_VERIFY_BENCHMARK` | `ON`/`OFF` | Verify the results of the benchmarks instead of only measuring the performance. See the documentation of the benchmarks for more details. `ON` by default |
+| `MODEL` | name | Pass a model name here to use optimized GEMM configurations for specific convolution models/sizes. Currently this only affects the `ARM_GPU` platform. The supported models are: `RESNET_50`, `VGG_16` |
 
 
 ### Cross-Compile
--- a/cmake/CmakeFunctionHelper.cmake
+++ b/cmake/CmakeFunctionHelper.cmake
@@ -68,7 +68,12 @@ function(set_target_compile_def in_target)
     message(STATUS "Gemm vectorization support enabled for target ${in_target}")
     target_compile_definitions(${in_target} PUBLIC GEMM_VECTORIZATION_SUPPORT=1)
   endif()
-
+  #Set optimized model configs
+  if(${MODEL} STREQUAL "RESNET_50")
+    target_compile_definitions(${in_target} PUBLIC MODEL_RESNET_50=1)
+  elseif(${MODEL} STREQUAL "VGG_16")
+    target_compile_definitions(${in_target} PUBLIC MODEL_VGG_16=1)
+  endif()
 endfunction()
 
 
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -338,6 +338,7 @@ Some of the supported options are:
 | `BLAS_ENABLE_STATIC_LIBRARY` | `ON`/`OFF` | Build as a static library (`OFF` by default) |
 | `ENABLE_EXPRESSION_TESTS` | `ON`/`OFF` | Build additional tests that use the header-only framework (e.g to test expression trees); `OFF` by default |
 | `BLAS_VERIFY_BENCHMARK` | `ON`/`OFF` | Verify the results of the benchmarks instead of only measuring the performance. See the documentation of the benchmarks for more details. `ON` by default |
+| `MODEL` | name | Pass a model name here to use optimized GEMM configurations for specific convolution models/sizes. Currently this only affects the `ARM_GPU` platform. The supported models are: `RESNET_50`, `VGG_16` |
```
ProGTX, line 341: Not sure what the convention is, but options are often prefixed, so in this case it might be `BLAS_MODEL` or something like that.

cmake/CmakeFunctionHelper.cmake
```diff
@@ -424,6 +429,41 @@ elseif(${TARGET} STREQUAL "ARM_GPU")
     "float"
   )
   foreach(data ${supported_types})
+  if(${MODEL} STREQUAL "RESNET_50")
+  add_gemm_configuration(
```
ProGTX, line 433: Needs one level of indentation for the `if` and another for the body

cmake/CmakeFunctionHelper.cmake
```diff
@@ -424,6 +429,41 @@ elseif(${TARGET} STREQUAL "ARM_GPU")
...
+    "${data}" 64 "false" "false" "false"
+    64 4 4 8 8 1 1 1 1 "no_local" "standard" "partial" 2 "strided")
+  add_gemm_configuration(
+    "${data}" 128 "false" "false" "false"
+    64 4 8 16 8 1 1 1 1 "no_local" "standard" "partial" 4 "strided")
+  add_gemm_configuration(
+    "${data}" 16 "false" "false" "false"
+    64 4 4 4 4 1 1 1 1 "no_local" "standard" "partial" 2 "strided")
+  elseif(${MODEL} STREQUAL "VGG_16")
```
ProGTX, line 466: So the existing configurations are considered to be VGG ones?

cmake/Modules/ConfigureSYCLBLAS.cmake
```diff
@@ -59,3 +59,7 @@ endif()
 SET(TARGET "DEFAULT_CPU" CACHE STRING "Default Platform 'DEFAULT_CPU'")
 SET(BACKEND_DEVICE ${TARGET})
 message(STATUS "${TARGET} is chosen as a backend platform")
+
+# the MODEL variable defines for which model optimized configs should
```
ProGTX, line 63:
```suggestion
# the MODEL variable defines which model optimized configs should
```
There's already a `for` at the end of the sentence (next line).

src/interface/blas3/backend/backend.hpp
```diff
@@ -22,15 +22,15 @@
  *  @filename backend.hpp
  *
  **************************************************************************/
-#if RCAR
+#ifdef RCAR
```
ProGTX, line 25:
```suggestion
#if defined RCAR
```
For consistency. Or better yet:
```suggestion
#if defined(RCAR)
```

tools/auto_tuner/gen/generate_combinations.py
```diff
@@ -29,6 +29,7 @@
     "interleaved": "::blas::gemm_batch_type_t::interleaved",
 }
 
+
```
ProGTX, line 32: We should probably format all Python files at some point. Which formatter did you use? We generally use yapf in other projects, though that's not universal at all.

cmake/CmakeFunctionHelper.cmake
ProGTX: In general it seems like the CMake formatting needs a bit more consistency.

ProGTX (approved):
Looks good, some minor comments.

README.md
CodeplayBen: I think I was going for similar to ```TARGET``` but the convention is a bit inconsistent in this table! How about ```BLAS_MODEL_OPTIMIZATION``` a bit more verbose but I think it was a bit ambiguous before.

cmake/CmakeFunctionHelper.cmake
CodeplayBen: Yes the configurations as they are right now on tip are optimised for VGG_16 as far as I'm aware.

tools/auto_tuner/gen/generate_combinations.py
CodeplayBen: Ah I think ```autopep8``` was used as that's the vscode default, I can reformat with yapf though if that's better.

cmake/CmakeFunctionHelper.cmake
CodeplayBen: I'll give it a pass to make it more consistent! I'd like to have automatic formatting but the only one I found split the gemm configuration parameters to a single line for each which is pretty nasty.

tools/auto_tuner/gen/generate_combinations.py
CodeplayBen: Actually I was mistaken! The formatter in this case is ```yapf```

mehdi-goli (approved):

PR #258, opened 2020-09-10
Call quantized_copy_to_host in benchmarks

ProGTX:
The patch that quantized all tests and benchmarks (https://github.com/codeplaysoftware/sycl-blas/pull/252)
ignored the fact that `make_quantized_buffer`
cannot copy back to the original data.
This would only be possible with non-quantized types,
whereas for others it creates another buffer of a different type.

This patch addresses this by adding the missing copy-back to benchmarks.
It also adds a Doxygen note about data not being copied back.


CodeplayBen (approved):
LGTM!

mehdi-goli (approved):

PR #259, opened 2020-09-10
Perform a wait when quantizing buffer

ProGTX:
When using special data types (like `cl::sycl::half`)
calling `make_quantized_buffer` doesn't set up all required data dependencies.
This causes tests and benchmarks to fail.

This patch introduces a wait within `make_quantized_buffer`,
but only for special data types (not `float` or `double`).


CodeplayBen:
@ProGTX I'm not clear why this is needed for special types and not ```float```/```double```. Is this a ComputeCpp issue or something we need to address?

ProGTX:
I think you're correct - I thought it was just this one case (it fixed `asum`), but it doesn't fix all benchmarks, for example GEMM doesn't work no matter how many waits I insert.

The reason why it would be different for special types is because it does more work in order to quantize the data. `float` and `double` go through a "fast path" that is identical to old code.

I'm going to close this now to investigate what's really going on.

PR #261, opened 2020-09-21
Fix clblast compilation on Windows

mcleary:
Dynamic arrays are a non-standard C++ extension. This patch replaces two ocurrences of such arrays with standard C++11


mehdi-goli (approved):

CodeplayBen (approved):
LGTM!

PR #262, opened 2020-09-28
[Quantize] Use get_count to schedule quantization

ProGTX:
When launching the quantization kernel,
the wrong number of work-items was specified to `parallel_for`,
leading to out of bounds memory access.
This didn't affect `float` and `double` types
because they don't need a quantization kernel.

The fix is to only schedule as many work-items
as there are data elements (using `get_count`).

### DIFF
--- a/src/quantize/quantize.hpp
+++ b/src/quantize/quantize.hpp
@@ -82,7 +82,7 @@ struct Quantize {
           const auto kernel = QuantizeKernel<input_t, output_t>{
               quantized_input_acc_t<input_t>{input, cgh},
               quantized_output_acc_t<output_t>{output, cgh}};
-          cgh.parallel_for(cl::sycl::range<1>{input.get_size()}, kernel);
+          cgh.parallel_for(cl::sycl::range<1>{input.get_count()}, kernel);
         })};
   }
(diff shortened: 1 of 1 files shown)


CodeplayBen (approved):
LGTM!

mehdi-goli (approved):

src/quantize/quantize.hpp
```diff
@@ -82,7 +82,7 @@ struct Quantize {
           const auto kernel = QuantizeKernel<input_t, output_t>{
               quantized_input_acc_t<input_t>{input, cgh},
               quantized_output_acc_t<output_t>{output, cgh}};
-          cgh.parallel_for(cl::sycl::range<1>{input.get_size()}, kernel);
```
mehdi-goli, line 85: We usually pass the buffer in uint8_t (void* pointer style) and then apply the type inside the kernel. Unless you break this rule,  the get_count and get_size should not make any difference here.

ProGTX: I don't think quantization will work with that at the moment, it relies on having a proper type. The tests and benchmarks seem to use typed buffers?

PR #265, opened 2020-12-11
Add TRSM as a BLAS Level 3 Routine

mcleary:
- New function _trsm exposed to the user that can be used to solve Triangular Systems
- New kernel to invert the diagonal blocks of a matrix DiagonalBlocksInverter
- New utility function to fill a sycl-blas BufferIterator with a value
- Unit tests for the TRSM testing all combinations of parameters


src/interface/blas3/backend/default_cpu.hpp
```diff
@@ -94,7 +94,27 @@ typename executor_t::policy_t::event_t _gemm(
 
 #endif
 }
+
+
+
```
ProGTX, line 99: That's a lot of new empty space. Wouldn't clang-format fix that?

src/interface/blas3_interface.hpp
```diff
@@ -160,6 +160,421 @@ typename executor_t::policy_t::event_t _gemm_batched(
...
 
+/**
+ * @brief Implementation of Triangle Solve with Multiple Right Hand Sides
+ * (TRSM).
+ * @param Side Indicates if A is on the left or right of X
+ * @param Triangle If A is lower or upper triangular
+ * @param Transpose indicates the form that the matrix A will take in the
+ * multiplication
+ * @param Diagonal If A has a unit or non-unit diagonal
+ * @param M The number of rows in of matrix B, must be at least 0
```
ProGTX, line 171:
```suggestion
 * @param M The number of rows in of matrix B, must be at least 1
```

I believe `invalid_argument` is thrown if it's 0. The same for the other parameters.

src/interface/blas3_interface.hpp
```diff
@@ -160,6 +160,421 @@ typename executor_t::policy_t::event_t _gemm_batched(
...
+ * @param N The number of columns of B, must be at least 0
+ * @param alpha The scalar alpha that is applied to B
+ * @param A Buffer that hold the input matrix A
```
ProGTX, line 174:
```suggestion
 * @param A Buffer that holds the input matrix A
```

src/interface/blas3_interface.hpp
```diff
@@ -160,6 +160,421 @@ typename executor_t::policy_t::event_t _gemm_batched(
...
+ * TRSM solves one of the following matrix equations
+ *
+ * op(A)*X = alpha*B      or     X*op(A) = alpha*B
+ *
+ * where alpha is a scalar, X and B are m by n matrices, A is a unit or
+ * non-unit, upper or lower triangular matrix and op(A) is
+ *
+ * op(A) = A    or     op(A) = A^{T}
+ *
+ * The matrix X, which contains the result, is copied to B and the end.
```
ProGTX, line 193:
```suggestion
 * The matrix X, which contains the result, is copied to B at the end.
```

src/operations/blas3/trsm_ref.hpp
```diff
@@ -0,0 +1,135 @@
...
+      isInRange = (i >= j) && ((blockIndexPerBlock + i) < N_);
+    }
+    if (isInRange) {
+      const size_t srcIndex = j * lda_ + i + srcBlockOffset;
+      local[j + i * internalBlockSize] = A[srcIndex];
+    } else {
+      local[j + i * internalBlockSize] = value_t{0};
+    }
+  }
+  item.barrier(sycl::access::fence_space::local_space);
```
ProGTX, line 83: I wonder if a `mem_fence` would work here as well?

test/unittest/blas3/blas3_trsm_test.cpp
```diff
@@ -0,0 +1,114 @@
+
+#include "blas_test.hpp"
+
+template<typename scalar_t>
+using combination_t = std::tuple<int, int>;
+
+template <typename T>
+void transpose(T* A, size_t m, size_t n, size_t lda) {
```
ProGTX, line 8: Isn't there already an existing function that can do this?

test/unittest/blas3/blas3_trsm_test.cpp
```diff
@@ -0,0 +1,114 @@
...
+  std::vector<data_t> B = {
+    1, 2, -1, -2,
+    1, 2, -1, -2,
+    1, 2, -1, -2,
+    1, 2, -1, -2
+  };
+  scalar_t alpha = scalar_t{ 1 };
+  int lda = m;
+  int ldb = m;
+  // clang-format on
```
ProGTX, line 52: You can probably move `clang-format on` a big higher, I presume you just want the matrices to be nicely formatted.

test/unittest/blas3/blas3_trsm_test.cpp
```diff
@@ -0,0 +1,114 @@
...
+  auto a_gpu = blas::make_sycl_iterator_buffer<scalar_t>(A, A.size());
+  auto b_gpu = blas::make_sycl_iterator_buffer<scalar_t>(B, B.size());
+
+  char side = 'l';
+  char transA = 'n';
+  char diag = 'u';
+  char triangle = 'l';
+  _trsm(ex, side, triangle, transA, diag, m, n, alpha, a_gpu, lda, b_gpu, ldb);
+
+  // Temporary printing the solution for inspection
```
ProGTX, line 76: Will this be removed before merging or is the plan to remove it later?

test/unittest/blas3/blas3_trsm_test.cpp
```diff
@@ -0,0 +1,114 @@
...
+void run_test(const combination_t<scalar_t> combi) {
+  int m;
+  int n;
+  std::tie(m, n) = combi;
+
+  using data_t = utils::data_storage_t<scalar_t>;
+
+  // clang-format off
+  std::vector<data_t> A = {
+    1, 0, 0, 0,
```
ProGTX, line 38: It doesn't look like the input data will work with anything else than `combi == {4, 4}`. I think that's fine, but we should probably have a TODO to generate the matrix using a function or something.

ProGTX (commented):
Left a few comments, but looks good to me

test/unittest/blas3/blas3_trsm_test.cpp
mcleary: No, but this is temporary and will be removed soon

test/unittest/blas3/blas3_trsm_test.cpp
mcleary: The clang-format will be removed. This was just to test a small matrix

src/operations/blas3/trsm_ref.hpp
mehdi-goli: I dont think it is possible to use mem_fence here. The barrier here is for the synchronization to make sure that the below computation is not started until all the data are loaded. In order to avoid the race condition in accessing the local memory by different threads.  However, mem_fence is to guarantee loads and stores preceding the mem_fence will be committed to memory before any loads and stores following the mem_fence.

include/operations/blas3_trees.h
```diff
@@ -249,6 +249,63 @@ make_gemm(input_t buffer_a, input_t buffer_b, output_t buffer_c,
...
+ * Since we only invert the diagonal blocks of A, B must be a buffer that can
+ * store multiples of blockSize*blockSize.
+ *
+ * @Note This kernel assumes the column-major matrices
+ * @Note This kernel uses fixed size blocks of 16, but this can be changed
+ */
+template <bool UnitDiag, bool Upper, typename matrix_t>
+struct DiagonalBlocksInverter {
+  using index_t = size_t;
+  static constexpr index_t internalBlockSize = 16;
```
mehdi-goli, line 284: Is this fixed. It would be better to make it a template parameter if it needed to be changed for different devices

include/operations/blas3_trees.h
```diff
@@ -249,6 +249,63 @@ make_gemm(input_t buffer_a, input_t buffer_b, output_t buffer_c,
...
+  static constexpr index_t outterBlockSize = 16;
```
mehdi-goli, line 285: Same as above

src/operations/blas3/trsm_ref.hpp
```diff
@@ -0,0 +1,135 @@
...
+      isInRange = (i <= j) && (blockIndexPerBlock + j < N_);
+    } else {
+      isInRange = (i >= j) && ((blockIndexPerBlock + i) < N_);
+    }
+    if (isInRange) {
+      const size_t srcIndex = j * lda_ + i + srcBlockOffset;
+      local[j + i * internalBlockSize] = A[srcIndex];
+    } else {
+      local[j + i * internalBlockSize] = value_t{0};
+    }
```
mehdi-goli, line 81:
```suggestion
local[j + i * internalBlockSize] = (isInRange) ?  A[j * lda_ + i + srcBlockOffset] : value_t{0};
```

src/operations/blas3/trsm_ref.hpp
```diff
@@ -0,0 +1,135 @@
...
+
+  // Loads the source lower triangle into local memory. Any values in the upper
+  // triangle or outside of the matrix are set to zero
+  for (size_t j = 0; j < internalBlockSize; ++j) {
+    bool isInRange = false;
+    if (upper) {
+      isInRange = (i <= j) && (blockIndexPerBlock + j < N_);
+    } else {
+      isInRange = (i >= j) && ((blockIndexPerBlock + i) < N_);
+    }
```
mehdi-goli, line 75:
```suggestion
    bool isInRange = (upper) ?  (i <= j) && (blockIndexPerBlock + j < N_): (i >= j) && ((blockIndexPerBlock + i) < N_);
```

src/operations/blas3/trsm_ref.hpp
```diff
@@ -0,0 +1,135 @@
+
```
mehdi-goli, line 1: Is there any reason you call this file `trsm_ref.hpp` I mean what `ref` stands for?

src/interface/blas3/backend/default_cpu.hpp
mcleary: Already removed

src/interface/blas3_interface.hpp
mcleary: Fixed

src/interface/blas3_interface.hpp
mcleary: Fixed

test/unittest/blas3/blas3_trsm_test.cpp
mcleary: Removed

test/unittest/blas3/blas3_trsm_test.cpp
mcleary: Fixed

test/unittest/blas3/blas3_trsm_test.cpp
mcleary: It was already removed

test/unittest/blas3/blas3_trsm_test.cpp
mcleary: Already removed

src/operations/blas3/trsm_ref.hpp
mcleary: I changed it to `trsm.hpp`

mcleary:
@mehdi-goli The structure of TRSM was changed to accommodate the change you requested. Please let me know if this is what you had in mind.

src/interface/trsm.hpp
```diff
@@ -0,0 +1,432 @@
...
+        std::ptrdiff_t offsetB = i;
+        std::ptrdiff_t offsetC = i;
+        auto gemmBufferA = invA + offsetA;
+        auto gemmBufferB = B + offsetB;
+        auto gemmBufferC = X + offsetC;
+        auto gemmEvent =
+            internal::_gemm(ex, gemmTransA, gemmTransB, gemmM, gemmN, gemmK,
+                            gemmAlpha, gemmBufferA, gemmLda, gemmBufferB,
+                            gemmLdb, gemmBeta, gemmBufferC, gemmLdc);
+        trsmEvents = concatenate_vectors(trsmEvents, gemmEvent);
```
mehdi-goli, line 203: These operations are repeated multiple times. Can you wrap them in a function and call the function instead? you can create a function `call_gemm` to do the operation

src/interface/trsm.hpp
```diff
@@ -0,0 +1,432 @@
...
+        offsetB = i;
+        offsetC = i + blockSize;
+        gemmBufferA = A + offsetA;
+        gemmBufferB = X + offsetB;
+        gemmBufferC = B + offsetC;
+        gemmEvent =
+            internal::_gemm(ex, gemmTransA, gemmTransB, gemmM, gemmN, gemmK,
+                            gemmAlpha, gemmBufferA, gemmLda, gemmBufferB,
+                            gemmLdb, gemmBeta, gemmBufferC, gemmLdc);
+        trsmEvents = concatenate_vectors(trsmEvents, gemmEvent);
```
mehdi-goli, line 230: same as above. These operations are repeated multiple times. Can you wrap them in a function and call the function instead?

src/interface/trsm.hpp
```diff
@@ -0,0 +1,432 @@
...
+        std::ptrdiff_t gemmOffsetB = i;
+        std::ptrdiff_t gemmOffsetC = i;
+        auto gemmBufferA = invA + gemmOffsetA;
+        auto gemmBufferB = B + gemmOffsetB;
+        auto gemmBufferC = X + gemmOffsetC;
+        auto gemmEvent =
+            internal::_gemm(ex, gemmTransA, gemmTransB, gemmM, gemmN, gemmK,
+                            gemmAlpha, gemmBufferA, gemmLda, gemmBufferB,
+                            gemmLdb, gemmBeta, gemmBufferC, gemmLdc);
+        trsmEvents = concatenate_vectors(trsmEvents, gemmEvent);
```
mehdi-goli, line 262: Same as above

src/interface/trsm.hpp
```diff
@@ -0,0 +1,432 @@
...
+        gemmOffsetB = i;
+        gemmOffsetC = 0;
+        gemmBufferA = A + gemmOffsetA;
+        gemmBufferB = X + gemmOffsetB;
+        gemmBufferC = B + gemmOffsetC;
+        gemmEvent =
+            internal::_gemm(ex, gemmTransA, gemmTransB, gemmM, gemmN, gemmK,
+                            gemmAlpha, gemmBufferA, gemmLda, gemmBufferB,
+                            gemmLdb, gemmBeta, gemmBufferC, gemmLdc);
+        trsmEvents = concatenate_vectors(trsmEvents, gemmEvent);
```
mehdi-goli, line 288: Same as above

src/interface/trsm.hpp
```diff
@@ -0,0 +1,432 @@
...
+        std::ptrdiff_t gemmOffsetC = i * ldx;
+        auto gemmBufferA = B + gemmOffsetA;
+        auto gemmBufferB = invA + gemmOffsetB;
+        auto gemmBufferC = X + gemmOffsetC;
+        auto gemmEvent =
+            internal::_gemm(ex, gemmTransA, gemmTransB, gemmM, gemmN, gemmK,
+                            gemmAlpha, gemmBufferA, gemmLda, gemmBufferB,
+                            gemmLdb, gemmBeta, gemmBufferC, gemmLdc);
+        trsmEvents = concatenate_vectors(trsmEvents, gemmEvent);
+
```
mehdi-goli, line 326: Same as above these operations are repeated multiple times. Can you wrap them in a function and call the function instead?

src/interface/trsm.hpp
```diff
@@ -0,0 +1,432 @@
...
+        gemmOffsetB = !isTranspose ? i : (i * lda);
+        gemmOffsetC = 0;
+        gemmBufferA = X + gemmOffsetA;
+        gemmBufferB = A + gemmOffsetB;
+        gemmBufferC = B + gemmOffsetC;
+        gemmEvent =
+            internal::_gemm(ex, gemmTransA, gemmTransB, gemmM, gemmN, gemmK,
+                            gemmAlpha, gemmBufferA, gemmLda, gemmBufferB,
+                            gemmLdb, gemmBeta, gemmBufferC, gemmLdc);
+        trsmEvents = concatenate_vectors(trsmEvents, gemmEvent);
```
mehdi-goli, line 351: Same here

src/interface/trsm.hpp
```diff
@@ -0,0 +1,432 @@
...
+        std::ptrdiff_t gemmOffsetC = i * ldx;
+        auto gemmBufferA = B + gemmOffsetA;
+        auto gemmBufferB = invA + gemmOffsetB;
+        auto gemmBufferC = X + gemmOffsetC;
+        auto gemmEvent =
+            internal::_gemm(ex, gemmTransA, gemmTransB, gemmM, gemmN, gemmK,
+                            gemmAlpha, gemmBufferA, gemmLda, gemmBufferB,
+                            gemmLdb, gemmBeta, gemmBufferC, gemmLdc);
+        trsmEvents = concatenate_vectors(trsmEvents, gemmEvent);
+
```
mehdi-goli, line 382: Same here

src/interface/trsm.hpp
```diff
@@ -0,0 +1,432 @@
...
+                                   : (i + blockSize) + (i * lda);
+        gemmOffsetC = (i + blockSize) * ldb;
+        gemmBufferA = X + gemmOffsetA;
+        gemmBufferB = A + gemmOffsetB;
+        gemmBufferC = B + gemmOffsetC;
+        gemmEvent =
+            internal::_gemm(ex, gemmTransA, gemmTransB, gemmM, gemmN, gemmK,
+                            gemmAlpha, gemmBufferA, gemmLda, gemmBufferB,
+                            gemmLdb, gemmBeta, gemmBufferC, gemmLdc);
+        trsmEvents = concatenate_vectors(trsmEvents, gemmEvent);
```
mehdi-goli, line 408: Same here

src/interface/trsm.hpp
```diff
@@ -0,0 +1,432 @@
+
```
mehdi-goli, line 1: Can you call this fine trsm_interface.hpp. Since the operation has trsm.hpp, when giving relative path it may get confusion

src/interface/gemm.hpp
```diff
@@ -19,12 +19,10 @@
  *
```
mehdi-goli, line 19: Same as here can you call this gemm_interface.hpp as in the operation there are `operation/gemm.hpp` that may cause confusion when giving relative path

src/operations/blas3/trsm.hpp
```diff
@@ -0,0 +1,158 @@
...
+      isInRange = (i <= j) && (blockIndexPerBlock + j < N_);
+    } else {
+      isInRange = (i >= j) && ((blockIndexPerBlock + i) < N_);
+    }
+    if (isInRange) {
+      const size_t srcIndex = j * lda_ + i + srcBlockOffset;
+      local[j + i * internalBlockSize] = A[srcIndex];
+    } else {
+      local[j + i * internalBlockSize] = value_t{0};
+    }
```
mehdi-goli, line 103:
```suggestion
    local[j + i * internalBlockSize] = (isInRange) ?  A[ j * lda_ + i + srcBlockOffset] : value_t{0};
```

src/operations/blas3/trsm.hpp
```diff
@@ -0,0 +1,158 @@
...
+
+  // Loads the source lower triangle into local memory. Any values in the upper
+  // triangle or outside of the matrix are set to zero
+  for (size_t j = 0; j < internalBlockSize; ++j) {
+    bool isInRange = false;
+    if (upper) {
+      isInRange = (i <= j) && (blockIndexPerBlock + j < N_);
+    } else {
+      isInRange = (i >= j) && ((blockIndexPerBlock + i) < N_);
+    }
```
mehdi-goli, line 97:
```suggestion
    isInRange = (upper) ? (i <= j) && (blockIndexPerBlock + j < N_) : (i >= j) && ((blockIndexPerBlock + i) < N_);
```

src/operations/blas3/trsm.hpp
```diff
@@ -0,0 +1,158 @@
...
+      if (i < j) {
+        for (cl::sycl::cl_int k = 0; k < j; ++k) {
+          sum = cl::sycl::mad(local[k + i * internalBlockSize],
+                          local[j + k * internalBlockSize], sum);
+        }
+      }
+      item.barrier(cl::sycl::access::fence_space::local_space);
+      if (i < j) {
+        const value_t negDiagValue = -local[j + j * internalBlockSize];
+        local[j + i * internalBlockSize] = sum * negDiagValue;
```
mehdi-goli, line 126:
```suggestion
        local[j + i * internalBlockSize] = sum * (-local[j + j * internalBlockSize]);
```

src/operations/blas3/trsm.hpp
```diff
@@ -0,0 +1,158 @@
...
+      if (i > j) {
+        for (cl::sycl::cl_int k = j + 1; k < internalBlockSize; ++k) {
+          sum = cl::sycl::mad(local[k + i * internalBlockSize],
+                          local[j + k * internalBlockSize], sum);
+        }
+      }
+      item.barrier(cl::sycl::access::fence_space::local_space);
+      if (i > j) {
+        const value_t negDiagValue = -local[j + j * internalBlockSize];
+        local[j + i * internalBlockSize] = sum * negDiagValue;
```
mehdi-goli, line 143:
```suggestion
        local[j + i * internalBlockSize] = sum * (-local[j + j * internalBlockSize]);
```

src/operations/blas3/trsm.hpp
```diff
@@ -0,0 +1,158 @@
...
+      local[j + i * internalBlockSize] = value_t{0};
+    }
+  }
+  item.barrier(cl::sycl::access::fence_space::local_space);
+
+  // Inverts the diagonal elements
+  if (!unitDiag_) {
+    local[i + i * internalBlockSize] =
+        value_t{1} / local[i + i * internalBlockSize];
+    item.barrier(cl::sycl::access::fence_space::local_space);
```
mehdi-goli, line 111: I know the `unitDiag_` is compile-time value and all threads will go through the if since the `unitDiag_` is the same for all threads, but we have seen in the past that some compiler had an issue putting barrier inside the if statement, even it is trivial and all taking the same path. Try to see if this could cause a problem on devices that we are supporting

src/operations/blas3/trsm.hpp
```diff
@@ -0,0 +1,158 @@
...
+
+template <bool UnitDiag, bool Upper, typename matrix_t>
+template <typename local_memory_t>
+SYCL_BLAS_INLINE void DiagonalBlocksInverter<UnitDiag, Upper, matrix_t>::eval(
+    local_memory_t localMem, cl::sycl::nd_item<1> item) noexcept {
+  auto A = A_.get_data().get_pointer() + A_.get_access_displacement();
+  auto invA = invA_.get_data().get_pointer() + invA_.get_access_displacement();
+  value_t* local = localMem.localAcc.get_pointer();
+
+  const size_t threadIndex = item.get_local_id(0);
```
mehdi-goli, line 69: Is there any reason you are using size_t here for index? for the rest of the BLAS we are using `int` as BLAS interface strictly saying that the index type should be `signed int`. Can you convert all the index type to be the template parameter  `indext_t` the one that has been passed by the user in calling TRSM and propagate it to this class instead of hardcode `size_t`. So, in this case, all the value type are `value_t` and the index types are `index_t` so you don't need to convert `for` loop index to have an arbitrary type and all will be `index_t` and no hardcoded type will be in the kernel. This will allow us to later switch the index type to whatever is required

src/operations/blas3/trsm.hpp
```diff
@@ -0,0 +1,158 @@
...
+
+  // Inverts the diagonal elements
+  if (!unitDiag_) {
+    local[i + i * internalBlockSize] =
+        value_t{1} / local[i + i * internalBlockSize];
+    item.barrier(cl::sycl::access::fence_space::local_space);
+  }
+
+  if (upper) {
+    for (cl::sycl::cl_int j = 1; j < internalBlockSize; ++j) {
```
mehdi-goli, line 115: can you convert the `cl::sycl::cl_int` to `index_t` as well

src/operations/blas3/trsm.hpp
```diff
@@ -0,0 +1,158 @@
...
+      value_t sum = value_t{0};
+      if (i < j) {
+        for (cl::sycl::cl_int k = 0; k < j; ++k) {
```
mehdi-goli, line 118: can you convert the `cl::sycl::cl_int` to `index_t` as well

src/operations/blas3/trsm.hpp
```diff
@@ -0,0 +1,158 @@
...
+      item.barrier(cl::sycl::access::fence_space::local_space);
+      if (i < j) {
+        const value_t negDiagValue = -local[j + j * internalBlockSize];
+        local[j + i * internalBlockSize] = sum * negDiagValue;
+      }
+      item.barrier(cl::sycl::access::fence_space::local_space);
+    }
+  } else {
+    // Computes the elements j+1:internalBlock-1 of the j-th column
+    for (cl::sycl::cl_int j = internalBlockSize - 2; j >= 0; --j) {
```
mehdi-goli, line 132: can you convert the `cl::sycl::cl_int` to `index_t` as well

src/operations/blas3/trsm.hpp
```diff
@@ -0,0 +1,158 @@
...
+      value_t sum = value_t{0};
+      if (i > j) {
+        for (cl::sycl::cl_int k = j + 1; k < internalBlockSize; ++k) {
```
mehdi-goli, line 135: can you convert the `cl::sycl::cl_int` to `index_t` as well

src/operations/blas3/trsm.hpp
```diff
@@ -0,0 +1,158 @@
...
+      if (i > j) {
+        const value_t negDiagValue = -local[j + j * internalBlockSize];
+        local[j + i * internalBlockSize] = sum * negDiagValue;
+      }
+      item.barrier(cl::sycl::access::fence_space::local_space);
+    }
+  }
+
+  // Writes the result to global memory
+  for (cl::sycl::cl_int j = 0; j < internalBlockSize; ++j) {
```
mehdi-goli, line 150: can you convert the `cl::sycl::cl_int` to `index_t` as well

include/operations/blas3_trees.h
```diff
@@ -249,6 +249,63 @@ make_gemm(input_t buffer_a, input_t buffer_b, output_t buffer_c,
...
+ *
+ * Since we only invert the diagonal blocks of A, B must be a buffer that can
+ * store multiples of blockSize*blockSize.
+ *
+ * @Note This kernel assumes the column-major matrices
+ * @Note This kernel uses fixed size blocks of 16, but this can be changed
+ */
+template <bool UnitDiag, bool Upper, typename matrix_t>
+struct DiagonalBlocksInverter {
+  using index_t = size_t;
```
mehdi-goli, line 283: The BLAS index explicitly said to be signed int in BLAS standard interface. the index type should be propagated from the _trsm function passed by the user to this function. Can you please propagate `indext_t` here and on any functions that you hardcoded `size_t`

test/blas_test.hpp
```diff
@@ -99,6 +99,19 @@ inline cl::sycl::queue make_queue() {
   return queue;
 }
 
+/**
+ * @brief Generates a random scalar in the specified range
+ * @param rangeMin range minimum
+ * @param rangeMax range maximum
+ */
+template <typename scalar_t>
+static inline scalar_t random_scalar(scalar_t rangeMin, scalar_t rangeMax) {
```
mehdi-goli, line 108: BLAS has `fill_random` function that accepts the vector and fills it. To unify all the BLAS data to follow the same random seed and random generator algorithm, can you make them align with each other. e,g  `fill_random` can use the random_scaler to populate each element and you can replace the seed generator here with what is  used in `fill_random`

test/unittest/blas3/blas3_trsm_test.cpp
```diff
@@ -0,0 +1,95 @@
...
+  }
+  // Copy the verification result to X
+  ex.get_policy_handler().wait(
+      ex.get_policy_handler().copy_to_host(x_gpu, X.data(), X.size()));
+
+  ASSERT_TRUE(utils::compare_vectors(X, B));
+  ex.get_policy_handler().wait();
+}
+
+const auto combi = ::testing::Combine(::testing::Values(4, 7, 12, 16),  // m
```
mehdi-goli, line 87: Can you try more combinations with a larger size that can populate the entire GPU. See the sizes used in BLAS and try to add them also you need to consider LDA, etc

test/unittest/blas3/blas3_trsm_test.cpp
```diff
@@ -0,0 +1,95 @@
...
+  const int sizeB = n * ldb;
+
+  std::vector<scalar_t> A(sizeA);
+  std::vector<scalar_t> B(sizeB);
+
+  const scalar_t diagValue =
+      diag == 'u' ? scalar_t{1} : random_scalar(scalar_t{1}, scalar_t{10});
+
+  fill_triangle(A, k, k, lda, triangle, diagValue);
+  fill_random(B);
```
mehdi-goli, line 58: This is one of the examples that you need all random function to use the same seed and algorithm to generate so that it will be reproducible and predictable, So in this case B and A will use the same random seed to generate the matrix

src/interface/blas3/backend/default_cpu.hpp
```diff
@@ -94,6 +94,7 @@ typename executor_t::policy_t::event_t _gemm(
 
 #endif
 }
+
```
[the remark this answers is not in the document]
mcleary: Fixed

src/interface/blas3_interface.hpp
mcleary: Fixed

src/operations/blas3/trsm.hpp
mcleary: Done

src/operations/blas3/trsm.hpp
mcleary: Done

src/operations/blas3/trsm.hpp
mcleary: Done

src/operations/blas3/trsm.hpp
mcleary: Done

src/operations/blas3/trsm.hpp
mcleary: Done

test/unittest/blas3/blas3_trsm_test.cpp
mcleary: I can, but the INTEL_GPU backend is not building for me so I am investigating that. If I have `M>16` gemm fails with incorrect workgroup size

src/operations/blas3/trsm.hpp
mcleary: Done

src/operations/blas3/trsm.hpp
mcleary: Done

src/operations/blas3/trsm.hpp
mehdi-goli: You can see what we have done in gemm. You can see it there : https://github.com/codeplaysoftware/sycl-blas/blob/master/src/operations/blas3/gemm_local.hpp#L270
You can cast it to signed if it gave you a warning

mehdi-goli: The more important thing is the combination of signed and unsigned should not exist

test/unittest/blas3/blas3_trsm_test.cpp
mehdi-goli: That might be some bug in the breaking and passing size to gemm. Gemm has a static_assert to avoid incorrect workgroup/tile sizes

src/interface/gemm.hpp
mcleary: Renamed to `interface/gemm_interface.hpp`

src/interface/trsm.hpp
mcleary: What would be the advantage of wrapping this in a function? All this is doing is calling gemm with different parameters depending on which block should be multiplied. Having this is a function would just encapsulate gemm.

src/interface/trsm.hpp
mcleary: They are repeated multiple times but with different parameters

src/interface/trsm.hpp
mcleary: Renamed to `interface/trsm_interface.hpp`

test/unittest/blas3/blas3_trsm_test.cpp
```diff
@@ -0,0 +1,95 @@
...
+  int n;
+  char transA;
+  char side;
+  char diag;
+  char triangle;
+  std::tie(m, n, transA, side, diag, triangle) = combi;
+
+  using data_t = utils::data_storage_t<scalar_t>;
+
+  scalar_t alpha = scalar_t{1};
```
CodeplayBen, line 43: We should probably include alpha as part of the testing parameters for completeness.

test/unittest/blas3/blas3_trsm_test.cpp
```diff
@@ -0,0 +1,95 @@
...
+  auto b_gpu = blas::make_sycl_iterator_buffer<scalar_t>(B, B.size());
+
+  _trsm(ex, side, triangle, transA, diag, m, n, alpha, a_gpu, lda, b_gpu, ldb);
+
+  std::vector<data_t> X(B.size(), data_t{0});
+
+  auto x_gpu = blas::make_sycl_iterator_buffer<scalar_t>(X.size());
+  if (side == 'l') {
+    // Verification of the result. AX = B
+    _gemm(ex, transA, 'n', m, n, m, data_t{1}, a_gpu, lda, b_gpu, ldb,
```
CodeplayBen, line 72: We should perhaps be using the reference blas TRSM to verify results here in line with other tests such as GEMM.

CodeplayBen (commented):
Couple of minor comments but otherwise looks good to me!

include/operations/blas3_trees.h
mcleary: Changed into a template parameter

include/operations/blas3_trees.h
mcleary: Changed to a template parameter

src/operations/blas3/trsm.hpp
mcleary: Everything was changed to `index_t`

src/operations/blas3/trsm.hpp
mcleary: The tests are supposed to fail if there is a problem with this barrier so we can see if this causes a problem

test/unittest/blas3/blas3_trsm_test.cpp
mcleary: The idea of using `_gemm` to verify the results is to simplify this test as it doesn't need to rely on a external library call to check if the result is correct. This is, in fact, just verifying that the output vector is in fact **a** solution for the system. If we change the internal algorithm later, this could potentially return a different, but valid, solution. Using _gemm just makes sure the result is a valid solution for the system, and _gemm is already tested.

include/operations/blas3_trees.h
mcleary: This was changed to a signed type

test/unittest/blas3/blas3_trsm_test.cpp
mehdi-goli: This is incorrect, depending on the type of TRSM the value for the upper or lower triangle can have an invalid value or can be padded with NAN. Your test should cover this case as well. If you use GEMM in this case you can get the wrong results.  Also, all other tests in the BLAS test against an external library. Please follow the same rule applied on other unit tests and add the nan tests and test against the external TRSM from the BLAS library that is already included in CMake.

src/operations/blas3/trsm.hpp
mehdi-goli: yes, What I am saying is that have you run them internally to see if it is not failing. Since travis only test on one simple default CPU

test/unittest/blas3/blas3_trsm_test.cpp
mehdi-goli: discussed internally. We need to create a bug report for GEMM when double type is used to match the cache line size with tile sizes

test/unittest/blas3/blas3_trsm_test.cpp
mcleary: Already included

test/unittest/blas3/blas3_trsm_test.cpp
mcleary: Is there an existing test that tests that uses NaN values for me to take as reference?

mcleary: > We should perhaps be using the reference blas TRSM to verify results here in line with other tests such as GEMM.

I've replaced the use of GEMM to verify the result with a call to system blas.

mcleary: Added test with NaN values in the input matrix

test/unittest/blas3/blas3_trsm_test.cpp
mcleary: The internals of the fill triangle and fill random were updated to use the same random generator engine

test/blas_test.hpp
mcleary: See https://github.com/codeplaysoftware/sycl-blas/pull/265#discussion_r547370173

src/interface/trsm.hpp
ProGTX: I was thinking the same thing when I first saw the code, but I realized most of the code is setting up parameters so it doesn't seem like the amount of code can be reduced.

test/unittest/blas3/blas3_trsm_test.cpp
ProGTX: I believe this can be resolved for now?

src/interface/gemm_interface.hpp
```diff
@@ -19,12 +19,10 @@
...
  *  SYCL-BLAS: BLAS implementation using SYCL
  *
- *  @filename blas3_interface.hpp
- *
  **************************************************************************/
 
-#ifndef SYCL_BLAS_BLAS3_INTERFACE_HPP
-#define SYCL_BLAS_BLAS3_INTERFACE_HPP
+#ifndef SYCL_BLAS_BLAS3_GEMM_HPP
+#define SYCL_BLAS_BLAS3_GEMM_HPP
```
ProGTX, line 25:
```suggestion
#define SYCL_BLAS_GEMM_INTERFACE_HPP
```

To match the file name

ProGTX (approved):

src/interface/trsm.hpp
mcleary: What @mehdi-goli wants is for all the parameters to be inline in the `_gemm` call. I don't like this.

test/unittest/blas3/blas3_trsm_test.cpp
mehdi-goli: The proper testing for size has not been added yet. The main type that we care about is float. Adding double for SYCL-BLAS is experimental. And we dont have any cases that uses double. That is the reason Double test path is disable by default. The test size added here is not adequate for float, so it requires more test to see the algorithm actually works. Once different sizes similar to GEMM has been added we can solve the case

src/interface/trsm.hpp
mehdi-goli: There are lots of boilerplate variable here which is not needed. e.g `int gemmN=N;` is boilerplate. `int gemmLdb=ldb;` is boilerplate, etc. there are full of them. if you write a call to gemm function`_gemm(...., currentBlock, N, currenrtBlock,..., B+i, ....);` everyone who is familiar with linear algebra and gemm understand what it says. GEMM algorithm and the call to it is standard. This is part of the internal code and anyone works with it is expected to understand gemm and linear algebra. if you want to be expressive you should put some comments on top of each block and explain which block each gemm is computing.

test/unittest/blas3/blas3_trsm_test.cpp
mcleary: This is outdated. In the latest version we are testing 7, 16, 70, 300. It is easy to add more but I don't it is necessary for the tests. Maybe for benchmarks but not tests.

src/interface/gemm_interface.hpp
```diff
@@ -19,12 +19,10 @@
  *
  *  SYCL-BLAS: BLAS implementation using SYCL
  *
- *  @filename blas3_interface.hpp
- *
  **************************************************************************/
 
-#ifndef SYCL_BLAS_BLAS3_INTERFACE_HPP
-#define SYCL_BLAS_BLAS3_INTERFACE_HPP
+#ifndef SYCL_BLAS_BLAS3_GEMM_HPP
```
mehdi-goli, line 24: Also:
```suggestion
#ifndef SYCL_BLAS_BLAS3_GEMM_INTERFACE_HPP
```

test/unittest/blas3/blas3_trsm_test.cpp
mehdi-goli: For the test in order to caver various sizes can you add one small size the original one that you have one medium size the one you have added now, and one large size. This will make sure that the Work-item/Workgroup/ComputeUnit over/under population has been correctly considered in creating the global/local thread size

mcleary: * Which size should I add that is smaller than 7x7?
* Which size should I add that is larger than 300x300?

mehdi-goli: let's say 7X7, 513x513, 1027x1027

mcleary: Test sizes updated

src/interface/trsm.hpp
mcleary: > There are lots of boilerplate variable here which is not needed

This boilerplate code, despite not being strictly needed makes the code a lot more readable than a call to a function that has 16 inline parameters. It was a lot easier to write this code having all the parameters in separate variables.

> everyone who is familiar with linear algebra and gemm understand what it says

This is an assumption that breaks really easily. Yes, there is a call to gemm there but where it is operating and what is it doing is far from obvious. It is a lot easier to see what the value of each argument is, which in turns makes it easier to think about what the operation is doing

> GEMM algorithm and the call to it is standard

Being standard doesn't mean it should be difficult to read.

> This is part of the internal code and anyone works with it is expected to understand gemm and linear algebra

Exactly because it is internal code it should be easier to maintain. 

> you should put some comments on top of each block and explain which block each gemm is computing

I agree with this, good comments are always good, but readable code requires less comments because the code is easier to read and understand.

With this said, I will proceed and make the change you want but I still don't like it.

mcleary: changes are in 108700435ecdf3742e87b6e9c869c03be20e0881

ProGTX: IMO it's more readable this way (with named arguments) because you don't have to go to the definition of the GEMM function to figure out what each one does.

mehdi-goli: > With this said, I will proceed and make the change you want but I still don't like it.
Thank you

mehdi-goli: > I agree with this, good comments are always good,
Let me know when the comment is added so we can resolve this issue

mehdi-goli: > IMO it's more readable this way (with named arguments) because you don't have to go to the definition of the GEMM function to figure out what each one does.

If it is too hard to read gemm parameters, we can comment out gemm definition on top of the code rather than adding boilerplates. In this case, you don't have boilerplates and can easily see one to one mapping of the parameters without going to the gemm definition

mcleary: Comments added in cfcd84278807ab512c361f6bfe60f2eedccef95f

mehdi-goli (approved):
Thank you @mcleary for your hard work. The PR looks good to me! Later on, we have to benchmark it  to see the performance and further improvement. But the this process is not part of this PR

PR #268, opened 2021-01-08
Add sub-group configuration parameters

npmiller:


include/executors/kernel_constructor.h
```diff
@@ -33,10 +33,17 @@ namespace blas {
 /*!using_local_memory.
  * @brief Indicates whether if the kernel uses shared memory or not.
  This is a work-around for the following enum class.
+ //enum class using_shared_mem { enabled, disabled };
```
ProGTX, line 36: This isn't needed, right?

ProGTX (approved):

include/executors/kernel_constructor.h
npmiller: I believe it's the `enum class` referred to by the rest of the comment

mehdi-goli: If the member of the `enum` class changed can you either remove this or update the comment

src/executors/kernel_constructor.hpp
```diff
@@ -80,6 +80,40 @@ struct LocalMemory<value_t, using_local_memory::disabled> {
...
+  @brief Constructor that creates a local accessor from a size and a SYCL
+  command group handler.
+  @param size Size in elements of the local accessor.
+  @param cgh SYCL command group handler.
+  */
+  SYCL_BLAS_INLINE LocalMemory(size_t size, cl::sycl::handler &cgh)
+      : subgroupAcc(cl::sycl::range<1>(size), cgh) {}
+
+  /*!
+  @brief Subscirpt operator that forwards on to the subgroup accessor subscript
```
CodeplayBen, line 101:
```suggestion
  @brief Subscript operator that forwards on to the subgroup accessor subscript
```

CodeplayBen (approved):
Looks good to me!

mehdi-goli:
Travis Faild with 
```bash
../../libsycl_blas.so.0.1.0: undefined reference to `blas::Executor<blas::PolicyHandler<blas::codeplay_policy> >::policy_t::event_t blas::Gemm_Launcher<64, false, false, false, 64, blas::Tile<2, 2, 4, 4, 1, 1, 4, 4, 1, 1>, true, false, 1, 1, 2, true, 4, 1>::_select_gemm<blas::Executor<blas::PolicyHandler<blas::codeplay_policy> >, blas::BufferIterator<float, blas::codeplay_policy>, blas::BufferIterator<float, blas::codeplay_policy>, blas::BufferIterator<float, blas::codeplay_policy>, float, int>(blas::Executor<blas::PolicyHandler<blas::codeplay_policy> >&, int, int, int, float, blas::BufferIterator<float, blas::codeplay_policy>, int, blas::BufferIterator<float, blas::codeplay_policy>, int, float, blas::BufferIterator<float, blas::codeplay_policy>, int, int)'
collect2: error: ld returned 1 exit status
```
Seems to be an issue with adding the subgroup to the tile location: perhaps Missmatching for intel CPU. as Travis will only build it for intel CPU. Could you please have a look at that? also, can you check for other devices with the internal CI that we have so that we can confirm all the configs added correctly

include/executors/kernel_constructor.h
npmiller: @CodeplayBen do you know more about this, was this what is now named  `using_local_memory`?

include/operations/blas3_trees.h
```diff
@@ -108,12 +108,14 @@ enum class gemm_batch_type_t : int { strided = 0, interleaved = 1 };
  *
  * @see Gemm
  */
-template <int ItemRows = 8, int ItemCols = 8, int WgRows = 16, int WgCols = 16,
+template <int ItemRows = 8, int ItemCols = 8,  int SgRows = 1, int SgCols = 1, int WgRows = 16, int WgCols = 16,
```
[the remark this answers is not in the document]
npmiller: Basic doxygen added

include/executors/kernel_constructor.h
CodeplayBen: Seems like it but it's been this way for 2 years at least according to the history. The enum class is mentioned in a few other comments in this file so I think they've just not been updated/tidied up correctly. I would probably remove this line specifically as I don't think it's necessary.

npmiller: Ok, line removed

src/executors/kernel_constructor.hpp
npmiller: Should be fixed now

PR #269, opened 2021-01-08
Add more TRSM bench configs, organise config_csv folder

CodeplayBen:
Adds several more configs for use with the TRSM benchmark - small and large sizes as well as one varying the other non-size parameters. The blas3 benchmark configurations have now also been reorganised into separate folders for each function.


mehdi-goli (approved):

mehdi-goli:
Looks good to me! However, the internal CI may need to change since the location of the files used for benchmarking has been changed. Can you check if it is done before merging, otherwise it may crash the CI

CodeplayBen:
> Looks good to me! However, the internal CI may need to change since the location of the files used for benchmarking has been changed. Can you check if it is done before merging, otherwise it may crash the CI

@mehdi-goli I've looked into the CI and I don't believe it is an issue as it doesn't actually use these csv files but instead generates them itself. The TRSM benchmarks will need to be added to it but that is a separate issue and shouldn't prevent this from being merged.

mehdi-goli:
> > Looks good to me! However, the internal CI may need to change since the location of the files used for benchmarking has been changed. Can you check if it is done before merging, otherwise it may crash the CI
> 
> @mehdi-goli I've looked into the CI and I don't believe it is an issue as it doesn't actually use these csv files but instead generates them itself. The TRSM benchmarks will need to be added to it but that is a separate issue and shouldn't prevent this from being merged.

Thanks for checking!

ProGTX (approved):

PR #274, opened 2021-02-05
Refactor the utils library into a common folder that can be used by both tests and benchmarks

mcleary:
- Removes the sycl_blas dependency from clblas benchmarks (reduces compilation time of clblas benchmarks)
- Make a common library of utilities that both tests and benchmarks can use


benchmark/clblast/blas1/asum.cpp
```diff
@@ -42,7 +42,7 @@ void run(benchmark::State& state, ExecutorType* executorPtr, index_t size,
   state.counters["n_fl_ops"] = 2 * size_d;
   state.counters["bytes_processed"] = size_d * sizeof(scalar_t);
 
-  using data_t = utils::data_storage_t<scalar_t>;
+  using data_t = scalar_t;
```
ProGTX, line 45: Why this change? This was originally done to support multiple data types, like half.

benchmark/clblast/main.cpp
```diff
@@ -1,5 +1,5 @@
 #include "utils.hpp"
-#include "utils/extract_vendor_type.hpp"
+#include <common/extract_vendor_type.hpp>"
```
ProGTX, line 2:
```suggestion
#include "common/extract_vendor_type.hpp"
```

Looks like there's a typo there, but also it seems to me like double quotes would be preferable for including this header, also elsewhere.

benchmark/clblast/blas1/asum.cpp
mcleary: I don't think clblast/clblas have support for that. If this is required more refactoring will be needed. I will double check if you can use half in clblast

benchmark/clblast/main.cpp
mcleary: So, I transformed `common` into a library with it's own target. The quotes would mean this is a local file no?

mcleary: See d786661f8318649ab60b2b32fe897fdad333edb7 for the fix but I say we stick with `<>` for includes outside of target.

benchmark/clblast/blas1/asum.cpp
mcleary: So, clBlast doesn't have TRSM with half support...

common/include/common/common_utils.hpp
```diff
@@ -17,13 +17,15 @@
 #include <vector>
 
 #include "benchmark_cli_args.hpp"
-#include "blas_meta.h"
+//#include "blas_meta.h"
+//
+//#include "utils/quantization.hpp"
```
CodeplayBen, line 22: Are these still needed?

CodeplayBen (approved):
Looks good to me!

common/include/common/common_utils.hpp
mcleary: @CodeplayBen removed in 8d305ac7f01b42e357cc20acb5d766157726b1f4

mcleary:
@ProGTX @CodeplayBen I had to move the `concatenate_vectors` functions to the common library because gcc was complaining (with reason) that the function was not available in the benchmarks and tests so I'm not sure how MSVC was able to build everything but now it builds on both. See 8d305ac7f01b42e357cc20acb5d766157726b1f4 for the changes

benchmark/clblast/main.cpp
ProGTX: Makes sense to me, if it's meant to be a utility

benchmark/clblast/blas1/asum.cpp
ProGTX: BLAS implementations generally only support `float` and `double`, this is only to compare SYCL-BLAS to a reference implementation. Maybe then we don't need to enable it in this benchmark, but this particular change doesn't seem necessary, the original line should still work, right?

mcleary: In this case I had to change this because the `data_storage_t` is part of the internal utils of sycblas, located in `include/utils/quantization.hpp`, since I'm not linking the target `sycl_blas` into the clblast benchmarks, `utils::data_storage_t` will not be available here. Note that this will still be available in syclblas benchmarks.

ProGTX: I see. I think we'll need to revisit this at a later point, though it's possible we don't have to do anything else.

ProGTX (approved):

Bensuo (approved):

PR #276, opened 2021-02-10
Add documentation on how to add a new Level 3 operation in SYCL-BLAS

mcleary:
Update the README.md with to fix typos and add a refence to the new document


doc/AddingBlas3Op.md
```diff
@@ -0,0 +1,324 @@
...
+```
+
+## Instantiate the new operation
+
+The next step is to define the `_trsm` function in a `.cpp` file. In this way, when the client
+binary is linked against SYCL-BLAS, the linker will find the definition of the missing symbol.
+
+To do this, we create the source file that will contain instantiation of the new `_trsm` operation.
+The file is located at `src/interface/blas3/trsm.cpp.in`. This is not the file that will be 
+compiled, but a template file that the python script `python_generator/py_gen_blas_ternary.py`
```
CodeplayBen, line 175:
```suggestion
compiled, but a template file that the python script `python_generator/py_gen_blas_binary.py`
```

This should be binary if we are talking about TRSM right?

doc/AddingBlas3Op.md
```diff
@@ -0,0 +1,324 @@
...
+to instantiate the kernel. With this new interface defined, we can implement it internally by creating
+a new file called `src/operations/blas3/trsm.hpp`. This file will contain the implementation
+of `DiagonalBlocksInverter` and `make_diag_blocks_inverter`. The actual kernel is implemented
+in `DiagonalBlocksInverter::eval` in this file.
+
+The last part is to add this file in the include list in `src/operations/blas3/blas3_trees.hpp`.
+
+## Adding a call to the system blas version of the new operation
+
+SYCL-BLAS implements the netlib-blas interface, so there will be system blas version of the operation
```
CodeplayBen, line 292:
```suggestion
SYCL-BLAS implements the netlib-blas interface, so there will be a system blas version of the operation
```

CodeplayBen (approved):
Looks good to me just a couple of minor suggestions!

doc/AddingBlas3Op.md
```diff
@@ -0,0 +1,324 @@
...
+operation to SYCL-BLAS. It is mostly based on the work done when adding the
+TRSM algorithm to SYCL-BLAS.
+
+The steps described here are just a guideline and may vary when doing the
+actual implementation.
+
+## Add the operation interface
+
+The folder `include/interface` contains the public interface of SYCL-BLAS.
+The files in this folder contains the functions that users can call to
```
ProGTX, line 13:
```suggestion
The files in this folder contain the functions that users can call to 
```

doc/AddingBlas3Op.md
```diff
@@ -0,0 +1,324 @@
...
+
+The folder `include/interface` contains the public interface of SYCL-BLAS.
+The files in this folder contains the functions that users can call to 
+run the available blas operations.
+
+This folder has one file for each level of blas. For example, the `trsm`
+function will be located in `include/interface/blas3_interface.h`.
+
+When defining a new level 3 operation, the first step is to declare the
+user-facing function in this file, similar to the following:
```
ProGTX, line 20: `internal::_trsm` is not quite user-facing, the `internal` namespace is not meant for users

doc/AddingBlas3Op.md
```diff
@@ -0,0 +1,324 @@
...
+  //
+
+  // Create the executor for the test
+  auto q = make_queue();
+  test_executor_t ex(q);
+
+  // Invoke the newly added operation
+  _trsm(ex, side, triangle, transA, diag, m, n, alpha, a_gpu, lda, b_gpu, ldb);
+
+  // Very the results
```
ProGTX, line 103:
```suggestion
  // Verify the results
```

doc/AddingBlas3Op.md
```diff
@@ -0,0 +1,324 @@
...
+                                      ::testing::Values(0.0, NaN)   // unused
+);
+
+// Finaly register the new test
+BLAS_REGISTER_TEST(Trsm, combination_t, combi);
+```
+
+By this point, you are expected to get a linker error saying the symbol for the internal
+version of `blas::internal::_trsm` is undefined, and that is expected since we didn't 
+provide the internal implementation the newly added `_trsm` function.
```
ProGTX, line 125:
```suggestion
provide the internal implementation for the newly added `_trsm` function.
```

doc/AddingBlas3Op.md
```diff
@@ -0,0 +1,324 @@
...
+} // namespace internal
+} // namespace blas
+```
+
+## Instantiate the new operation
+
+The next step is to define the `_trsm` function in a `.cpp` file. In this way, when the client
+binary is linked against SYCL-BLAS, the linker will find the definition of the missing symbol.
+
+To do this, we create the source file that will contain instantiation of the new `_trsm` operation.
```
ProGTX, line 173:
```suggestion
To do this, we create the source file that will contain instantiations of the new `_trsm` operation.
```

doc/AddingBlas3Op.md
```diff
@@ -0,0 +1,324 @@
...
+
+} // namespace internal
+} // namespace blas
+```
+
+Where `${EXECUTOR}, ${INDEX_TYPE}, ${DATA_TYPE}, ${container_t0}` and `${container_t1}` are going
+to be replaced by the appropriate types required to explicitly instantiate the new function.
+
+Finally, the file `src/interface/blas3/CMakeLists.txt` must be changed in order for the source code
+where the instantiation of `_trsm` is to be generated. The following entry must be added:
```
ProGTX, line 211:
```suggestion
Finally, the file `src/interface/blas3/CMakeLists.txt` must be changed
in order to generate instantiations of `_trsm`.
The following entry must be added:
```

doc/AddingBlas3Op.md
```diff
@@ -0,0 +1,324 @@
...
+            )
+endfunction(build_library)
+```
+
+After this step the test defined for the new function will compile and link successfully and 
+`blas::internal::_trsm` will be called from the test.
+
+## Adding a kernel
+
+So far we have looked on how to add a new public facing function and how to instantiate it internally so that
```
ProGTX, line 243:
```suggestion
So far we have looked at how to add a new public facing function and how to instantiate it internally so that
```

doc/AddingBlas3Op.md
```diff
@@ -0,0 +1,324 @@
...
+of `DiagonalBlocksInverter` and `make_diag_blocks_inverter`. The actual kernel is implemented
+in `DiagonalBlocksInverter::eval` in this file.
+
+The last part is to add this file in the include list in `src/operations/blas3/blas3_trees.hpp`.
+
+## Adding a call to the system blas version of the new operation
+
+SYCL-BLAS implements the netlib-blas interface, so there will be system blas version of the operation
+being added. In the case of `_trsm`, the [netlib-blas version of the function](http://www.netlib.org/lapack/explore-html/d2/d8b/strsm_8f.html) can be called to verify that our implementation produces the correct results for a certain combination of
+parameters.
```
ProGTX, line 294:
```suggestion
being added. In the case of `_trsm`, the
[netlib-blas version of the function](http://www.netlib.org/lapack/explore-html/d2/d8b/strsm_8f.html)
can be called to verify that our implementation produces the correct results for a certain combination of parameters.
```

ProGTX (approved):

doc/AddingBlas3Op.md
mcleary: Yes, thanks!

doc/AddingBlas3Op.md
mcleary: But you still need to declare the function in the `internal` namespace located `blas3_interface.h`, so in my view this is still visible to users in a way

Bensuo (approved):

doc/AddingBlas3Op.md
ProGTX: I think it needs to be obvious that there's also a `blas::_trsm` function - that would be the user facing one.

mcleary: Text improved in e3ba0b6a1aab686575ada22fe396d3384712cebf

PR #279, opened 2021-02-16
Adds developer documentation for Gemm

Bensuo:
Adds a document detailing how to work with Gemm from a developer perspective.


doc/Gemm.md
```diff
@@ -0,0 +1,405 @@
...
+- `GEMM_TALL_SKINNY_SUPPORT` (Default: `ON`) - Enables optimizations for tall, skinny matrices. Not used on all targets.
+- `BLAS_MODEL_OPTIMIZATION` - Passing a machine learning model name here (`VGG_16` or `RESNET_50`) enables optimizations for the `Gemm` sizes used in these models. Only applies to the `ARM_GPU` target.
+
+## Kernel Structure
+
+Kernels are created as partial template specializations of the `Gemm` class to optimize for more specific cases or use different features (such as using local memory). The `Gemm` class has a number of template parameters, some of which are typically used for partial specialization. The class definition (along with the naive, reference implementation) is located in `gemm_ref.hpp`.
+
+There are some small member functions which do things such as determine the number of work groups required to execute each `Gemm`. However, the actual work of each kernel is done in `Gemm::eval()`, though the approach each kernel takes within this function is slightly different.
```
mcleary, line 78:
```suggestion
There are some small member functions which do things such as determine the number of work groups required to execute each `Gemm`. However, the actual work of each kernel is done in `Gemm::eval()`, though this approach each kernel takes within this function is slightly different. 
```

I could be wrong here, but I think that's what you meant

doc/Gemm.md
```diff
@@ -0,0 +1,405 @@
...
+
+# GEMM Configurations
+
+As previously touched on, tailored configurations for `GEMM` are provided on a per-target basis (along with the default CPU target configurations). Typically these are based on things like input size to provide optimal configurations for different use cases.
+
+## Backend Configurations
+
+Each backend header calls `Gemm_Launcher` with various configurations of template parameters to select different `GEMM` kernels and achieve the best performance within those kernels. The relevant parameters are:
+
+- Tile size by passing a `Tile<>` (found in `include/operations/blas3_trees.h`) Has parameters for batch sizes and for rows and columns of tile sizes at several levels:
```
mcleary, line 261:
```suggestion
- Tile size by passing a `Tile<>` (found in `include/operations/blas3_trees.h`) has parameters for batch sizes and for rows and columns of tile sizes at several levels:
```

doc/Gemm.md
```diff
@@ -0,0 +1,405 @@
...
+## CMake Configurations
+
+The generation of the `Gemm`, `Gemm_Launcher` and other operation's instantiations are driven through CMake and make use of the template files and python scripts previously touched on in [the section on source code generation](#source-code-generation).
+
+The configurations to be generated, along with associated functions, are located in `cmake/CmakeFunctionHelper.cmake` and these functions are called from `src/interface/<blas_level>/CMakeLists.txt`. Configurations are provided per backend target and will be generated for each data type set during CMake configuration with the variable `BLAS_DATA_TYPES`.
+
+As an example let's look at the configurations in `CmakeFunctionHelper.cmake` for the `RCAR` target backend, inside the function `generate_blas_gemm_objects`:
+
+```cmake
+if(${TARGET} STREQUAL "RCAR") # need investigation
```
mcleary, line 367: should we keep the `#need investigation` comment?

mcleary (commented):
Looks good to me, just a few comments. 

Also, I see you are using `GEMM` and `Gemm`, shouldn't we use just `GEMM`?

doc/Gemm.md
Bensuo: No I shall remove it, good catch!

doc/Gemm.md
Bensuo: I don't quite understand the suggested change, but I think the second half of that sentence is probably unclear and a little pointless so I'll change it.

Bensuo:
> Also, I see you are using `GEMM` and `Gemm`, shouldn't we use just `GEMM`?

I've tried to use `Gemm` when referring to the actual class and `GEMM` everywhere else but it's possible I've not been consistent enough with it, I'll give it another pass. Or would you prefer `GEMM` for all usages?

mcleary:
> > Also, I see you are using `GEMM` and `Gemm`, shouldn't we use just `GEMM`?
> 
> I've tried to use `Gemm` when referring to the actual class and `GEMM` everywhere else but it's possible I've not been consistent enough with it, I'll give it another pass. Or would you prefer `GEMM` for all usages?

Ah, I see now it makes sense. I think it is a good idea to keep it like this then

doc/Gemm.md
```diff
@@ -0,0 +1,405 @@
+# SYCL-BLAS GEMM Developer Documentation
+
+The following is documentation for the `GEMM` kernels and associated areas of code within `SYCL-BLAS` . This should give you a good understanding of how everything works and where/how to do things such as:
```
ProGTX, line 3:
```suggestion
The following is documentation for the `GEMM` kernels and associated areas of code within `SYCL-BLAS`.
This should give you a good understanding of how everything works and where/how to do things such as:
```

I suggest splitting up lines in the whole document. It doesn't have to fit 80 columns, but it should be something like one sentence per line or even more split.

doc/Gemm.md
```diff
@@ -0,0 +1,405 @@
...
+- `gemm_interleaved.hpp` - An alternative approach to batched `GEMM` calculations where the inputs are interleaved in contiguous memory. Uses no local memory and corresponds to HWN data layout (NWH in column major, which is what `SYCL-BLAS` uses).
+
+## Relevant CMake Variables
+
+There are several CMake variables which are specific to `GEMM` :
+
+- `NAIVE_GEMM` (Default: `OFF`) - Forces the use of the naive, reference `GEMM` implementation.
+- `GEMM_VECTORIZATION_SUPPORT` (Default: `OFF`) - Enables vectorization within the `GEMM` kernels. If `OFF` it is equivalent to passing `1` for the vector size to the `Gemm` launcher.
+- `GEMM_TALL_SKINNY_SUPPORT` (Default: `ON`) - Enables optimizations for tall, skinny matrices. Not used on all targets.
+- `BLAS_MODEL_OPTIMIZATION` - Passing a machine learning model name here (`VGG_16` or `RESNET_50`) enables optimizations for the `Gemm` sizes used in these models. Only applies to the `ARM_GPU` target.
```
ProGTX, line 72:
```suggestion
- `BLAS_MODEL_OPTIMIZATION` - Passing a machine learning model name here (`VGG_16` or `RESNET_50`) enables optimizations for the `GEMM` sizes used in these models. Only applies to the `ARM_GPU` target.
```

I think you use the all caps `GEMM` more often than the lowecase one when refering to the operation.

doc/Gemm.md
```diff
@@ -0,0 +1,405 @@
...
+
+## Kernel Structure
+
+Kernels are created as partial template specializations of the `Gemm` class to optimize for more specific cases or use different features (such as using local memory). The `Gemm` class has a number of template parameters, some of which are typically used for partial specialization. The class definition (along with the naive, reference implementation) is located in `gemm_ref.hpp`.
+
+There are some small member functions which do things such as determine the number of work groups required to execute each `Gemm`. However, the actual work of each kernel is done in `Gemm::eval()`, though the approach each kernel takes within this function is slightly different. 
+
+The general goals for programming a `GEMM` kernel can be summarised as follows:
```
ProGTX, line 80:
```suggestion
The general goals for programming a `GEMM` kernel can be summarized as follows:
```

doc/Gemm.md
```diff
@@ -0,0 +1,405 @@
...
+
+- Balance register pressure with loop unrolling to achieve an optimal balance of performance.
+- Using SFINAE and templates where applicable to minimise branching and keep as many things const and compile-time as possible for increased performance.
```
ProGTX, line 83:
```suggestion
- Using SFINAE and templates where applicable to minimize branching and keep as many things const and compile-time as possible for increased performance.
```

doc/Gemm.md
```diff
@@ -0,0 +1,405 @@
...
+
+Outside of the naive, reference `Gemm` all the kernels are tile based, where each instance of the kernel calculates a small portion of the overall matrix. Much of the work in the `::eval()` functions tends to be precalculating indices, offsets and other values for use in the actual computation. Because they are tile based one of the main considerations is whether the current tile is internal or external. If it's internal that means that boundary checks can be avoided which is a significant time save and performance increase.
+
+The core of the `Gemm` computation is as follows:
+
+- Loop over the K dimension.
```
ProGTX, line 89: This list should probably be enumerated if it represents steps

doc/Gemm.md
```diff
@@ -0,0 +1,405 @@
...
+
+/*!
+ * @brief Wrapper around Gemm. Creates the views, then makes and launches Gemm
+ */
+template <int WgSize, bool DoubleBuffer, bool ConflictA, bool ConflictB, 
+          int ClSize, typename TileT, bool TransA, bool TransB,
+          int GemmMemoryType, int GemmAlgorithm, int GemmVectorization,
+          bool is_beta_zero, int VectorSize, int BatchType>
+
+template <typename Executor, typename container_t0, typename container_t1,
```
ProGTX, line 186: Is this second `template` required?

doc/Gemm.md
```diff
@@ -0,0 +1,405 @@
...
+      static_cast<int>(
+          gemm_batch_type_t::strided)>::template _select_gemm(ex, _M, _N, _K,
+                                                              _alpha, _a, _lda,
+                                                              _b, _ldb, _beta,
+                                                              _c, _ldc,
+                                                              batch_size);
+#else
+```
+
+Next we have an `#if` directive for if we want to force naive `gemm` configurations. This is triggered by a cmake variable. You can see other examples like this in `arm_gpu.hpp` which does similar things for different values of the cmake variable `BLAS_MODEL_OPTIMIZATION` .
```
ProGTX, line 324:
```suggestion
Next we have an `#if` directive for when we want to force naive `gemm` configurations. This is triggered by a cmake variable. You can see other examples like this in `arm_gpu.hpp` which does similar things for different values of the cmake variable `BLAS_MODEL_OPTIMIZATION` .
```

ProGTX (changes requested):
Looks very good, thanks.

doc/Gemm.md
Bensuo: I believe so yes, at least according to https://en.cppreference.com/w/cpp/language/member_template

doc/Gemm.md
Bensuo: I've removed this part of the sentence!

doc/Gemm.md
Bensuo: I've done a pass on this and it should be more consistent now, `Gemm` should only be used when explicitly referring to the class and `GEMM` in all other cases.

doc/Gemm.md
Bensuo: Fixed, thanks!

ProGTX (approved):

mehdi-goli (approved):
LGTM!

mehdi-goli (approved):

PR #283, opened 2021-02-24
Fix ACL benchmark compilation failure

Bensuo:
Fix ACL benchmark compilation failing after the refactor of the utilities into a common library. Links this target in CMake and modifies the include slightly.


mcleary (approved):
Good catch

mehdi-goli (approved):

PR #286, opened 2021-03-05
Add Dependabot to track Actions dependencies

Bensuo:
Adds a Dependabot configuration file to enable dependency tracking for Github Actions workflows. Checks daily for updates to actions used in the workflows.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: daily


.github/dependabot.yml
```diff
@@ -0,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: daily
```
ProGTX, line 6: What does this actually do, though? Do we need to define dependencies somewhere? Or what even are dependencies in this case?

ProGTX (approved):

.github/dependabot.yml
Bensuo: Ah I probably should've added an explanation, it checks any external actions used in the workflows, which if you look in the workflow `.yml` file look like: `- uses: actions/checkout@v2` and exist as Github repos, for example https://github.com/actions/checkout.

It checks if there are newer versions available. So in that example if there was a `v3` release of the checkout action available Dependabot would create a PR here automatically to bump the version inside the workflow file and keep it up to date.

mehdi-goli (approved):

PR #289, opened 2021-03-23
Fix actions case sensitivity

Bensuo:
Fixes an issue where PRs from branches which had upper case letters would fail to clone inside the docker image when running the CI. Switches the environment variable in question to a case sensitive one.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -60,7 +60,7 @@ jobs:
       if: github.event_name == 'pull_request'
       run: |
         echo "GIT_SLUG=${{github.event.pull_request.head.repo.full_name}}" >> $GITHUB_ENV
-        echo "GIT_BRANCH=${{ env.GITHUB_HEAD_REF_SLUG }}" >> $GITHUB_ENV
+        echo "GIT_BRANCH=${{ env.GITHUB_HEAD_REF_SLUG_CS }}" >> $GITHUB_ENV
     - name: Build the Docker image
       env:
         C_COMPILER: ${{matrix.cc_compiler}}
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -50,7 +50,6 @@ jobs:
     
     steps:
     - uses: actions/checkout@v2
-    - uses: rlespinasse/github-slug-action@v3.x
     - name: Push Event Environment Vars
       if: github.event_name == 'push'
       run: |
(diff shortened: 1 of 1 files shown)


mehdi-goli (approved):

mehdi-goli:
can you rename this branch to have one capital letter and `_` on it? to make sure it is working :D

.github/workflows/build-and-test.yml
```diff
@@ -60,7 +60,7 @@ jobs:
       if: github.event_name == 'pull_request'
       run: |
         echo "GIT_SLUG=${{github.event.pull_request.head.repo.full_name}}" >> $GITHUB_ENV
-        echo "GIT_BRANCH=${{ env.GITHUB_HEAD_REF_SLUG }}" >> $GITHUB_ENV
+        echo "GIT_BRANCH=${{ env.GITHUB_HEAD_REF_SLUG_CS }}" >> $GITHUB_ENV
```
ProGTX, line 63: I find it strange that the default (without `_CS`) is basically wrong

ProGTX (approved):

Bensuo:
> can you rename this branch to have one capital letter and `_` on it? to make sure it is working :D

Ah sadly I can't do that, since until this is merged the change won't be reflected in the jobs run on this PR (which comes from the master branch). However, I tested this on my fork with a pull request and it worked just fine!

If you want you can look at one of the runs that failed on your PR https://github.com/codeplaysoftware/sycl-blas/runs/2166173533 and check the `PR Environment Vars` step of the job. If you expand the terminal output with the small arrow you can see all the environment vars including those with the `_CS` suffix which is what I've switched to here, those maintain the case instead of converting everything to lower case.

.github/workflows/build-and-test.yml
mehdi-goli: The previous branch I have created had `_` on it as well and the slug was converting `_` to `-`. Does this flag fix this issue as well?

Bensuo: Ah I missed that it had an underscore, will fix that also.

Bensuo: OK I have removed the slug action altogether since I found the built in Github equivalent. I've tested it with both upper case and underscores in a PR on my fork [here](https://github.com/Bensuo/sycl-blas/actions/runs/679855280) that's running now, if you check the PR environment vars step again you can see the correct branch name is passed.

PR #291, opened 2021-04-15
Added flags for header only tests and fpga copy kernel

ShanoToni:
Cmake files have been modified to include:
* BLAS_HEADER_ONLY_TESTING - allows test to be compiled with sycl-blas in header only mode ( to fix "programmed with different xclbin" error on fpga)
* SYCL_BLAS_FPGA - allows use of a ```copy_to_device()``` function which uses a kernel to copy data to the device (to handle the "xclbin not loaded error" on the fpga)


mehdi-goli (approved):
LGTM! We need to find a better way than copy kernel to bypass the `xclbin` error

PR #293, opened 2021-05-01
Only reference cl::sycl::half if BLAS_DATA_TYPE_HALF is set.

fodinabor:
During some experimentation trying to build sycl-blas with hipSYCL, which doesn't support half types, yet, I discovered that the usage of `cl::sycl::half` is not always correctly ifdef'd away.


mehdi-goli (approved):
Thanks! LGTM!

Bensuo (approved):

PR #294, opened 2021-05-01
Use sycl::vec default initialization instead of using 0 as argument.

fodinabor:
This is required as the spec expects the same data type for the argument as the vector has, otherwise the variadic ctor will be chosen, which has different semantics. So either `static_cast`s would've to be added or we just can rely on the default ctor to zero initialize the contents.


mehdi-goli (approved):
LGTM!

Bensuo (approved):

PR #295, opened 2021-05-01
Use SYCL 2020 style selectors.

fodinabor:
The main change is that it's no longer assumed that the selectors derive from `sycl::device_selector` but instead they are callables.

Sorry for the spam, just wanted to have the fixes to be considered one-by-one if something would break otherwise :)

Also just for reference, I adapted the CMake configuration files quite rudimentary to build with hipSYCL in this branch, not sure what the status of interest would be to get (a polished version of this) merged: https://github.com/fodinabor/sycl-blas/tree/feature/hipsycl-build

### DIFF
--- a/test/blas_test.hpp
+++ b/test/blas_test.hpp
@@ -64,14 +64,7 @@ using test_executor_t =
  * using the default device if not specified.
  */
 inline cl::sycl::queue make_queue_impl() {
-  std::function<int(const cl::sycl::device&)> selector;
-  if (args.device.empty()) {
-    selector = cl::sycl::default_selector();
-  } else {
-    selector = utils::cli_device_selector(args.device);
-  }
-
-  auto q = cl::sycl::queue(selector, [=](cl::sycl::exception_list eL) {
+  auto async_handler = [=](cl::sycl::exception_list eL) {
     for (auto &e : eL) {
       try {
         std::rethrow_exception(e);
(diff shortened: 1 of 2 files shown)


ProGTX (changes requested):
Hi, thanks for the PR.

I think we should at least wrap SYCL 2020 features with `#ifdef SYCL_LANGUAGE_VERSION >= 202002`. But even then this might be a problem since an implementation might not support it yet - ComputeCpp 2.5 was just released and it doesn't have this feature (even though it's high on our list). So for now I'd like to ask if we could cover both of those limitations - hopefully we'll be able to remove the ComputeCpp limitation soon, and at some point later we might even shift to SYCL 2020 properly.

Thinking out loud, maybe some of these features could be opt-in CMake options, that's probably the easiest way to check for both SYCL 2020 and implementation support.

Bensuo:
Thanks for the PR!

> Also just for reference, I adapted the CMake configuration files quite rudimentary to build with hipSYCL in this branch, not sure what the status of interest would be to get (a polished version of this) merged

Having support for hipSYCL in the CMake would be great! I think the approach you've taken already makes sense for the most part. For setting the specific GPU architecture (`sm` version etc.) I would probably favour having some sensible default which the user can override by setting an environment variable, rather than adding a new CMake option, since this is a hipSYCL detail. Can hipSYCL accept an environment variable directly for this or does it need to be passed as a command line argument?

fodinabor:
Thanks @ProGTX for the suggestion.
Where would be the right place to add those options? I guess it'd make sense to provide defaults for the implementations which could be updated as the implementations adopt SYCL 2020 features. Thinking of doing this in ConfigureSYCLBLAS.cmake. Is that sensible?

test/blas_test.hpp
```diff
@@ -85,7 +76,27 @@ inline cl::sycl::queue make_queue_impl() {
...
+  std::unique_ptr<cl::sycl::device_selector> selector;
+  if (args.device.empty()) {
+    selector = std::unique_ptr<cl::sycl::device_selector>(
+        new cl::sycl::default_selector());
+  } else {
+    selector = std::unique_ptr<cl::sycl::device_selector>(
+        new utils::cli_device_selector(args.device));
+  }
+  auto q = cl::sycl::queue(*selector, async_handler);
+#endif
```
ProGTX, line 99:
```suggestion
#endif  // HAS_SYCL2020_SELECTORS
```

ProGTX (approved):
Sorry for the late response, the switch makes sense, thank you.

mehdi-goli:
Thank you @fodinabor. Could you put SYCL related flags in [SYCL.cmake](https://github.com/codeplaysoftware/sycl-blas/blob/master/cmake/Modules/SYCL.cmake) file. You can put any HIPSYCL related enablement flags there (Similar to computeCpp and DPC++). Also, would it be possible to use the flag suggested by @ProGTX  to check if it is SYCL-2020 or not. It would be preferable to categorize all SYCL-2020 features under the same SYCL-2020 flags rather than per feature flag.

ProGTX:
We discussed this a bit internally, and it's probably cleaner to just check for the `SYCL_LANGUAGE_VERSION` macro instead of adding new CMake options - there could potentially be a lot of CMake features. It's unfortunate that ComputeCpp 2.5 won't be able to work with this just yet, but that was bound to happen as people start using more SYCL 2020 features. It will still work with the 1.2.1 version of SYCL-BLAS, though.

mehdi-goli (approved):

PR #298, opened 2021-05-14
Anton/experimental fpga

ShanoToni:
Added modifications to the tests and benchmarks in order to use the ComputeCPP runtime fpga fix in the queue constructor.
When compiling with the SYCL_BLAS_FPGA flag the queue is constructed in the .hpp file included in the benchmark or test .cpp file so it passes in trough the compute++ compiler and has access to the new dummy kernel from the queue constructor. (Required for the FPGA)


CMakeLists.txt
```diff
@@ -124,6 +126,8 @@ if (WIN32)
     target_link_libraries(sycl_blas PRIVATE ComputeCpp::ComputeCpp)
   elseif(is_dpcpp)
     target_link_libraries(sycl_blas PRIVATE DPCPP::DPCPP)
+  elseif(is_hipsycl)
+    target_link_libraries(sycl_blas PRIVATE hipSYCL::hipSYCL-rt)
```
ProGTX, line 130: I think it'd be worth splitting out hipSYCL support and most other CMake changes (without FPGA support) into a separate PR, I believe we should be able to get that in much faster.

ProGTX (changes requested):

mehdi-goli (approved):

mehdi-goli (changes requested):
The change in the Benchmark and test are radical, we need to find a better way to remove the queue creation

Rbiessy:
Closing as this is outdated.

PR #315, opened 2021-11-11
Improve testing of Reduction

sgeor255:
- Enable testing of Product Reduction. The tests were failing since the input
range was a bit too large. For large input matrices (e.g. 513x8195) it was
outputting too large/small values to fit in a float. As a workaround, the input
range was made smaller for large matrices. 

- Remove testing of Division Reduction. The reduction algorithm assumes that the
binary operator used is commutative, which Division is not. Therefore, the
Division operator cannot be supported and was removed.

### DIFF
--- a/test/blas_test.hpp
+++ b/test/blas_test.hpp
@@ -125,19 +125,32 @@ static inline scalar_t random_scalar(scalar_t rangeMin, scalar_t rangeMax) {
 }
 
 /**
- * @fn random_data
+ * @brief Generates a random vector of scalar values, using a uniform
+ * distribution.
+ * @param vec Input vector to fill
+ * @param rangeMin Minimum value for the uniform distribution
+ * @param rangeMax Maximum value for the uniform distribution
+ */
+template <typename scalar_t>
+static inline void fill_random_with_range(std::vector<scalar_t> &vec,
+                                          scalar_t rangeMin,
+                                          scalar_t rangeMax) {
+  for (scalar_t &e : vec) {
+    e = random_scalar(rangeMin, rangeMax);
+  }
+}
+
+/**
  * @brief Generates a random vector of scalar values, using a uniform
  * distribution.
  */
 template <typename scalar_t>
 static inline void fill_random(std::vector<scalar_t> &vec) {
-  for (scalar_t &e : vec) {
-    e = random_scalar(scalar_t{-2}, scalar_t{5});
-  }
+  fill_random_with_range(vec, scalar_t{-2}, scalar_t{5});
 }
 
 /**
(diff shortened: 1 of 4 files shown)


test/blas_test.hpp
```diff
@@ -125,19 +125,33 @@ static inline scalar_t random_scalar(scalar_t rangeMin, scalar_t rangeMax) {
...
+                                          scalar_t rangeMax) {
   for (scalar_t &e : vec) {
-    e = random_scalar(scalar_t{-2}, scalar_t{5});
+    e = random_scalar(rangeMin, rangeMax);
   }
 }
 
 /**
- * @breif Fills a lower or upper triangular matrix suitable for TRSM testing
+ * @fn random_data
```
Bensuo, line 144: This should probably be removed as the name doesn't match, and I believe `fn` should only be used when the comment block does not directly proceed the function it is supposed to document ([see here](https://www.doxygen.nl/manual/commands.html#cmdfn))

Bensuo (changes requested):
LGTM just a minor comment about comments.

test/blas_test.hpp
sgeor255: Thanks @Bensuo , I removed it.

test/unittest/reduction/reduction_test.cpp
```diff
@@ -73,7 +70,13 @@ void run_test(const combination_t<scalar_t> combi) {
   std::vector<data_t> out_v_gpu(out_size);
   std::vector<data_t> out_v_cpu(out_size);
 
-  fill_random(in_m);
+  if (op == operator_t::Product && (rows == 513 || cols == 8195)) {
```
mehdi-goli, line 73: Is it particularly for this size or larger size? The former can be a bug. If it is the latter, the issue is related to a limited size that a floating-point operation can accommodate. So the ad-hoc if is not the solution. You can either reduce the size of the test when the type is float in all cases or for the entire reduction test control the random value via fill_random_with_Range as depending on the operation this issue can happen in all cases. So the test should be defined carefully to consider floating-point limitation when the type is float.

test/blas_test.hpp
sgeor255: Thanks @Bensuo , removed it.

test/unittest/reduction/reduction_test.cpp
sgeor255: Thanks @mehdi-goli . The tests fail for larger sizes too, so I the issue appears to be overflow. I changed the range for all product tests to workaround this issue and added a comment.

Bensuo (approved):

mehdi-goli (approved):

PR #321, opened 2022-02-08
Add test parameters to GTest names

Rbiessy:
The tests are renamed with the parameters value.
This allows for a finer granularity than using `STRESS_TESTING`
when excluding tests. The parameters name are of the form
`(__<paramName>_<paramValue>)*`.

The tests are now run with an extra flag to produce an xml summary when
running ctest.

Also fix undefined variable that led to tests printing the help message.


mehdi-goli:
Thanks, @Rbiessy  for the PR. I think a lot of operations for generating the name string can be wrapped in C++ variadic template. So instead of repeating the function you can use one variadic template function and use that for all tests. 

Maybe Something like: 
```c++
template <class... Ts>
void get_name(std::sstream& ss, Ts const&... args) {
    ((ss <<"__" << std::to_string(args) << "__" <<args), ...);
}
```

Rbiessy:
Thanks @mehdi-goli. This is a bit more complicated than that because I need to alternate between a string for the parameter name and the parameter value. I'll see if I can improve that with a variadic zip function.

DuncanMcBain (approved):

mehdi-goli (approved):

Issue #337, opened 2022-10-10
Incorrect library name for DPC++ sycl-runtime on Windows.


hjabird:
In [this](https://github.com/intel/llvm/commit/c8427d6d4c4b4cad22cb135a5367689a1724068d) commit to Intel's DPC++ nightly, the Windows SYCL-runtime library 's name was changed from `sycl.dll/lib` to `sycl{MAJOR_VERSION}.dll/lib`. This breaks the FindDPCPP.cmake module in SYCL-BLAS, which looks for just `sycl.dll/lib`.

PR #338, opened 2022-10-10
[DPC++] Look for renamed DPC++ runtime shared library.

hjabird:
This PR fixes https://github.com/codeplaysoftware/sycl-blas/issues/337

* The intel DPC++ nightly renames sycl.lib/dll to sycl{MAJOR_VER}.lib/dll.
* This breaks the FindDPCPP.cmake script.
* This commit allows sycl6.dll to be found.
  * Tested on Windows against:
	* SYCL-nightly 2022/10/26 (w/ sycl6.lib, passes).
	* SYCL-nightly 2022/09/11 (w/ sycl.lib, passes).
  * Tested on Linux against 2022/09/11, passes.

### DIFF
--- a/cmake/Modules/FindDPCPP.cmake
+++ b/cmake/Modules/FindDPCPP.cmake
@@ -38,7 +38,7 @@ set(DPCPP_USER_FLAGS "" CACHE STRING
     "Additional user-specified compiler flags for DPC++")
 
 get_filename_component(DPCPP_BIN_DIR ${CMAKE_CXX_COMPILER} DIRECTORY)
-find_library(DPCPP_LIB_DIR NAMES sycl PATHS "${DPCPP_BIN_DIR}/../lib")
+find_library(DPCPP_LIB_DIR NAMES sycl sycl6 PATHS "${DPCPP_BIN_DIR}/../lib")
 
 add_library(DPCPP::DPCPP INTERFACE IMPORTED)
 


Rbiessy (approved):

mehdi-goli (approved):

cmake/Modules/FindDPCPP.cmake
```diff
@@ -38,7 +38,7 @@ set(DPCPP_USER_FLAGS "" CACHE STRING
     "Additional user-specified compiler flags for DPC++")
 
 get_filename_component(DPCPP_BIN_DIR ${CMAKE_CXX_COMPILER} DIRECTORY)
-find_library(DPCPP_LIB_DIR NAMES sycl PATHS "${DPCPP_BIN_DIR}/../lib")
+find_library(DPCPP_LIB_DIR NAMES sycl sycl6 PATHS "${DPCPP_BIN_DIR}/../lib")
```
KumudhaN, line 41: May be we should have a comment here to say that this was the last sycl/dpcpp version for which windows support was tested on.

KumudhaN (approved):

cmake/Modules/FindDPCPP.cmake
mehdi-goli: It might worth considering backward compatibility as well, if and when needed

PR #339, opened 2022-10-12
Fix support for 64b index types

hjabird:
SYCL-BLAS requires minor modifications to compile using `int64_t` as an index type.  Additionally, the tests and benchmarks require minor modifications.

* Fixes issues for using `int64_t` as an index type.
* Updates benchmarks and tests to test using chosen index type.
  * Fix test name issue when using `cl::sycl::half` or `double`.
* Introduces `BLAS_INDEX_TYPES` cmake variable
* Documents `BLAS_INDEX_TYPES` and `BLAS_DATA_TYPES` in README.

### DIFF
--- a/common/include/common/common_utils.hpp
+++ b/common/include/common/common_utils.hpp
@@ -21,7 +21,12 @@
 #include <common/float_comparison.hpp>
 #include <common/system_reference_blas.hpp>
 
+#ifdef BLAS_BENCHMARK_INDEX_T
+using index_t = BLAS_BENCHMARK_INDEX_T;
+#else
 using index_t = int;
+#endif // BLAS_BENCHMARK_INDEX_T
+
 using blas1_param_t = index_t;
 
 template <typename scalar_t>
--- a/test/blas_test.hpp
+++ b/test/blas_test.hpp
@@ -60,6 +60,13 @@ using namespace blas;
 using test_executor_t =
     blas::Executor<blas::PolicyHandler<blas::codeplay_policy>>;
 
+    
+#ifdef BLAS_BENCHMARK_INDEX_T
+using index_t = BLAS_BENCHMARK_INDEX_T;
+#else
+using index_t = int;
+#endif // BLAS_BENCHMARK_INDEX_T
+
 /**
  * Construct a SYCL queue using the device specified in the command line, or
  * using the default device if not specified.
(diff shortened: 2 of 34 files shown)


Rbiessy (approved):

common/include/common/common_utils.hpp
```diff
@@ -21,7 +21,12 @@
 #include <common/float_comparison.hpp>
 #include <common/system_reference_blas.hpp>
 
+#ifdef BLAS_BENCHMARK_INDEX_T
```
mehdi-goli, line 24: You can always define this on the Cmake and avoid having it in the code. Set the default value for BLAS_BENCHMARK_INDEX_T to be int in CMAKE and directly set `using index_t=BLAS_BENCHMARK_INDEX_T`

hjabird: In [107e289](https://github.com/codeplaysoftware/sycl-blas/pull/339/commits/107e289ed09eef9c570e5cda6fd6fe9f909139cc):
* The default values for `BLAS_BENCHMARK_INDEX_T` have been removed from the .hpp files.
  * The CMake should always define something - there should be no need for a default in the CMake.
* For the unit tests, the `BLAS_BENCHMARK_INDEX_T` has been renamed to `BLAS_TEST_INDEX_T`.

mehdi-goli: Is there any reason that we need to use two different one? can we call it BLAS_INDEX_T and use it for all?

hjabird: Possibly. The intention was to avoid confusion with `BLAS_INDEX_TYPES`, but instead they may have confused things.
* `BLAS_INDEX_TYPES` can support multiple types eg. `-DBLAS_INDEX_TYPES="int32_t;int64_t"`.
* The unit tests and benchmarks are not written to support testing with multiple index types.
  * This would require far more substantial modifications than this PR.
  * Instead, the first type of the list of `BLAS_INDEX_TYPES` is used.

In [3c27a08](https://github.com/codeplaysoftware/sycl-blas/pull/339/commits/3c27a08309f7f507aac7f8a31b3836b0f3edd8d0) I've renamed them as `BLAS_INDEX_T` as suggested.

mehdi-goli: Yes, I did not meant to touch BLAS_INDEX_TYPES. I just wanted you to use BLAS_INDEX_T instead of BLAS_BENCHMARK_INDEX_T and BLAS_TEST_INDEX_T.

test/blas_test.hpp
```diff
@@ -226,6 +229,29 @@ inline void dump_arg<float>(std::ostream &ss, float f) {
   }
 }
 
+template <>
```
mehdi-goli, line 232: Is this change also related to the index type?

hjabird: No, these are related to `BLAS_DATA_TYPES`.

The test names are generated using the test parameters. However, a test's name can't contain `-` or `.`. This causes a problem for floating point test parameters. To circumvent this, the tests replace `-` for `m` and `.` for `p`. 

However, this fix has only been implemented for `float`. Consequently, before this PR, testing with `half` or `double` results in `-` or `.` in test names, causing the tests to fail. This changes ensures that a valid test names are generated when `BLAS_DATA_TYPES` includes `half` or `double`.

mehdi-goli: The code in the template body is exact copy of float. I think there is a lof of code duplication here. You can abstract the code in the main template or put it in a struct so you can partially instantiate it for float, double or half. Also it is better to be in a separate PR as it is not related to the 64 index type.

hjabird: This change is removed in [010ab05](https://github.com/codeplaysoftware/sycl-blas/pull/339/commits/010ab05c690200c552f03736b8dc2afe3d412f05), and I've filed a new issue https://github.com/codeplaysoftware/sycl-blas/issues/340.

mehdi-goli (approved):

PR #346, opened 2022-11-03
Add option to install SYCL-BLAS as a header only library

Rbiessy:
This moves the generated config and version files to the root of the install directory so that they can be detected with CMake's find_package [config mode](https://cmake.org/cmake/help/latest/command/find_package.html#id6).

I have confirmed that this does not break SYCL-DNN build. SYCL-DNN is currently using CMake's find_package "module mode" (i.e. using a FindSyclBlas.cmake file) so it does not conflict. With these changes FindSyclBlas.cmake should not be needed anymore.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -207,7 +207,8 @@ export(EXPORT sycl_blas
 
 option(BLAS_ENABLE_TESTING "Whether to enable testing" ON)
 option(ENABLE_EXPRESSION_TESTS "Whether to build expression tree fusion tests" OFF)
-if (INSTALL_HEADER_ONLY)
+if (INSTALL_HEADER_ONLY AND BLAS_ENABLE_TESTING)
+  message(STATUS "Tests are disabled when installing SYCL-BLAS in header only mode")
   set(BLAS_ENABLE_TESTING OFF)
 endif()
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -204,6 +223,10 @@ option(BUILD_CLBLAST_BENCHMARKS "Whether to build clBLAST benchmarks" OFF)
 option(BUILD_CLBLAS_BENCHMARKS "Whether to build clBLAS benchmarks" OFF)
 option(BUILD_ACL_BENCHMARKS "Whether to build ARM Compute Library benchmarks" OFF)
 option(BLAS_BUILD_SAMPLES "Whether to build SYCL-BLAS samples" ON)
+if (INSTALL_HEADER_ONLY)
+  set(BLAS_ENABLE_BENCHMARK OFF)
+  set(BLAS_BUILD_SAMPLES OFF)
+endif()
```
hjabird, line 229: I think it would be better to create an error here if these are on instead.

Rbiessy: I find it annoying to have to set 3 flags to off when you turn on  `INSTALL_HEADER_ONLY`. I've added a message instead, I hope that's enough.

KumudhaN (approved):

PR #353, opened 2022-11-21
Rotm

fabiomestre:
Adds support for the BLAS level 1 rotm api.
Adds a unit test and benchmark for rotg

### DIFF
--- a/src/interface/blas1_interface.hpp
+++ b/src/interface/blas1_interface.hpp
@@ -347,6 +347,92 @@ typename executor_t::policy_t::event_t _rot(
 }
 
 /**
+ * @brief Performs a modified Givens rotation of points.
+ * Given two vectors x and y and a modified Givens transformation matrix, each
+ * element of x and y is replaced as follows:
+ *
+ * [xi] = [h11 h12] * [xi]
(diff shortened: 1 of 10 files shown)


t4c1 (approved):
LGTM

KumudhaN (approved):

src/interface/blas1_interface.hpp
```diff
@@ -346,6 +346,92 @@ typename executor_t::policy_t::event_t _rot(
...
+  element_t h21 = param_host[2];
+  element_t h12 = param_host[3];
+  element_t h22 = param_host[4];
+
+  using m_two = constant<element_t, const_val::m_two>;
+  using m_one = constant<element_t, const_val::m_one>;
+  using zero = constant<element_t, const_val::zero>;
+  using one = constant<element_t, const_val::one>;
+
+  if (flag == zero::value()) {
```
Rbiessy, line 410: You could leave another comment that this can be further optimized if we fall under one of these 3 cases. Some computations below could be skipped. In the case of `flag == -2` it could even return the input without any computation.

Rbiessy: Nevermind I'm too late.

PR #354, opened 2022-11-23
Fix expression tree tests

sgeor255:


test/exprtest/blas1_axpy_copy_test.cpp
```diff
@@ -23,6 +23,7 @@
  *
  **************************************************************************/
 #include "blas_test.hpp"
+#include "blas_test_macros.hpp"
```
Rbiessy, line 26: The other tests don't include this file as it's already included by `blas_test.hpp`. Can we remove them?

Rbiessy (commented):
Thanks for the fix!

test/exprtest/blas1_axpy_copy_test.cpp
sgeor255: My bad, removed it.

Rbiessy (approved):

KumudhaN (approved):
LGTM. Thanks for fixing this!

PR #355, opened 2022-11-24
Add the correct target for the CPU backend

KumudhaN:
The correct target triple is available here - https://intel.github.io/llvm-docs/UsersManual.html


hjabird:
Why does this need changing? The original looks valid from the docs. And why not `spir64_x86_64`?

cmake/Modules/SYCL.cmake
```diff
@@ -90,7 +90,7 @@ elseif(is_dpcpp)
   set(CMAKE_CXX_STANDARD 17)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D__SYCL_DISABLE_NAMESPACE_INLINE__=ON -O3 -Xclang -cl-mad-enable")
   if(${BACKEND_DEVICE} STREQUAL "DEFAULT_CPU") 
-    set(DPCPP_SYCL_TARGET spir64_x86_64-unknown-unknown)
+    set(DPCPP_SYCL_TARGET spir64)
```
hjabird, line 93: Why does this need changing? The original looks valid from the docs. And why not `spir64_x86_64`?

KumudhaN: `spir64_x86_64-unknown-unknown` generates code AOT which makes the CI and any other compilations for developing and debugging very slow. Hence, made the default a non-AOT one and added a specific target for Intel CPU which does AOT. 

If people don't like this option, another one would be to accept target triple from the user and remove this elif ladder. Happy to change to that as well. Let me know your thoughts.

mehdi-goli: The proper way is to allow user to pass target-tripple. the `-DTARGET` name should be removed and target triple should allow choosing the backend.

hjabird: Is `DEFAULT_CPU` still a good name for this target?

pgorlani:
Hi everybody, here is the solution I implemented after talking with @KumudhaN and @mehdi-goli.

Basically, in the current `master`

1. `TARGET` effects the compilation in `../cmake/Modules/SYCL.cmake`
    * `TARGET -> BACKEND_DEVICE -> DPCPP_SYCL_TARGET`
    * `TARGET -> HIPSYCL_TARGETS`

2. `TARGET` effects the tuning in `../cmake/CmakeFunctionHelper.cmake` and `../tools/auto_tuner/CMakeLists.txt`

The solution of this patch implies

1. removing `TARGET` from **compilation**, allowing the user to set directly `DPCPP_SYCL_TARGET` and `HIPSYCL_TARGETS`. If not set, the actual default behaviors are preserved.

2. merging `TARGET` and `BACKEND_DEVICE`, renaming them in `TUNING_TARGET` for specifying the **tuning** configuration.

README.md
```diff
@@ -329,17 +329,23 @@ cd build
 cmake -GNinja ../ -DSYCL_COMPILER=dpcpp
 ninja
 ```
+The target triplet can be set by adding `-DDPCPP_SYCL_TARGET=<triplet>`. If it
+is not set, the default values is `spir64_x86_64-unknown-unknown`, which targets
```
KumudhaN, line 333: I think we should change the default target to - spir64-unknown-unknown to be consistent with what DPCPP specifies: 
`spir64 - this is the default generic SPIR-V target;`

README.md
```diff
@@ -329,17 +329,23 @@ cd build
...
+the CPU. Other possible triplets targeting GPUs are: `spir64-unknown-unknown`
+(Intel), `nvptx64-nvidia-cuda` (Nvidia), and `amdgcn-amd-amdhsa` (AMD).
 
-If building for `AMD_GPU` devices, it is also required to provide the
-specific architecture through the `DPCPP_SYCL_ARCH` CMake variable.
+If building for NVIDIA or AMD devices, it is advisable to specify the actual
```
KumudhaN, line 337: We can call it as advisable for nvidia but required for AMD.

KumudhaN (commented):
@mehdi-goli and @Rbiessy - Should we have a section in the README which says we no longer support `TARGET` - This might help anyone who used the library before with TARGET and puller a later version after this change.

Rbiessy (commented):
> @mehdi-goli and @Rbiessy - Should we have a section in the README which says we no longer support `TARGET` - This might help anyone who used the library before with TARGET and puller a later version after this change.

Sure that'd be nice. Also the README still mentions `TARGET` in a couple of places: for PowerVR support and in the cross-compile section.

README.md
pgorlani: I did it in the last commits.

README.md
pgorlani: I rephrased the sentence.

pgorlani:
> > @mehdi-goli and @Rbiessy - Should we have a section in the README which says we no longer support `TARGET` - This might help anyone who used the library before with TARGET and puller a later version after this change.
> 
> Sure that'd be nice. Also the README still mentions `TARGET` in a couple of places: for PowerVR support and in the cross-compile section.

I removed the missed TARGETs.

Do I mention that fact that TARGET is no more supported at the beginning  (as a note below the main title) or in the 'Setup' section?

Rbiessy:
> Do I mention the fact that TARGET is no more supported at the beginning (as a note below the main title) or in the 'Setup' section?

I'd say it makes sense to have it in 'Setup'.

pgorlani:
I added a note about TARGET removal in the README, and an error message in CMake in case of TARGET is set.

README.md
```diff
@@ -310,6 +310,11 @@ added to the `CMAKE_PREFIX_PATH` when building SYCL-BLAS (see
 
 ## Setup
 
+**IMPORTANT NOTE:** The `TARGET` CMake variable is no more supported, it has
+been replaced by `TUNING_TARGET`, which accepts the same options.
+`TUNING_TARGET` does not effect compilation, but only the tuning configuration.
```
Rbiessy, line 315: I don't understand why we say that `TUNING_TARGET` doesn't affect compilation. It is still used to determine which file to include in level 2 or 3 [backend.hpp](https://github.com/codeplaysoftware/sycl-blas/blob/master/src/interface/blas3/backend/backend.hpp). I think you want to say that it is not affecting the sycl target anymore. I'm not sure it's worth explaining that, we could just say that `TARGET` is not supported.
Same comment for `ConfigureSYCLBLAS.cmake`.

pgorlani: Good point!

README.md
```diff
@@ -310,6 +310,12 @@ added to the `CMAKE_PREFIX_PATH` when building SYCL-BLAS (see
 
 ## Setup
 
+**IMPORTANT NOTE:** The `TARGET` CMake variable is no more supported, it has
```
fabiomestre, line 313:
```suggestion
**IMPORTANT NOTE:** The `TARGET` CMake variable is no longer supported. It has
```

README.md
```diff
@@ -310,6 +310,12 @@ added to the `CMAKE_PREFIX_PATH` when building SYCL-BLAS (see
...
+been replaced by `TUNING_TARGET`, which accepts the same options.
+`TUNING_TARGET` effects only the tuning configuration, no more the target
```
fabiomestre, line 315:
```suggestion
`TUNING_TARGET` affects only the tuning configuration and has no effect on the target
```

cmake/Modules/ConfigureSYCLBLAS.cmake
```diff
@@ -64,10 +64,18 @@ if(NAIVE_GEMM)
...
-SET(TARGET "DEFAULT_CPU" CACHE STRING "Default Platform 'DEFAULT_CPU'")
-SET(BACKEND_DEVICE ${TARGET})
-message(STATUS "${TARGET} is chosen as a backend platform")
+# the TUNING_TARGET variable defines the platform for which the sycl library is tuned
+SET(TUNING_TARGET "DEFAULT_CPU" CACHE STRING "Default Platform 'DEFAULT_CPU'")
+message(STATUS "${TUNING_TARGET} is chosen as a tuning target")
+
+if(DEFINED TARGET)
+  message(FATAL_ERROR
+            "\nSetting the TARGET CMake variable is no more supported. "
```
fabiomestre, line 73:
```suggestion
            "\nSetting the TARGET CMake variable is no longer supported. "
```

README.md
fabiomestre: I got confused with the wording here. I think this should make it a bit more clear.

pgorlani: Sure, thanks @fabiomestre!

fabiomestre (approved):

Rbiessy (approved):

PR #357, opened 2022-11-25
Add support for the rotmg api (blas level 1)

fabiomestre:
Add support for the rotmg api (blas level 1).
Fix minor bug on documentation of rotm.
Fix minor bug in python script for rotg.
Fix bug when compiling rotg with double support.
Update test framework to accept templated test parameters.


benchmark/syclblas/blas1/rotmg.cpp
```diff
@@ -0,0 +1,149 @@
...
+  auto buf_d1 = utils::make_quantized_buffer<scalar_t>(ex, d1);
+  auto buf_d2 = utils::make_quantized_buffer<scalar_t>(ex, d2);
+  auto buf_x1 = utils::make_quantized_buffer<scalar_t>(ex, x1);
+  auto buf_y1 = utils::make_quantized_buffer<scalar_t>(ex, y1);
+  auto buf_param = utils::make_quantized_buffer<scalar_t>(ex, param);
+
+#ifdef BLAS_VERIFY_BENCHMARK
+
+
+
```
hjabird, line 56: There are a few places where I feel like the number of blank lines is excessive.

benchmark/syclblas/blas1/rotmg.cpp
```diff
@@ -0,0 +1,149 @@
...
+      ex.get_policy_handler().copy_to_host(buf_verify_y1, &y1_verify, 1);
+  auto event5 = ex.get_policy_handler().copy_to_host(
+      device_param, param_verify.data(), param_size);
+
+
+  ex.get_policy_handler().wait(event1);
+  ex.get_policy_handler().wait(event2);
+  ex.get_policy_handler().wait(event3);
+  ex.get_policy_handler().wait(event4);
+
```
hjabird, line 95: We don't wait for `event5`. Why?

cmake/CmakeFunctionHelper.cmake
```diff
@@ -451,6 +451,57 @@ function(generate_blas_rotg_return_objects blas_level func)
   add_sycl_to_target(TARGET ${func} SOURCES ${FUNC_SRC})
 endfunction(generate_blas_rotg_return_objects)
 
+# blas function for generating source code for the rotg operator (asynchronous version with containers)
+function(generate_blas_rotmg_objects blas_level func)
+  set(LOCATION "${SYCLBLAS_GENERATED_SRC}/${blas_level}/${func}/")
+  foreach (executor ${executor_list})
+    foreach (data ${data_list})
+      cpp_type(cpp_data ${data})
+      set(container_list_in)
```
hjabird, line 460: I don't see where `container_list_in` gets used.

cmake/CmakeFunctionHelper.cmake
```diff
@@ -451,6 +451,57 @@ function(generate_blas_rotg_return_objects blas_level func)
...
+                        ${executor}
+                        ${cpp_data}
+                        ${container0}
+                        ${container1}
+                        ${container2}
+                        ${container3}
+                        ${container4}
+                        ${file_name}
+                        MAIN_DEPENDENCY ${SYCLBLAS_SRC}/interface/${blas_level}/${func}.cpp.in
+                        DEPENDS ${SYCLBLAS_SRC_GENERATOR}/py_gen_blas_rotg.py
```
hjabird, line 485: Why does it depend on rotg?

src/operations/blas1_trees.hpp
```diff
@@ -619,6 +619,278 @@ Rotg<operand_t>::adjust_access_displacement() {
...
+  using sltc_matrix = zero; /* sin less than co-sin */
+  using rescaled_matrix = m_one;
+  using unit_matrix = m_two;
+
+  /* Gamma is a magic number used to re-scale the output and avoid underflows or
+   * overflows. Consult the papers above for more info */
+  constexpr value_t gamma = static_cast<value_t>(4096.0);
+
+  /* Square root of gamma. It is hardcoded to avoid computing it on every call */
+  constexpr value_t gamma_sq = static_cast<value_t>(1.67772e7);
```
Rbiessy, line 666: That's not the square root of 4096 but its square. I find it clearer to set it to `gamma * gamma`. Similarly for the variable below, it can be set to `1 / gamma_sq`.

include/interface/blas1_interface.h
```diff
@@ -243,6 +243,45 @@ typename executor_t::policy_t::event_t _rotm(
     executor_t &ex, index_t _N, container_0_t _vx, increment_t _incx,
     container_1_t _vy, increment_t _incy, container_2_t _param);
 
+/**
+ * Given the Cartesian coordinates (x1, y1) of a point, the rotmg routines
+ * compute the components of a modified Givens transformation matrix H that
+ * zeros the y-component of the resulting point.
```
hjabird, line 249: Missing that the x-coord should be correctly scaled to be equal to x1. Comment applies to everywhere this description is used.

python_generator/py_gen_blas_rotmg.py
```diff
@@ -0,0 +1,106 @@
...
+# *
+# *  Unless required by applicable law or agreed to in writing, software
+# *  distributed under the License is distributed on an "AS IS" BASIS,
+# *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# *  See the License for the specific language governing permissions and
+# *  limitations under the License.
+# *
+# *  SYCL-BLAS: BLAS implementation using SYCL
+# *
+# *  @filename py_gen_blas_rotg.py
```
hjabird, line 22:
```suggestion
# *  @filename py_gen_blas_rotmg.py
```

src/operations/blas1_trees.hpp
```diff
@@ -619,6 +619,278 @@ Rotg<operand_t>::adjust_access_displacement() {
...
+  using two = constant<value_t, const_val::two>;
+  using m_one = constant<value_t, const_val::m_one>;
+  using m_two = constant<value_t, const_val::m_two>;
+
+  using error = two;
+
+  using clts_matrix = one; /* co-sin less than sin */
+  using sltc_matrix = zero; /* sin less than co-sin */
+  using rescaled_matrix = m_one;
+  using unit_matrix = m_two;
```
hjabird, line 659: I find this confusing. Constant types are being aliased as types that describe themselves as matrices?

src/operations/blas1_trees.hpp
hjabird, line 666: This is the square of gamma.

src/operations/blas1_trees.hpp
```diff
@@ -619,6 +619,278 @@ Rotg<operand_t>::adjust_access_displacement() {
...
+
+  /* Gamma is a magic number used to re-scale the output and avoid underflows or
+   * overflows. Consult the papers above for more info */
+  constexpr value_t gamma = static_cast<value_t>(4096.0);
+
+  /* Square root of gamma. It is hardcoded to avoid computing it on every call */
+  constexpr value_t gamma_sq = static_cast<value_t>(1.67772e7);
+
+  /* Inverse of the square root of gamma (i.e. 1 / sqrt(gamma)) */
+  constexpr value_t inv_gamma_sq = static_cast<value_t>(5.96046e-8);
```
hjabird, line 669: $$ \frac{1}{\gamma^2}$$

src/operations/blas1_trees.hpp
```diff
@@ -619,6 +619,278 @@ Rotg<operand_t>::adjust_access_displacement() {
...
+  /* If the input is of the form (c, 0), then we already have the expected output.
+   * No calculations needed in this case */
+  else if (d2 == zero::value() || y1 == zero::value()) {
+    flag = unit_matrix::value();
+  }
+  /* If the input is of the form (0, c) - just need to swap the elements.
+   * Scaling may be needed */
+  else if ((d1 == zero::value() || x1 == zero::value()) && d2 > zero::value()) {
+    flag = clts_matrix::value();
+    /* clts_matrix assumes h12 and h21 values. But they still need to be set
```
hjabird, line 706: I'm not seeing where this happens. Can you say it will happen later and label it perhaps?

test/unittest/blas1/blas1_rotmg_test.cpp
```diff
@@ -0,0 +1,256 @@
...
+);
+
+template <class T>
+static std::string generate_name(
+    const ::testing::TestParamInfo<combination_t<T>>& info) {
+  T d1, d2, x1, y1;
+  BLAS_GENERATE_NAME(info.param, d1, d2, x1, y1);
+}
+
+BLAS_REGISTER_TEST_ALL(Rotg, combination_t, combi, generate_name);
```
Rbiessy, line 256: Missing new line at the end of the file

test/unittest/blas1/blas1_rotmg_test.cpp
```diff
@@ -0,0 +1,256 @@
...
+    test.validate_with_rotm();
+  }
+}
+
+const auto combi = ::testing::Combine(
+    ::testing::Values(0, 15.5, -2.2, max_float, min_float),  // d1
+    ::testing::Values(0, 3.0, -2.2, max_float, min_float),   // d2
+    ::testing::Values(0, 12.1, -7.3, max_float, min_float),  // x1
+    ::testing::Values(0, 0.5, -4.3, max_float, min_float)    // y1
+);
```
hjabird, line 247: There are so many possible code paths that I think you need many more combinations are needed. Ideally we'd need a way to check we have good code coverage.

test/unittest/blas1/blas1_rotmg_test.cpp
hjabird, line 256:
```suggestion
BLAS_REGISTER_TEST_ALL(Rotmg, combination_t, combi, generate_name);
```

hjabird (commented):
The rotmg algorithm is compicated enough that it might as well be a black box. I think we need to test more thoroughly to make up for the fact that its difficult to review.

src/operations/blas1_trees.hpp
fabiomestre: Don't know why I assumed it was square root. But that's why I left the magic numbers (there is no constexpr of square root). Since it's not square root, what you suggest makes total sense!

src/operations/blas1_trees.hpp
fabiomestre: I kind of got this notation from the papers tbh. I'm open to suggestions on how to make it more clear. Each flag represent a specific matrix type which is what this aliases mean.

include/interface/blas1_interface.h
fabiomestre: I don't really understand what you mean here. Why does the x-coord need to be scaled?

cmake/CmakeFunctionHelper.cmake
fabiomestre: Bad copy paste from rotg ¯\_(ツ)_/¯. Will fix

cmake/CmakeFunctionHelper.cmake
fabiomestre: Actually I think it should come in for y parameter which is readonly. So need to update the foreach.

benchmark/syclblas/blas1/rotmg.cpp
fabiomestre: It's a mistake. But actually it doesn't matter much here since I'm not checking the output of param in the benchmark. But will update to keep the code consistent

benchmark/syclblas/blas1/rotmg.cpp
fabiomestre: Agreed. Will try to remove a few of this situations

src/operations/blas1_trees.hpp
fabiomestre: If you go to the scaling part of the algorithm. The matrix will be rescaled based on the value of d1 or d2. In those cases. h12 or h21 might be rescaled as well. Where would you like me to put the label?

test/unittest/blas1/blas1_rotmg_test.cpp
```diff
@@ -0,0 +1,256 @@
...
+  scalar_t h11_sycl = sycl_out.param[1];
+  scalar_t h12_sycl = sycl_out.param[3];
+  scalar_t h21_sycl = sycl_out.param[2];
+  scalar_t h22_sycl = sycl_out.param[4];
+
+  scalar_t flag_ref = sycl_out.param[0];
+  scalar_t h11_ref = sycl_out.param[1];
+  scalar_t h12_ref = sycl_out.param[3];
+  scalar_t h21_ref = sycl_out.param[2];
+  scalar_t h22_ref = sycl_out.param[4];
```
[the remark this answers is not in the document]
fabiomestre: yes, it was bad copy paste. Will update.

test/unittest/blas1/blas1_rotmg_test.cpp
fabiomestre: It's quite hard to even think of tests that can trigger every code path. Many of them rely on overflow or underflows  to be triggered.

In addition, I think the current unit test framework is not adequate for rotmg testing. I need to define tests with very specific inputs and using different values for half, float and double (i.e. a templated combi). Instead of combinations of all of them which increases the numbers of tests by a lot.

So, while I can definitely work on it if we agree it makes sense. It will require some effort to update the framework.

fabiomestre: Nevermind, found a clean way to do this without modifying the test framework. I will add more tests for specific code paths :)

include/interface/blas1_interface.h
hjabird: We are solving for $$H$$ where
```math
\begin{Bmatrix}  x  \\ y  \end{Bmatrix} 
= \begin{Bmatrix}  x_i  \\ 0 \end{Bmatrix} 
= H 
\begin{Bmatrix} x_i \sqrt{d_1}  \\ y_i \sqrt{d_2} \end{Bmatrix} 
```
where at the end $$y=0$$ and $$x=x_i$$.

You've stated the $$y=0$$ bit with *that zeros the y-component of the resulting point*, but not that we also need $$x=x_i$$.

benchmark/syclblas/blas1/rotmg.cpp
```diff
@@ -0,0 +1,149 @@
...
+  std::vector<scalar_t> param = std::vector<scalar_t>(param_size);
+  scalar_t d1 = blas_benchmark::utils::random_data<scalar_t>(1)[0];
+  scalar_t d2 = blas_benchmark::utils::random_data<scalar_t>(1)[0];
+  scalar_t x1 = blas_benchmark::utils::random_data<scalar_t>(1)[0];
+  scalar_t y1 = blas_benchmark::utils::random_data<scalar_t>(1)[0];
+
+  ExecutorType& ex = *executorPtr;
+
+  auto buf_d1 = utils::make_quantized_buffer<scalar_t>(ex, d1);
+  auto buf_d2 = utils::make_quantized_buffer<scalar_t>(ex, d2);
```
mehdi-goli, line 48: These should be re-written as there is no `quantized_buffer` anymore

benchmark/syclblas/blas1/rotmg.cpp
```diff
@@ -0,0 +1,149 @@
...
+  scalar_t d1_verify = d1;
+  scalar_t d2_verify = d2;
+  scalar_t x1_verify = x1;
+  scalar_t y1_verify = y1;
+  std::vector<scalar_t> param_verify = param;
+
+  auto buf_verify_d1 = utils::make_quantized_buffer<scalar_t>(ex, d1_verify);
+  auto buf_verify_d2 = utils::make_quantized_buffer<scalar_t>(ex, d2_verify);
+  auto buf_verify_x1 = utils::make_quantized_buffer<scalar_t>(ex, x1_verify);
+  auto buf_verify_y1 = utils::make_quantized_buffer<scalar_t>(ex, y1_verify);
```
mehdi-goli, line 73: Same here

test/unittest/blas1/blas1_rotmg_test.cpp
```diff
@@ -0,0 +1,256 @@
...
+);
+
+template <class T>
+static std::string generate_name(
+    const ::testing::TestParamInfo<combination_t<T>>& info) {
+  T d1, d2, x1, y1;
+  BLAS_GENERATE_NAME(info.param, d1, d2, x1, y1);
+}
+
+BLAS_REGISTER_TEST_ALL(Rotmg, combination_t, combi, generate_name);
```
t4c1, line 256: Missing newline at the end of the file.

benchmark/syclblas/blas1/rotmg.cpp
fabiomestre: I think I addressed this. Let me know if I missed something

src/operations/blas1_trees.hpp
```diff
@@ -653,20 +653,21 @@ SYCL_BLAS_INLINE typename Rotmg<operand_t>::value_t Rotmg<operand_t>::eval(
...
+  using unit_flag = m_two;
 
   /* Gamma is a magic number used to re-scale the output and avoid underflows or
    * overflows. Consult the papers above for more info */
   constexpr value_t gamma = static_cast<value_t>(4096.0);
 
-  /* Square root of gamma. It is hardcoded to avoid computing it on every call */
-  constexpr value_t gamma_sq = static_cast<value_t>(1.67772e7);
+  /* Square of gamma. It is hardcoded to avoid computing it on every call */
+  constexpr value_t gamma_sq = static_cast<value_t>(gamma * gamma);
```
Rbiessy, line 666: The `static_cast` is not needed, same for the global cast below.

src/operations/blas1_trees.hpp
fabiomestre: As discussed on private chate, updated the comment on the rescale section to be more clear

test/unittest/blas1/blas1_rotmg_test.cpp
fabiomestre: Updated the tests to test as many code paths as possible

src/operations/blas1_trees.hpp
```diff
@@ -619,6 +619,276 @@ Rotg<operand_t>::adjust_access_displacement() {
...
+  using error = two;
+
+  using clts_flag = one; /* co-sin less than sin */
+  using sltc_flag = zero; /* sin less than co-sin */
+  using rescaled_flag = m_one;
+  using unit_flag = m_two;
+
+  /* Gamma is a magic number used to re-scale the output and avoid underflows or
+   * overflows. Consult the papers above for more info */
+  constexpr value_t gamma = static_cast<value_t>(4096.0);
```
hjabird, line 663: Should this be different for double or half?

fabiomestre: The papers don't distinguish between float and double. So I don't think it needs to be different (and it will fail reference function checks). Regarding half 4096 * 4096 seems to be larger than what it is able to represent. So maybe I think that's a good point. Will check.

src/operations/blas1_trees.hpp
```diff
@@ -619,6 +619,275 @@ Rotg<operand_t>::adjust_access_displacement() {
...
+
+  /* Square of gamma. It is hardcoded to avoid computing it on every call */
+  constexpr value_t gamma_sq = gamma * gamma;
+
+  /* Inverse of the square of gamma (i.e. 1 / (gamma * gamma)) */
+  constexpr value_t inv_gamma_sq = static_cast<value_t>(1.0) / gamma;
```
hjabird, line 669: Do these ever get used? The netlib lapack uses 1 / (gamma * gamma), instead of 1 * inv_gamma_sq too.  I'd be interested if these gave different results!

Rbiessy (commented):
We should update the README table to mention that we support rotmg. We forgot to do it for the other operators too.

src/operations/blas1_trees.hpp
fabiomestre: Good catch. This is a bug that I introduced when moving from magic numbers to variables. And somehow justified to myself with cblas implementation being different.

Rbiessy (approved):

t4c1 (approved):

PR #361, opened 2022-12-09
Added Tensorcore support for GEMM using joint_matrix extension

muhammad-tanvir-1211:
This PR adds support for Tensorcores for GEMM using `joint_matrix` extension from DPC++. The code is enabled only for the latest NVIDIA GPUs (Ampere Arch onwards). Support was added for all matrix sizes by padding the `shared local memory` with `0`. We support all 10 `wmma` configurations (except the integer and double configurations) listed [here](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#wmma).

To run the Tensorcore specific code, we need to set `SB_ENABLE_JOINT_MATRIX` environment variable to `1`. This environment variable can be set at runtime and the right kernels would only get invoked if this variable is set to `1`, it won't work for any other value.


src/operations/blas3/gemm_local_tensorcore.hpp
```diff
@@ -0,0 +1,814 @@
...
+ *                   level tiles to use, see Tile
+ * @tparam TransA  iff true, matrix A will be transposed on the fly
+ * @tparam TransB  iff true, matrix B will be transposed on the fly
+ * @tparam element_t  type of matrix elements
+ * @tparam is_beta_zero True if beta == 0.
+ * @tparam VectorSize The packet size to be used for vectorization.
+ * @tparam batch_type the type of batch strideded /interleaved
+ * @tparam UseTensorcores boolean parameter to decide whether to use Tensorcores
+ * or not
+ */
```
hjabird, line 68: I think the iff either looks like a misspelling or unnecessary cleverness.

src/operations/blas3/gemm_local_tensorcore.hpp
```diff
@@ -0,0 +1,814 @@
...
+  static constexpr index_t item_rows = tile_type::item_rows;
+  static constexpr index_t item_cols = tile_type::item_cols;
+  static constexpr index_t wg_rows = tile_type::wg_rows;
+  static constexpr index_t wg_cols = tile_type::wg_cols;
+  static constexpr index_t sg_rows = tile_type::sg_rows;
+  static constexpr index_t sg_cols = tile_type::sg_cols;
+  static constexpr index_t tl_rows = tile_type::tl_rows;
+  static constexpr index_t tl_cols = tile_type::tl_cols;
+  static constexpr index_t tile_size = tl_rows * tl_cols;
+  static constexpr index_t sg_size = 32;
```
hjabird, line 97: Should this be computed from sg_rows and sg_cols? If it has to be 32, perhaps we should then have a static assert.

src/operations/blas3/gemm_local_tensorcore.hpp
```diff
@@ -0,0 +1,814 @@
...
+      : a_(A),
+        b_(B),
+        c_(C),
+        alpha_(alpha),
+        beta_(beta),
+        batch_size_(batch_size) {}
+
+  /*!
+   * @brief Get the type of this GemmFactory as a human readable string.
+   */
```
hjabird, line 177: Can some of these docs just be `/// Get the type of this GemmFactory as a human readable string.`? It's the same [to doxygen](https://doxygen.nl/manual/docblocks.html).

src/operations/blas3/gemm_local_tensorcore.hpp
```diff
@@ -0,0 +1,814 @@
...
+  static SYCL_BLAS_INLINE std::string get_type_string() noexcept {
+    std::ostringstream str{};
+    str << "Gemm <" << double_buffer << ", " << nbc_a << ", " << nbc_b << ", "
+        << cl_elems * sizeof(element_t) << ", " << tile_type::get_type_string()
+        << ", " << type_string<value_t>::get_value() << "gemm_memory:local, "
+        << "gemm_algorithm:standard, "
+        << "gemm_vectorization:full, "
+        << "vector size" << VectorSize << ", batch_type:strided>";
```
hjabird, line 185: Should this mention your tensor core magic?

src/operations/blas3/gemm_local_tensorcore.hpp
```diff
@@ -0,0 +1,814 @@
...
+   * block-column of B (transpose).
+   *
+   * This is a collective operation between all items in as work-group.
+   * This method should actually be a generic lambda (as in C++20), it
+   * only forwards parameters from GemmFactory::run().
+   *
+   * @tparam check_m_limit  iff true, check if row indexes of C are
+   *                        out-of-bound
+   * @tparam check_n_limit  iff true, check if no indexes of C are
+   *                        out-of-bound
```
hjabird, line 387: column indices?

src/operations/blas3/gemm_local_tensorcore.hpp
```diff
@@ -0,0 +1,814 @@
...
+
+  /*!
+   * @brief Store the computed gemm result to the C matrix after performing
+   * input and output scaling. Output scaling only takes place if beta is
+   * non-zero.
+   *
+   * @tparam check_m_limit  iff true, check if row indexes of C are
+   *                        out-of-bound
+   * @tparam check_n_limit  iff true, check if no indexes of C are
+   *                        out-of-bound
```
hjabird, line 456: When referring to a matrix, I think this should be indices - eg see [wikipedia](https://en.wikipedia.org/wiki/Index_notation). 

This comment also applies to use of indexes elsewhere.

src/operations/blas3/gemm_local_tensorcore.hpp
```diff
@@ -0,0 +1,814 @@
...
+   * @tparam ScratchPointerType the type of shared local memory pointer
+   * @tparam OutputPointerType the type of C
+   * @param mc the computed boundary limit of m in matrix C
```
hjabird, line 459:
```suggestion
   * @tparam OutputPointerType the type of C
   * @tparam CType .... ?
   * @param mc the computed boundary limit of m in matrix C
```

hjabird (commented):
Initial review. I'll try and review the detail of the implementation later.

include/executors/executor.h
```diff
@@ -75,11 +75,11 @@ class Executor {
             bool NbcB, int ClSize, typename tile_type, bool TransA, bool TransB,
             typename element_t, bool is_beta_zero, int GemmMemoryType,
             int GemmAlgorithm, int GemmVectorization, int VectorSize,
-            int BatchType>
+            int BatchType, bool UseTensorcores>
```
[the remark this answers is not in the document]
muhammad-tanvir-1211: Changed this to `UseJointMatrix`.

include/operations/blas3_trees.h
```diff
@@ -124,6 +125,12 @@ struct Tile {
   static constexpr int sg_cols = SgCols;
   static constexpr int tl_rows = TlRows;
   static constexpr int tl_cols = TlCols;
+
+  static constexpr int sg_M = jm_M;
+  static constexpr int sg_N = jm_N;
+  static constexpr int sg_K = jm_K;
```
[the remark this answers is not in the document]
muhammad-tanvir-1211: Changed these to `joint_matrix_*` instead of `sg_*`

src/executors/kernel_constructor.hpp
```diff
@@ -255,5 +269,59 @@ static SYCL_BLAS_INLINE cl::sycl::event execute_tree(queue_t q_,
...
+      printf(
+          "Configuration not supported.\nDevice Required: NVIDIA GPU\nCompute "
+          "Capability >= 8.0 required to use TF32 precision\n");
+      return ev;
+    }
+
+    ev = q_.submit(cg1);
+    return ev;
+  } catch (cl::sycl::exception e) {
+    std::cerr << e.what() << std::endl;
```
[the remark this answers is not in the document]
muhammad-tanvir-1211: Here again, this call to `std::cerr` was copied from the existing definition of `execute_tree`. I can make the suggested change if that is the better way of doing this.

src/operations/blas3/gemm_local_tensorcore.hpp
muhammad-tanvir-1211: This way of adding documentation is actually copied from the `src/operations/blas3/gemm_local.hpp` file. I think it is the same in all other `gemm_*` files in this directory.

src/executors/kernel_constructor.hpp
```diff
@@ -255,5 +269,59 @@ static SYCL_BLAS_INLINE cl::sycl::event execute_tree(queue_t q_,
     return ev;
   }
 }
+
+template <int using_local_memory, typename queue_t, typename expression_tree_t>
+static SYCL_BLAS_INLINE cl::sycl::event execute_tree(queue_t q_,
+                                                     expression_tree_t t,
+                                                     size_t _localSize_x,
+                                                     size_t _localSize_y,
```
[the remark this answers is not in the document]
muhammad-tanvir-1211: I can make this change, but that may go against whatever we have in SYCL-BLAS already. The existing call to `execute_tree` in SYCL-BLAS uses `size_t` as the data type for these variables.

src/interface/blas3/backend/nvidia_gpu.hpp
```diff
@@ -53,31 +54,70 @@ typename executor_t::policy_t::event_t _gemm(
                                                                     _ldc,
                                                                     batch_size);
   }
-  if (_M < 512 && _N < 512 && _K < 512) {
+
+#ifdef SB_ENABLE_JOINT_MATRIX
+  const char en_joint_matrix = *std::getenv("SB_ENABLE_JOINT_MATRIX");
```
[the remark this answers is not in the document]
muhammad-tanvir-1211: Fixed this in the latest changes.

src/interface/blas3/backend/nvidia_gpu.hpp
```diff
@@ -53,31 +54,70 @@ typename executor_t::policy_t::event_t _gemm(
...
-                                                                _lda, _b, _ldb,
-                                                                _beta, _c, _ldc,
-                                                                batch_size);
-  } else {
+        static_cast<int>(gemm_batch_type_t::strided),
+        false>::template _select_gemm(ex, _M, _N, _K, _alpha, _a, _lda, _b,
+                                      _ldb, _beta, _c, _ldc, batch_size);
+  }
+
+#else
```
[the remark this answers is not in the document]
muhammad-tanvir-1211: Added this in the code.

src/executors/kernel_constructor.hpp
```diff
@@ -255,5 +269,59 @@ static SYCL_BLAS_INLINE cl::sycl::event execute_tree(queue_t q_,
...
+                                                     size_t _globalSize_x,
+                                                     size_t _globalSize_y,
+                                                     size_t _shMem) {
+  using value_t =
+      typename LocalMemoryType<using_local_memory, expression_tree_t>::type;
+
+  auto localSize_x = _localSize_x;
+  auto localSize_y = _localSize_y;
+  auto globalSize_x = _globalSize_x;
+  auto globalSize_y = _globalSize_y;
```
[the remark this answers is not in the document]
hjabird: Fair enough.

src/executors/kernel_constructor.hpp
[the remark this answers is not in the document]
hjabird: I think exceptions are preferable to printing errors, but if this has been copied its probably better to remain consistent.

src/operations/blas3/gemm_local_tensorcore.hpp
t4c1: iff is used in math for "if and only if", but I am not sure if most developers are familiar with this.

cmake/Modules/FindDPCPP.cmake
```diff
@@ -66,6 +67,10 @@ else()
     INTERFACE_INCLUDE_DIRECTORIES "${DPCPP_BIN_DIR}/../include/sycl")
 endif()
 
+if (DPCPP_SYCL_TARGET MATCHES "nvptx64-nvidia-cuda")
```
pgorlani, line 70: Hi @muhammad-tanvir-1211, read this first.

I suggest to
* check also if `DPCPP_SYCL_ARCH >= sm_80`, see the comment below;
* specify in the README that the user need to set `DPCPP_SYCL_ARCH >= sm_80` in order to use joint matrices. If the user doesn't specify `DPCPP_SYCL_ARCH` this defaults to `sm_50`.

cmake/Modules/FindDPCPP.cmake
```diff
@@ -77,8 +82,20 @@ function(add_sycl_to_target)
     ${ARGN}
   )
   target_compile_options(${SB_ADD_SYCL_TARGET} PUBLIC ${DPCPP_FLAGS})
+  if (DPCPP_SYCL_TARGET MATCHES "nvptx64-nvidia-cuda")
+    target_compile_options(${SB_ADD_SYCL_TARGET} PUBLIC -Xsycl-target-backend 
+                              PUBLIC --cuda-gpu-arch=sm_80
```
pgorlani, line 87: Basically, if you compile every NVIDIA target for `sm_80`, you will get a runtime error in case of the actual device architecture is less than `sm_80`. Here again, I suggest setting the defines by checking the `DPCPP_SYCL_ARCH`.

cmake/Modules/FindDPCPP.cmake
```diff
@@ -50,6 +50,7 @@ if(NOT "${DPCPP_SYCL_ARCH}" STREQUAL "")
   elseif("${DPCPP_SYCL_TARGET}" STREQUAL "nvptx64-nvidia-cuda")
     list(APPEND DPCPP_FLAGS "-Xsycl-target-backend")
     list(APPEND DPCPP_FLAGS "--cuda-gpu-arch=${DPCPP_SYCL_ARCH}")
+    list(APPEND DPCPP_FLAGS "-fsycl;-fsycl-targets=${DPCPP_SYCL_TARGET};-Xclang;-cl-mad-enable;-fsycl-unnamed-lambda")
```
pgorlani, line 53: I don't think you need to repeat `-fsycl;-fsycl-targets=${DPCPP_SYCL_TARGET};`, which is already set in L45.
Probably, you shouldn't need `-fsycl-unnamed-lambda` as well. 

If needed, I suggest applying `-Xclang;-cl-mad-enable;` within the `if` in L70.

src/executors/kernel_constructor.hpp
```diff
@@ -255,5 +269,57 @@ static SYCL_BLAS_INLINE cl::sycl::event execute_tree(queue_t q_,
...
+    auto vendor =
+        q_.get_device().template get_info<cl::sycl::info::device::vendor>();
+    auto compute_capability = std::stof(
+        q_.get_device()
+            .template get_info<cl::sycl::info::device::backend_version>());
+
+    if (compute_capability < 8.0 ||
+        vendor.find("NVIDIA Corporation") == std::string::npos) {
+      throw std::runtime_error("\nConfiguration not supported.\nDevice Required: NVIDIA GPU\nCompute "
+          "Capability >= 8.0 required to use Joint Matrix extension.\n");
```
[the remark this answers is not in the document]
pgorlani: If you run a code compiled for `sm_80` on a previous architecture you will get an error similar to the following one the first time a device code is encountered.
```
  PI CUDA ERROR:                                                                  
          Value:           209                                                    
          Name:            CUDA_ERROR_NO_BINARY_FOR_GPU                           
          Description:     no kernel image is available for execution on the device
          Function:        build_program                                          
          Source Location: /llvm/sycl/plugins/cuda/pi_cuda.cpp:559
 ```
 
 For this reason, it's unlikely that the above error message will be displayed. However, I suggest making a little test to see if this check is encountered before running any device code.

File: src/interface/blas3/backend/nvidia_gpu.hpp [hunk 19]
```diff
@@ -53,31 +54,70 @@ typename executor_t::policy_t::event_t _gemm(
                                                                     _ldc,
                                                                     batch_size);
   }
-  if (_M < 512 && _N < 512 && _K < 512) {
+
+#ifdef SB_ENABLE_JOINT_MATRIX
+  const char* en_joint_matrix = std::getenv("SB_ENABLE_JOINT_MATRIX");
```
Rbiessy, line 59: I can see you've described this in the PR description. Why do we want users to set an environment variable? That's not very intuitive and not something we've done in SYCL-BLAS before. Shouldn't we use joint matrix whenever we detect that it is supported and that we support the sizes?

pgorlani: I would be a good idea introducing a CMake flag similar to the existing [ones](https://github.com/codeplaysoftware/sycl-blas/blob/master/doc/Gemm.md#relevant-cmake-variables) for enabling the joint matrices.

In any case, we could adopt an approach similar to NVIDIA by having an [environment variable](https://docs.nvidia.com/cuda/cutensor/user_guide.html#environment-variables) for disabling the TF32, such as `NVIDIA_TF32_OVERRIDE=0`.

muhammad-tanvir-1211: @mehdi-goli suggested that we should compile all supported kernels and then let the selection of the appropriate kernel depend on the environment variable. This is inline with what NVIDIA does with their DNN and BLAS libraries where they use the `NVIDIA_TF32_OVERRIDE` environment variable to select whether Tensorcores get invoked or not. So we are trying to mimic the same in our libraries for the `joint_matrix` based implementations.

src/operations/blas3/gemm_load_store.hpp
```diff
@@ -62,14 +62,40 @@ struct Packetize {
...
+    } else if constexpr (std::is_same<
+                             cl::sycl::multi_ptr<
+                                 cl::sycl::ext::oneapi::experimental::bfloat16,
+                                 address_t::local_space>,
+                             DestPointerType>::value) {
+      using dtype = cl::sycl::ext::oneapi::experimental::bfloat16;
+      *dest = static_cast<dtype>(val);
+    } else {
+      using namespace cl::sycl::ext::oneapi::experimental::matrix;
+      *dest = round_to_tf32(val);
```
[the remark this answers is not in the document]
muhammad-tanvir-1211: Thanks for detecting this. We don't have support for `double` at the moment.

doc/Gemm.md
```diff
@@ -57,6 +57,8 @@ where A, B and C are matrices and alpha and beta are scalars.
 
 - `gemm_local.hpp` - Uses local memory for increased performance on platforms that have it. Supports full vectorization.
 
+- `gemm_local_tensorcore.hpp` - Uses Tensorcores + local memory for increased performance on platforms that have it. Does not support vectorization.
```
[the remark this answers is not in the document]
muhammad-tanvir-1211: Fixed this.

src/executors/kernel_constructor.hpp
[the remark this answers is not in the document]
muhammad-tanvir-1211: Fixed this to check for the Compute Capability through CMake.

cmake/Modules/FindDPCPP.cmake
muhammad-tanvir-1211: Addressed these changes in the latest commits.

src/operations/blas3/gemm_load_store.hpp
```diff
@@ -62,14 +62,40 @@ struct Packetize {
    * checking is required.
    * @tparam ld The leading dimension of the destination memory.
    */
-
+#ifdef SB_ENABLE_JOINT_MATRIX
```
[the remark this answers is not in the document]
muhammad-tanvir-1211: Fixed this in the latest commit.

cmake/Modules/FindDPCPP.cmake
```diff
@@ -66,6 +66,11 @@ else()
     INTERFACE_INCLUDE_DIRECTORIES "${DPCPP_BIN_DIR}/../include/sycl")
 endif()
 
+if (${DPCPP_SYCL_TARGET} STREQUAL "nvptx64-nvidia-cuda" AND ${DPCPP_SYCL_ARCH} STREQUAL "sm_80")
+  add_definitions(-DSB_ENABLE_JOINT_MATRIX=1)
+  list(APPEND DPCPP_FLAGS "-Xclang;-cl-mad-enable")
```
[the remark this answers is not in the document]
muhammad-tanvir-1211: Thank you for suggesting this change. It is a much efficient way of setting the flags.

cmake/Modules/FindDPCPP.cmake
```diff
@@ -66,6 +66,11 @@ else()
     INTERFACE_INCLUDE_DIRECTORIES "${DPCPP_BIN_DIR}/../include/sycl")
 endif()
 
+if (${DPCPP_SYCL_TARGET} STREQUAL "nvptx64-nvidia-cuda" AND ${DPCPP_SYCL_ARCH} STREQUAL "sm_80")
```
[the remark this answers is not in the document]
muhammad-tanvir-1211: I have fixed this by using string manipulation in CMake to extract the sm value and then checked if the sm value is greater than 80 or not

src/interface/blas3/backend/nvidia_gpu.hpp
Rbiessy: Right, that's a weird design but I'll resolve for now.

Rbiessy (approved):

src/sb_handle/kernel_constructor.hpp
```diff
@@ -255,5 +270,58 @@ static SYCL_BLAS_INLINE cl::sycl::event execute_tree(queue_t q_,
     return ev;
   }
 }
+
+template <int using_local_memory, typename queue_t, typename expression_tree_t>
```
mehdi-goli, line 274: @muhammad-tanvir-1211  why the kernel constructor requires to change? This file will construct the expression tree kernel irrespective of what the expression is. The tree has nothing to do with a specific operation and it just pass whatever is the expression. This file should not be touched at all. If you have anything requires specific execution rule, it should go to the `SB_Handle` class or gemm launcher class not here

muhammad-tanvir-1211: The change in kernel_constructor* is required because we want to use 2D nd-range instead of 1D. For the 2D range, we need the size of global and local range for both dimensions and then explicitly specify the nd-range before calling q.submit(). 

The eval functions also need to change since the existing functions are all hardcoded to accept 1D nd-range whereas my implementation requires a 2D nd-range.

src/operations/blas3/gemm_local_tensorcore.hpp
hjabird: Resolving this since it is - for better or worse - consistent with other parts of SYCL-BLAS.

src/operations/blas3/gemm_local_tensorcore.hpp
hjabird: The is mentioned in the equivalement code in gemm_local_joint_matrix.hpp.

src/operations/blas3/gemm_local_tensorcore.hpp
hjabird: Correct in gemm_local_joint_matrix.hpp.

src/operations/blas3/gemm_local_tensorcore.hpp
hjabird: Appears to have been applied.

src/operations/blas3/gemm_local_tensorcore.hpp
hjabird: Done in gemm_local_joint_matrix.hpp.

src/operations/blas3/gemm_load_store.hpp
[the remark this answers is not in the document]
hjabird: So should there be a `static_assert` then?

hjabird (approved):
Looks good - just don't want pgorlani's question about double to get missed.

src/operations/blas3/gemm_load_store.hpp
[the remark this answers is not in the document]
muhammad-tanvir-1211: Added this `static_assert` in the `gemm_local_joint_matrix.hpp` file.

src/operations/blas3/gemm_local_joint_matrix.hpp
```diff
@@ -0,0 +1,866 @@
...
+    const index_t it_mod_cl = item_id % cl_elems;
+    const index_t it_div_cl = item_id / cl_elems;
+
+    ptr_B += (trans_b ? (item_id / block_cols) * ldb +
+                            (wg_col + item_id % block_cols)
+                      : it_mod_cl + (wg_col + it_div_cl) * ldb);
+
+    n = n - wg_col - (trans_b ? item_id % block_cols : it_div_cl);
+    ptr_A += (trans_a ? (wg_row + it_div_cl) * lda + (it_mod_cl)
+                      : (wg_row + it_mod_brows) + (item_id / block_rows) * lda);
```
hjabird, line 319:
```suggestion
    const index_t it_mod_brows = item_id % block_rows;
    const index_t it_div_brows = item_id / block_rows;

    const index_t it_mod_cl = item_id % cl_elems;
    const index_t it_div_cl = item_id / cl_elems;

    ptr_B += (trans_b ? (item_id / block_cols) * ldb +
                            (wg_col + item_id % block_cols)
                      : it_mod_cl + (wg_col + it_div_cl) * ldb);

    n = n - wg_col - (trans_b ? item_id % block_cols : it_div_cl);
    ptr_A += (trans_a ? (wg_row + it_div_cl) * lda + (it_mod_cl)
                      : (wg_row + it_mod_brows) + it_div_brows * lda);
```
Why does block rows get pre-computed but not block_cols?

muhammad-tanvir-1211: Fixed this in the latest commit.

src/operations/blas3/gemm_load_store.hpp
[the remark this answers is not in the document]
JackAKirk: Good question. The answer is that `double` should be treated as `double`. There is no corresponding conversion to a lower precision like tf32 for float. I think the reason is that `tf32` is aimed more typically at deep learning, whereas the double case is intended for more traditional applications that require higher precision.

PR #362, opened 2022-12-14
Add baseline `gbmv` implementation

pgorlani:
This patch adds a baseline implementation for the band matrix-vector product.

Moreover, this patch adds support for testing `double`s on `NVIDIA_GPU`.


benchmark/clBench/clblast/blas2/gbmv.cpp
```diff
@@ -0,0 +1,190 @@
+/**************************************************************************
+ *
+ *  @license
+ *  Copyright (C) 2016 Codeplay Software Limited
```
fabiomestre, line 4:
```suggestion
 *  Copyright (C) 2022 Codeplay Software Limited
```

benchmark/clBench/clblast/blas2/gbmv.cpp
```diff
@@ -0,0 +1,190 @@
...
+    blas_benchmark::utils::update_counters(state, times);
+  }
+
+  blas_benchmark::utils::calc_avg_counters(state);
+}
+
+template <typename scalar_t>
+void register_benchmark(blas_benchmark::Args& args, ExecutorType* exPtr,
+                        bool* success) {
+  auto gemm_params = blas_benchmark::utils::get_gbmv_params<scalar_t>(args);
```
fabiomestre, line 164:
```suggestion
  auto gbmv_params = blas_benchmark::utils::get_gbmv_params<scalar_t>(args);
```

benchmark/clBench/clblast/blas2/gbmv.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+  for (auto p : gemm_params) {
```
fabiomestre, line 166:
```suggestion
  for (auto p : gbmv_params) {
```

benchmark/syclblas/blas2/gbmv.cpp
```diff
@@ -0,0 +1,176 @@
+/**************************************************************************
+ *
+ *  @license
+ *  Copyright (C) 2016 Codeplay Software Limited
```
fabiomestre, line 4:
```suggestion
 *  Copyright (C) 2022 Codeplay Software Limited
```

benchmark/syclblas/blas2/gbmv.cpp
```diff
@@ -0,0 +1,176 @@
...
+    blas_benchmark::utils::update_counters(state, times);
+  }
+
+  blas_benchmark::utils::calc_avg_counters(state);
+}
+
+template <typename scalar_t>
+void register_benchmark(blas_benchmark::Args& args, ExecutorType* exPtr,
+                        bool* success) {
+  auto gemm_params = blas_benchmark::utils::get_gbmv_params<scalar_t>(args);
```
fabiomestre, line 150:
```suggestion
  auto gbmv_params = blas_benchmark::utils::get_gbmv_params<scalar_t>(args);
```

benchmark/syclblas/blas2/gbmv.cpp
```diff
@@ -0,0 +1,176 @@
...
+
+  for (auto p : gemm_params) {
```
fabiomestre, line 152:
```suggestion
  for (auto p : gbmv_params) {
```

include/interface/blas2_interface.h
```diff
@@ -213,6 +213,43 @@ typename executor_t::policy_t::event_t _syr2(
...
+ y = alpha*A*x + beta*y
+
+ See the netlib blas interface documentation for more details of the high level
+ interface: https://netlib.org/lapack/explore-html/d6/d46/sgbmv_8f.html
+ */
+template <typename executor_t, typename index_t, typename element_t,
+          typename container_0_t, typename container_1_t, typename increment_t,
+          typename container_2_t>
+typename executor_t::policy_t::event_t _gbmv(
+    executor_t& ex,     // executor_t (sycl, parallel, serial, etc)
```
fabiomestre, line 232: I think it would be preferable to have the documentation for the parameters in a format compatible with doxygen (i.e. inside /** */ or /*! */) . It would also match what is being done on other sycl-blas api's. Same comment would apply to other declarations of _gbmv

test/unittest/blas2/blas2_gbmv_test.cpp
```diff
@@ -0,0 +1,131 @@
...
+  index_t ku;
+  bool trans;
+  scalar_t alpha;
+  scalar_t beta;
+  index_t incX;
+  index_t incY;
+  index_t lda_mul;
+  std::tie(m, n, kl, ku, alpha, beta, trans, incX, incY, lda_mul) = combi;
+
+  const char* t_str = trans ? "t" : "n";
```
fabiomestre, line 45: Shouldn't we test "c" as well. It is meant for complex but it should still work on real matrices.

src/operations/blas2/gemv.hpp
```diff
@@ -921,5 +921,85 @@ SYCL_BLAS_INLINE void GemvCol<Lower, Diag, Upper, Unit, lhs_t, matrix_t,
...
+                               is_transposed>::index_t
+Gbmv<lhs_t, matrix_t, vector_t, local_range, is_transposed>::get_size() const {
+  return matrix_.get_size();
+}
+template <typename lhs_t, typename matrix_t, typename vector_t,
+          uint32_t local_range, bool is_transposed>
+SYCL_BLAS_INLINE bool
+Gbmv<lhs_t, matrix_t, vector_t, local_range, is_transposed>::valid_thread(
+    cl::sycl::nd_item<1> ndItem) const {
+  return true;
```
fabiomestre, line 954: Maybe add a comment saying that this is handled elsewhere. I got a bit confused (probably due to not being familiar with existing level 2 operations)

include/interface/blas2_interface.h
fabiomestre: After looking at the code, it seems that level 2 apis are all using this style. While level 1 are using doxygen. I still prefer the standardized style but would understand if you want to keep this one since the whole blas2_interface.h is using it.

test/unittest/blas2/blas2_gbmv_test.cpp
pgorlani: Thanks for your review, @fabiomestre!

Is the complex data type supported? I think that `c` stands for conjugate transpose and for real matrices it is equal to `t`.

fabiomestre: My understanding from the documentation is that 'c' will behave the same as 't' on  real matrices. So, even if we don't support complex yet, the API should still return correct results when using 'c' with real numbers.

https://spec.oneapi.io/versions/latest/elements/oneMKL/source/architecture/architecture.html#onemkl-datatypes

include/interface/blas2_interface.h
pgorlani: I have no preference. I followed the style of `gemv`.

Let me know which is the one that needs to be used.

test/unittest/blas2/blas2_gbmv_test.cpp
pgorlani: The documentation about this was changed recently by https://github.com/oneapi-src/oneAPI-spec/pull/420

```diff
- -  Perform Hermitian transpose (transpose and conjugate). Only applicable to complex matrices.
+ -  Perform Hermitian transpose (transpose and conjugate). Is the same as ``transpose::trans`` for real matrices.
```

Rbiessy (approved):
LGTM, I only have minor comments so approving now.

test/unittest/blas2/blas2_gbmv_test.cpp
```diff
@@ -0,0 +1,131 @@
...
+  std::vector<scalar_t> x_v(x_size, 10.0);
+  // output Vector
+  std::vector<scalar_t> y_v_gpu_result(y_size, scalar_t(10.0));
+  // output system vector
+  std::vector<scalar_t> y_v_cpu(y_size, scalar_t(10.0));
+
+  fill_random(a_m);
+  fill_random(x_v);
+
+  // SYSTEM GEMMV
```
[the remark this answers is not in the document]
pgorlani: Yes, I wrote these test by modifying `test/unittest/blas2/blas2_gemv_test.cpp`. This typo was present there.

src/operations/blas2/gemv.hpp
```diff
@@ -921,5 +921,85 @@ SYCL_BLAS_INLINE void GemvCol<Lower, Diag, Upper, Unit, lhs_t, matrix_t,
   vector_.adjust_access_displacement();
 }
 
+/**** GBMV ****/
```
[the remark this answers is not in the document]
pgorlani: Thanks for your review, @Rbiessy!

Yes, good idea! I did it in the last commit.

test/unittest/blas2/blas2_gbmv_test.cpp
fabiomestre: I think we should try to match what the most recent oneAPI spec says.  And my understanding of the current code is that c should be working already for real numbers (the code defaults to the 't' case). So it's just a matter of updating the test.

If you think it's more work than that, then we can skip it.

include/interface/blas2_interface.h
pgorlani: The last commit applies the doxygen format.

test/unittest/blas2/blas2_gbmv_test.cpp
pgorlani: I confirm that `gbmv` acts as expected when invoked with `c` and real numbers.  For the moment, I suggest maintaining the `gbmv` test consistent with the others testing just `t` and `n`.

fabiomestre (approved):

include/operations/blas2_trees.h
```diff
@@ -227,6 +227,45 @@ make_Gemv_Row(lhs_t &lhs_, matrix_t &matrix_, vector_t &vector_,
...
+  value_t eval(sharedT shrMem, cl::sycl::nd_item<1> ndItem);
+  void bind(cl::sycl::handler &h);
+  void adjust_access_displacement();
+};
+/*!
+ @brief Generator/factory for GBMV trees.
+ */
+template <uint32_t local_range, bool is_transposed, typename lhs_t,
+          typename matrix_t, typename vector_t>
+Gbmv<lhs_t, matrix_t, vector_t, local_range, is_transposed> make_Gbmv(
```
mehdi-goli, line 262: Why do you need make_Gbmv? you don't need that function as this function returns gbmv. Also, this GBMV class called internally so no need for this function

Rbiessy: I think it makes sense to keep this helper function. The advantage is that it can automatically deduce the template parameters `lhs_t`, `matrix_t` and `vector_t`. It's used quite a lot for `makeGemv` for instance but you're right that it's used internally only.

mehdi-goli: 👍

PR #364, opened 2022-12-19
Remove unnecessary dependencies for header only

Rbiessy:
SYCL compiler and Python dependencies are not needed in CMake when installing in header only mode.


hjabird (approved):
Looks good to me.

mehdi-goli (approved):

KumudhaN (approved):

PR #366, opened 2022-12-23
Remove dependency to ComputeCpp-SDK module

Rbiessy:
* FindComputeCpp.cmake is now copied from the SDK to SYCL-BLAS modules
* Remove remaining occurences to ComputeCpp-SDK include folder
* Set the VERSION property only for the shared object library


mehdi-goli (approved):

hjabird (approved):
Looks good to me.

Issue #368, opened 2023-01-02
TRSM should assume unit diagonal when diag arg indicates this


hjabird:
# Description

The `trsm` (solve triangular system of equations) takes the argument `diag` which can be set to indicate that the argument matrix is unit diagonal or not unit diagonal. Unit diagonal implies that the values on the diagonal of the matrix are all = 1.

When `diag` is set to unit diagonal, that this is true should be assumed ([LAPACK docs](https://netlib.org/lapack/explore-html/db/dc9/group__single__blas__level3_ga9893cceb3ffc7ce400eee405970191b3.html), [Intel docs](https://www.intel.com/content/www/us/en/develop/documentation/onemkl-developer-reference-fortran/top/blas-and-sparse-blas-routines/blas-routines/blas-level-3-routines/trsm.html)). Instead, the `trsm` routine depends on the actual value on the diagonal.

# Reproducing this issue

In the TRSM unit test [here](https://github.com/codeplaysoftware/sycl-blas/blob/7b5bfb4285d06a24c295c5216b86efd350c963be/test/unittest/blas3/blas3_trsm_test.cpp#L57):
```c++
const scalar_t diagValue =
      diag == 'u' ? scalar_t{1} : random_scalar(scalar_t{1}, scalar_t{10});
```
can be set to 
```c++
const scalar_t diagValue =
      diag == 'u' ? scalar_t{0.4} : random_scalar(scalar_t{1}, scalar_t{10});
```

The reference implementation ignores the 0.4 on the diagonal. SYCL-BLAS does not.

PR #369, opened 2023-01-03
[trsm] Assume 1 on diagonal when unit diagonal is set

hjabird:
Fixes: https://github.com/codeplaysoftware/sycl-blas/issues/368

* Takes values on the diagonal to equal 1, even if this isn't actually the case in argument matrix.
* Adjusts tests:
  * To have non-1 values on diagonal, even if unit diagonal.
  * To construct similar matrices in for both upper and lower triangular.
  * To construct trsm arg matrices with for unit diag.

Test locally on Intel iGPU with single precision compiled with DPC++2022-10-06, and using the open-source OneMKL library's test suite compiled with DPC++2022-06-25.

### DIFF
--- a/src/interface/trsm_interface.hpp
+++ b/src/interface/trsm_interface.hpp
@@ -41,7 +41,8 @@ namespace internal {
  * @param uplo Indicates if A is lower or upper triangular
  * @param trans Indicates the form that the matrix A will take in the
  * multiplication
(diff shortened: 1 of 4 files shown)


KumudhaN (approved):
LGTM

fabiomestre (approved):

PR #371, opened 2023-01-04
Test synchronous api for relevant operations

Rbiessy:
* Add missing waits for synchronous operations.
* Change iamax and iamin tests to only verify the index value as the synchronous version and OneMKL only return the index.

### DIFF
--- a/test/unittest/blas1/blas1_iamax_test.cpp
+++ b/test/unittest/blas1/blas1_iamax_test.cpp
@@ -6,10 +6,11 @@ template <typename scalar_t>
 void run_test(const combination_t<scalar_t> combi) {
   using tuple_t = IndexValueTuple<int, scalar_t>;
 
+  api_type api;
   index_t size;
   index_t incX;
   generation_mode_t mode;
-  std::tie(size, incX, mode) = combi;
+  std::tie(api, size, incX, mode) = combi;
 
   // Input vector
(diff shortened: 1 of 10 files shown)


test/unittest/blas1/blas1_iamax_test.cpp
```diff
@@ -31,24 +32,20 @@ void run_test(const combination_t<scalar_t> combi) {
...
-  auto event =
-      blas::helper::copy_to_host(sb_handle.get_queue(), gpu_out_s, &out_s, 1);
-  sb_handle.wait(event);
-
-  using data_tuple_t = IndexValueTuple<int, scalar_t>;
-  data_tuple_t out_data_s{out_s.ind, static_cast<scalar_t>(out_s.val)};
+  auto gpu_x_v = blas::make_sycl_iterator_buffer<scalar_t>(x_v, size * incX);
+
+  if (api == api_type::async) {
+    auto gpu_out_s = blas::make_sycl_iterator_buffer<tuple_t>(&out_s, 1);
```
fabiomestre, line 38: I'm not entirely sure why this function returns a tuple on the async version and a float on the sync version. Seems inconsistent. Regardless, we should probably still assert on out_s.data for the async version.

Same comment would apply for iamin

fabiomestre (approved):

mehdi-goli (approved):

PR #375, opened 2023-01-20
Added support for Level 2 spr operator

muhammad-tanvir-1211:
This PR targets the level 2 spr operator. All the tests are working locally and the benchmarks show that this implementation performs better than it's `clblast` counterpart on the `Intel(R) UHD Graphics 750 [0x4c8a]` device.


src/operations/blas2/gerp.hpp
```diff
@@ -0,0 +1,196 @@
...
+          typename rhs_1_t, typename rhs_2_t>
+template <int N>
+struct Gerp<Single, isColMajor, isUpper, lhs_t, rhs_1_t,
+            rhs_2_t>::compute_row_col<N, false, false> {
+  using index_t =
+      Gerp<Single, isColMajor, isUpper, lhs_t, rhs_1_t, rhs_2_t>::index_t;
+  using value_t =
+      Gerp<Single, isColMajor, isUpper, lhs_t, rhs_1_t, rhs_2_t>::value_t;
+  void operator()(const index_t id, const index_t, index_t& row, index_t& col) {
+    row = (-1 + cl::sycl::sqrt((value_t)(1 + 8 * id))) / 2;
```
hjabird, line 48: I don't like the use of a floating-point operation here. This comment applies everywhere you've used SYCL sqrt.

Maybe I've misunderstood quite how this works, but, if `value_t` is `float`, that means that we can have a roughly 1 million element matrix (1000 * 1000) before float is no longer able to represent (1 + 8 * 10000000) exactly. Then sqrt can have 2 ULP error, plus division of floating point can add more error. 

It looks like you test `n=1010` later on, so it might be that this is fine.

test/unittest/blas2/blas2_spr_test.cpp
```diff
@@ -0,0 +1,113 @@
...
+ *  SYCL-BLAS: BLAS implementation using SYCL
+ *
+ *  @filename blas2_spr_test.cpp
+ *
+ **************************************************************************/
+
+#include "blas_test.hpp"
+
+template <typename scalar_t>
+using combination_t = std::tuple<char, char, int, scalar_t, int>;
```
hjabird, line 29: `index_t`?

hjabird (commented):
Overall good. Mostly interested in how you compute indexes with `sqrt`.

test/unittest/blas2/blas2_spr_test.cpp
muhammad-tanvir-1211: Fixed this.

include/operations/blas2_trees.h
```diff
@@ -424,5 +424,39 @@ GerCol<Single, Lower, Diag, Upper, lhs_t, rhs_1_t, rhs_2_t> make_ger_col(
       lhs_, scalar_, rhs_1_, rhs_2_, nWG_row_, nWG_col_, local_memory_size_);
 }
 
+/**** GERP N COLS x (N + 1)/2 ROWS FOR PACKED MATRIX ****/
+template <bool Single, bool isColMajor, bool isUpper, typename lhs_t,
+          typename rhs_1_t, typename rhs_2_t>
+struct Gerp {
```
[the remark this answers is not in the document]
muhammad-tanvir-1211: Added more explanation in the comments.

common/include/common/common_utils.hpp
```diff
@@ -52,6 +52,9 @@ template <typename scalar_t>
 using sbmv_param_t =
     std::tuple<std::string, index_t, index_t, scalar_t, scalar_t>;
 
+template <typename scalar_t>
+using spr_param_t = std::tuple<std::string, int, scalar_t, int>;
```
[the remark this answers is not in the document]
muhammad-tanvir-1211: Fixed this.

src/operations/blas2/gerp.hpp
Rbiessy, line 48: Out of curiosity can you explain where is this 8 coming from?

Rbiessy (approved):

src/operations/blas2/gerp.hpp
muhammad-tanvir-1211: So this `8` comes up when we apply the quadratic formula to solve the equation for `row` or `col` index calculation. An example of this is shown in the image below.
image

Rbiessy: Thank you!

hjabird: Can you comment on whether there is a risk that `id` will be big enough for there to be issues with using floating point calculations - what about with half precision?

muhammad-tanvir-1211: Hi @hjabird, this issue is now fixed. I have tried running the benchmarks for upto `16384x16384` matrix size and the square roots are looking good (almost).

include/interface/blas2_interface.h
```diff
@@ -190,6 +190,32 @@ typename sb_handle_t::event_t _syr(
     index_t _lda             // >max(1, _N) The first dimension of _mA
 );
 
+/*!
```
pgorlani, line 193: Thanks for your patch, @muhammad-tanvir-1211!

Probably, it would be better to use the doxygen format for documenting the function arguments, as discussed [here](https://github.com/codeplaysoftware/sycl-blas/pull/362#discussion_r1049516404).

include/interface/blas2_interface.h
```diff
@@ -190,6 +190,32 @@ typename sb_handle_t::event_t _syr(
...
+ matrix.
+
+ Generalised vector squaring followed by a sum with a packed symmetric matrix,
+ i.e. computing the mathematical operation:
+
+ A = alpha*x*xT + A
+
+ See the netlib blas interface documentation for more details of the high level
+ interface:
+ https://netlib.org/lapack/explore-html/d6/d30/group__single__blas__level2_ga13dca9a765471c68248e8e00190f4d4d.html#ga13dca9a765471c68248e8e00190f4d4d
```
pgorlani, line 204:
```suggestion
https://netlib.org/lapack/explore-html/dd/dba/dspr_8f.html
```
For consistency with the other functions.

include/interface/blas2_interface.h
```diff
@@ -515,6 +541,36 @@ typename sb_handle_t::event_t inline _syr(
...
```
pgorlani, line 555:
```suggestion
https://netlib.org/lapack/explore-html/dd/dba/dspr_8f.html
```
For consistency with the other functions.

src/interface/blas2_interface.hpp
```diff
@@ -605,6 +605,59 @@ typename sb_handle_t::event_t _syr_impl(
...
+  auto vx = make_vector_view(_vx, std::abs(_incx), _N);
+
+  const index_t localSize = sb_handle.get_work_group_size();
+  const index_t nColsWG = localSize;
+
+  const index_t nWGPerCol = (_N * (_N + 1) / 2 - 1) / nColsWG + 1;
+  const index_t globalSize = localSize * nWGPerCol;
+
+  if (Upper) {
+    auto assignOp = make_gerp<true, true>(mA, _N, _alpha, vx, _incx, vx, _incx);
```
pgorlani, line 650:
```suggestion
    auto spr = make_gerp<true, true>(mA, _N, _alpha, vx, _incx, vx, _incx);
```

src/interface/blas2_interface.hpp
```diff
@@ -605,6 +605,59 @@ typename sb_handle_t::event_t _syr_impl(
...
+    return ret = concatenate_vectors(
+               ret, sb_handle.execute(assignOp, localSize, globalSize));
```
pgorlani, line 652:
```suggestion
               ret, sb_handle.execute(spr, localSize, globalSize));
```

src/interface/blas2_interface.hpp
```diff
@@ -605,6 +605,59 @@ typename sb_handle_t::event_t _syr_impl(
...
+  } else {
+    auto assignOp =
```
pgorlani, line 654:
```suggestion
    auto spr =
```

src/interface/blas2_interface.hpp
```diff
@@ -605,6 +605,59 @@ typename sb_handle_t::event_t _syr_impl(
...
+        make_gerp<true, false>(mA, _N, _alpha, vx, _incx, vx, _incx);
+    return ret = concatenate_vectors(
+               ret, sb_handle.execute(assignOp, localSize, globalSize));
```
pgorlani, line 657:
```suggestion
               ret, sb_handle.execute(spr, localSize, globalSize));
```

src/interface/blas2_interface.hpp
```diff
@@ -605,6 +605,58 @@ typename sb_handle_t::event_t _syr_impl(
...
+  if (_alpha == (element_t)0) {
+    std::vector<cl::sycl::event> event;
+    return event;
+  }
+
+  typename sb_handle_t::event_t ret;
+  _Uplo = tolower(_Uplo);
+  const int Upper = _Uplo == 'u';
+  auto mA = make_matrix_view<col_major>(_mPA, _N, (_N + 1) / 2, _N);
+  auto vx = make_vector_view(_vx, std::abs(_incx), _N);
```
pgorlani, line 641: Do we really need the `abs` value of `_incx`?

muhammad-tanvir-1211: This `std::abs` was required because I was calculating the final indices myself inside the kernel rather than using the vector view. I have updated the kernel to use the vector and matrix view `eval` functions instead and removed this `std::abs`.

include/operations/blas2_trees.h
```diff
@@ -424,5 +424,44 @@ GerCol<Single, Lower, Diag, Upper, lhs_t, rhs_1_t, rhs_2_t> make_ger_col(
       lhs_, scalar_, rhs_1_, rhs_2_, nWG_row_, nWG_col_, local_memory_size_);
 }
 
+/**** GERP N COLS x (N + 1)/2 ROWS FOR PACKED MATRIX ****/
+/* This is a specialization of the GER class for the packed
+ * symmetric matrices (P stands for Packed in the name). For more details
+ * on matrix layouts, refer to the explanation here:
+ * https://spec.oneapi.io/versions/1.1-rev-1/elements/oneMKL/source/domains/matrix-storage.html#matrix-storage
+ */
+template <bool Single, bool isUpper, typename lhs_t,
```
pgorlani, line 433: What is it the purpose of `Single` template parameter?

include/operations/blas2_trees.h
```diff
@@ -424,5 +424,44 @@ GerCol<Single, Lower, Diag, Upper, lhs_t, rhs_1_t, rhs_2_t> make_ger_col(
...
+  lhs_t lhs_;
+  rhs_1_t rhs_1_;
+  rhs_2_t rhs_2_;
+  value_t alpha_;
+  index_t N_, incX_1_, incX_2_;
+
+  Gerp(lhs_t &_l, index_t N_, value_t _alpha, rhs_1_t &_r1, index_t _incX_1,
+       rhs_2_t &_r2, index_t _incX_2);
+  index_t get_size() const;
+  template <int N, bool Upper>
```
pgorlani, line 448: What is the purpose of `N` template parameter?

muhammad-tanvir-1211: I added the `N` template parameter to this `compute_row_col` struct to be able to partially specialize it in the code. I wasn't able to fully specialize this struct as I was getting compile time errors when trying to do so.

include/operations/blas2_trees.h
muhammad-tanvir-1211: The `Single` template parameter is just for distinguishing between `spr` and `spr2` operators. Once we implement the `spr2` operator, we would be able to use the same `Gerp` kernel and use this `Single` template parameter to change the way we perform the computation.

include/operations/blas2_trees.h
muhammad-tanvir-1211: I have removed the `compute_row_col` struct and replaced it with a static function of the same name.

include/operations/blas2_trees.h
pgorlani: Thanks, I understand. 

Maybe, we could rename it to `isRank1` or similar.

pgorlani (approved):

OuadiElfarouki (approved):

PR #378, opened 2023-01-31
Disable -ffast-math in tests

pgorlani:
This patch disables the fast math optimizations for the test compilation. These optimizations cause test failures in `blas1_rotmg_test` and `blas1_rotg_test`.

Fast math optimizations are applied by default by `Intel(R) oneAPI DPC++/C++ Compiler 2023.0.0` binary release.


muhammad-tanvir-1211 (approved):

Rbiessy (approved):

pgorlani:
Thanks to everybody for your reviews!

> LGTM. Just one minor question, would it be possible to apply this flag only on the rotmg test and not others. It would be good to monitor this flag for other tests.

@mehdi-goli Yes, why not!

I implemented it in the last commit. Let me know if it is fine.

PR #379, opened 2023-02-07
Fixed a GEMM Config causing drop in performance with DPCPP on intel GPU

OuadiElfarouki:
This small patch rectifies the tiling parameters of a tuned GEMM configuration for intel GPUs that was causing a considerable drop in performance _(2 orders of magnitude)_ when used with DPC++ _(observed using the default benchmarking setup, mainly visible on high `mnk` values as they end up picking that particular GEMM kernel)_.


muhammad-tanvir-1211:
@Rbiessy FYI: it seems that the performance drop on the Intel GPU was because of a misplaced GEMM config in the `src/interface/blas3/backend/intel_gpu.hpp` file. Although this patch is able to fix the performance drop for the two compilers, we still need to investigate further as to why the same Gemm config is unable to perform at the same level across different compilers.

muhammad-tanvir-1211 (approved):

Rbiessy:
Thanks for letting me know. I'd be curious to know this too if you ever find out!

pgorlani (approved):

PR #382, opened 2023-02-24
Add Level1 cuBLAS benchmarks implementation

s-Nick:
This PR adds the implementation of benchmarks for all the level1 operators available in *sycl-BLAS* and *cuBLAS*.
Update CMakelists to provide a convenient way to compile benchmarks only if desired.


benchmark/cublas/CMakeLists.txt
```diff
@@ -0,0 +1,51 @@
...
+  blas1/rotm.cpp
+  blas1/rotmg.cpp
+  blas1/scal.cpp
+  #blas1/sdsdot.cpp -> not implemented in cuBLAS
+) 
+
+#if(${BLAS_ENABLE_EXTENSIONS})
+#  list(APPEND sources "extension/reduction.cpp")
+#endif()
+
```
pgorlani, line 26: Thanks for your patch, @s-Nick!

```suggestion
```

benchmark/cublas/main.cpp
```diff
@@ -0,0 +1,38 @@
...
+int main(int argc, char** argv) {
+  // Read the command-line arguments
+  auto args = blas_benchmark::utils::parse_args(argc, argv);
+
+  // Initialize googlebench
+  benchmark::Initialize(&argc, argv);
+
+  // Create a sycl blas sb_handle from the queue
+  cublasHandle_t cublas_handle = NULL;
+  cublasCreate(&cublas_handle);
```
pgorlani, line 23:
```suggestion
  CUBLAS_CHECK(cublasCreate(&cublas_handle));
```

I would be good idea to check for errors the return values of `cublas` and `cuda` function calls.

benchmark/cublas/blas1/asum.cpp
```diff
@@ -0,0 +1,156 @@
+/***************************************************************************
+ *
+ *  @license
+ *  Copyright (C) 2016 Codeplay Software Limited
```
muhammad-tanvir-1211, line 4: Could you please remove the year from the license in all of the files? We had recently decided to remove the year from all of the licenses in SYCL-BLAS/DNN but I guess there are still files which have these years.

benchmark/cublas/main.cpp
```diff
@@ -0,0 +1,38 @@
+#include "cublas_v2.h"
+#include "cuda_runtime_api.h"
+#include "utils.hpp"
+#include <common/cli_device_selector.hpp>
+#include <common/print_queue_information.hpp>
+
+// Create a shared pointer to a sycl blas Handle, so that we don't keep
```
muhammad-tanvir-1211, line 7: This should be `cublas` handle instead of `sycl-blas` handle.

benchmark/cublas/main.cpp
s-Nick: I am using them when I write the benchmark to understand if everything works fine, then I remove them. Should I keep them all?

pgorlani: Yes, it is advisable to keep them in order to be aware of possible runtime errors, e.g., `cublas_handler` creation problems due to system configurations, memory allocation errors, and so on.

benchmark/cublas/blas1/asum.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+#ifdef BLAS_VERIFY_BENCHMARK
+  // Run a first time with a verification of the results
+  scalar_t vr_ref = reference_blas::asum(size, v1.data(), 1);
+  scalar_t vr_temp = 0;
+  {
+    scalar_t* vr_temp_gpu = nullptr;
+    cudaMalloc(&vr_temp_gpu, sizeof(scalar_t));
+    cudaDeviceSynchronize();
+    cublasSetPointerMode(cuda_handle, CUBLAS_POINTER_MODE_DEVICE);
```
pgorlani, line 80: I suggest moving `cublasSetPointerMode` outside `#ifdef BLAS_VERIFY_BENCHMARK` (eventually in `main.cpp`) since we would like to have it in any case.

benchmark/cublas/utils.hpp
```diff
@@ -0,0 +1,59 @@
...
+  cudaEventElapsedTime(&elapsed_time, cuda_events[0], cuda_events[1]);
+  // convert result from ms to ns
+  return static_cast<double>(elapsed_time) * 1'000'000.;
+}
+
+}  // namespace utils
+}  // namespace blas_benchmark
+
+#define CUDA_CHECK(err)                                                  \
+  do {                                                                   \
```
pgorlani, line 42: Is it necessary to wrap them in a `do {} while(0) ` loop?

Can `{}` be enough for fitting the purpose?

benchmark/cublas/blas1/iamax.cpp
```diff
@@ -0,0 +1,169 @@
...
+  CUDA_CHECK(cudaMalloc(&d_inx, size * sizeof(scalar_t)));
+  CUDA_CHECK(cudaMemcpy(d_inx, v1.data(), size * sizeof(scalar_t),
+                        cudaMemcpyHostToDevice));
+
+  index_t* d_out = nullptr;
+  CUDA_CHECK(cudaMalloc(&d_out, sizeof(index_t)));
+  CUDA_CHECK(cudaDeviceSynchronize());
+
+  CUBLAS_CHECK(
+      cublasSetPointerMode_v2(cuda_handle, CUBLAS_POINTER_MODE_DEVICE));
```
pgorlani, line 78: Do you really need `cublasSetPointerMode_v2` in place of `cublasSetPointerMode`?

s-Nick: yes it is. Almost every cublas call in reality is a ```define``` like this 
```
#define cublas_operator cublas_operator_v2
```
My lsp doens't give me hint if use the first one and, like in this case, I forgot to remove the _v2. I should fix it for consistency...

benchmark/cublas/CMakeLists.txt
s-Nick: I let it commented because we may need it in the near future. Should I remove it?

benchmark/cublas/blas1/axpy.cpp
```diff
@@ -0,0 +1,159 @@
...
+}
+
+template <typename scalar_t>
+void run(benchmark::State& state, cublasHandle_t* cuda_handle_ptr, index_t size,
+         bool* success) {
+  // Google-benchmark counters are double.
+  double size_d = static_cast<double>(size);
+  state.counters["size"] = size_d;
+  state.counters["n_fl_ops"] = 2.0 * size_d;
+  state.counters["bytes_processed"] = 3.0 * size_d * sizeof(scalar_t);
```
mehdi-goli, line 52: I can see that these numbers are setting probably for all AXPY benchmarks. Similarly, for other BLAS routine benchmarks these numbers are the same. Can we put per routine config in the benchmark utils so that we can call them and don't repeat them every time we are writing the benchmark? In that case, it would be less error prune as well.

benchmark/cublas/CMakeLists.txt
pgorlani: I don't think there are cuBlas extension we want to benchmark.

benchmark/cublas/blas1/axpy.cpp
pgorlani: If these counters need to record the number of floating-point operations and memory transactions performed by a routine, they cannot be the same for all BLAS1 routines: e.g., for AXPY `"bytes_processed"` is equal to 3 times the array size, for IAMAX is 1 time the array size (+1 int, if we want to consider the output).

benchmark/cublas/blas1/asum.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+  CUBLAS_CHECK(cublasSetPointerMode(cuda_handle, CUBLAS_POINTER_MODE_DEVICE));
+#ifdef BLAS_VERIFY_BENCHMARK
+  // Run a first time with a verification of the results
+  scalar_t vr_ref = reference_blas::asum(size, v1.data(), 1);
+  scalar_t vr_temp = 0;
+  {
+    // Temp result on device copied back upon destruction
+    blas_benchmark::utils::CUDAScalar<scalar_t, true> vr_temp_gpu(vr_temp);
+    CUDA_CHECK(cudaDeviceSynchronize());
```
muhammad-tanvir-1211, line 75: Delete the synchronization here, on line 77. Move synchronization from 91 to 109.

common/include/common/common_utils.hpp
```diff
@@ -736,9 +742,19 @@ static inline double time_event(event_t&);
 template <typename event_t>
 static inline double time_events(std::vector<event_t> es) {
   double total_time = 0;
+#if defined(BUILD_CUBLAS_BENCHMARKS)
```
muhammad-tanvir-1211, line 745: Remove this timing here and create a cuda specific `time_events` function inside `cublas/utils.hpp`

benchmark/cublas/blas1/rotmg.cpp
```diff
@@ -0,0 +1,168 @@
...
+    err_stream << "Value mismatch." << std::endl;
+    const std::string& err_str = err_stream.str();
+    state.SkipWithError(err_str.c_str());
+    *success = false;
+  };
+#endif
+
+  auto blas_warmup = [&]() -> void {
+    cublas_routine<scalar_t>(cuda_handle, d_d1, d_d2, d_x1, d_y1, d_param);
+    CUDA_CHECK(cudaDeviceSynchronize());
```
muhammad-tanvir-1211, line 114: Replace this with `cudaStreamSynchronize` and see if it works or not. If yes, remove all device sync with stream sync.

OuadiElfarouki (approved):

common/include/common/common_utils.hpp
```diff
@@ -14,6 +14,7 @@
 #include <sstream>
 #include <stdexcept>
 #include <string>
+#include <type_traits>
```
OuadiElfarouki, line 17: Is this still needed by the way ?

s-Nick: no, removing it in next commit. Thanks for spotting it!

benchmark/cublas/main.cpp
```diff
@@ -0,0 +1,35 @@
+#include "cublas_v2.h"
```
muhammad-tanvir-1211, line 1: Could you please add the license at the beginning of this file, the `CMakeLists.txt` file and the `utils.hpp` file?

muhammad-tanvir-1211 (approved):

PR #383, opened 2023-03-07
Added RrocBLAS benchmarks for Level 1 operators

OuadiElfarouki:
This PR adds BLAS benchmarks targetting AMD GPUs using rocBLAS for all the Level1 operators available in both sycl-BLAS and rocBLAS.


benchmark/rocblas/blas1/asum.cpp
```diff
@@ -0,0 +1,148 @@
...
+      rocblas_asum_f<scalar_t>(rb_handle, size, d_v1, 1, d_vr);
+      CHECK_HIP_ERROR(hipStreamSynchronize(NULL));
+      return;
+    };
+
+    // Create a utility lambda describing the blas method that we want to run.
+    auto blas_method_def = [&]() -> std::vector<hipEvent_t> {
+      hipEvent_t start, stop;
+      CHECK_HIP_ERROR(hipEventCreate(&start));
+      CHECK_HIP_ERROR(hipEventCreate(&stop));
```
s-Nick, line 97: In cuBLAS create events inside the lambda gave host RAM memory problem not visible in level1 opertors but it is a concern for level2 operators . Did you check host RAM memory usage?

OuadiElfarouki: Good point, will check if it's causing a memory issue. thanks!

benchmark/rocblas/utils.hpp
```diff
@@ -0,0 +1,206 @@
...
+ * tuple of double)
+ */
+template <typename function_t, typename... args_t>
+static inline std::tuple<double, double> timef_hip(function_t func,
+                                                   args_t&&... args) {
+  auto start = std::chrono::system_clock::now();
+  std::vector<hipEvent_t> events = func(std::forward<args_t>(args)...);
+  auto end = std::chrono::system_clock::now();
+
+  double overall_time = (end - start).count();
```
pgorlani, line 194: Thank you so much for your patch, @OuadiElfarouki!

Are we sure that the `overall_time` will be expressed in nanoseconds in any case? Shall we use `std::chrono::nanoseconds` to be on the safe side?

benchmark/rocblas/blas1/scal.cpp
```diff
@@ -0,0 +1,145 @@
...
+  state.counters["bytes_processed"] = 2 * size_d * sizeof(scalar_t);
+
+  // Create data
+  std::vector<scalar_t> v1 = blas_benchmark::utils::random_data<scalar_t>(size);
+  scalar_t alpha = blas_benchmark::utils::random_scalar<scalar_t>();
+
+  {
+    // Device memory allocation & H2D Copy
+    blas_benchmark::utils::HIPVector<scalar_t> d_v1(size, v1.data());
+    blas_benchmark::utils::HIPScalar<scalar_t> d_alpha(alpha);
```
pgorlani, line 62: Good idea! 

@s-Nick Is it possible to replicate it also for the cuBLAS patch?

s-Nick: I think so, does it make the code more readable? 
I like it, it hides the hip (or cuda) part of the code. If also @muhammad-tanvir-1211 likes it, I can try to do it and update the cublas PR.

benchmark/rocblas/utils.hpp
```diff
@@ -0,0 +1,206 @@
...
+    if (hPtr != nullptr) {
+      CHECK_HIP_ERROR(hipMemcpy(hPtr, dPtr, bytes_, hipMemcpyDeviceToHost));
+    }
+    return;
+  }
+
+  // Free device memory
+  void free(T* d) {
+    if (d != nullptr) {
+      CHECK_HIP_ERROR((hipFree)(d));
```
s-Nick, line 104: I don't know much of hipAPI but this ```((hipFree)(d))``` looks different from other hipAPI calls, is it correct?

benchmark/rocblas/blas1/scal.cpp
pgorlani: > I think so, does it make the code more readable?
 Yes, I think so. But, you may also want to know @muhammad-tanvir-1211 and @mehdi-goli opinions.

muhammad-tanvir-1211: I think it will definitely be a good addition to the cublas benchmarks. The code remains clean and everything gets handled in the backend.

benchmark/rocblas/utils.hpp
OuadiElfarouki: Thanks @pgorlani. Actually doc says that `nanoseconds` is the highest resolution and is thus always used by default. All `overall_time` computations in SYCL-BLAS are assuming this as well.

benchmark/rocblas/blas1/dot.cpp
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    blas_benchmark::utils::calc_avg_counters(state);
+  }
+};
+
+template <typename scalar_t>
+void register_benchmark(blas_benchmark::Args& args, rocblas_handle& rb_handle,
+                        bool* success) {
+  auto blas1_params = blas_benchmark::utils::get_blas1_params(args);
```
muhammad-tanvir-1211, line 128: Could you please change this to the `operator` name here e.g. `dot_params` in all the files? This would make the names consistent across all the benchmarks that we have.

benchmark/rocblas/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
+if(BLAS_VERIFY_BENCHMARK)
```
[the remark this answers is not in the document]
OuadiElfarouki: Thanks! I'll go over all license / formatting to match the recent updates !

muhammad-tanvir-1211 (approved):

benchmark/rocblas/blas1/asum.cpp
```diff
@@ -0,0 +1,145 @@
...
+      err_stream << "Value mismatch: " << vr_temp << "; expected " << vr_ref;
+      const std::string& err_str = err_stream.str();
+      state.SkipWithError(err_str.c_str());
+      *success = false;
+    };
+#endif
+
+    auto blas_warmup = [&]() -> void {
+      rocblas_asum_f<scalar_t>(rb_handle, size, d_v1, 1, d_vr);
+      CHECK_HIP_ERROR(hipStreamSynchronize(NULL));
```
s-Nick, line 91: in cublas we moved this `streamSynchronize` after the `warmup`. Could you do the same?

benchmark/rocblas/blas1/asum.cpp
```diff
@@ -0,0 +1,145 @@
...
+    // Measure
+    for (auto _ : state) {
+      std::tuple<double, double> times =
+          blas_benchmark::utils::timef_hip(blas_method_def);
+
+      // Report
+      blas_benchmark::utils::update_counters(state, times);
+    }
+
+    blas_benchmark::utils::calc_avg_counters(state);
```
s-Nick, line 121: I see you usually don't destroy the `hipEvent_t` is it ok not to do it?

benchmark/rocblas/blas1/asum.cpp
OuadiElfarouki: Sure will do!

benchmark/rocblas/blas1/asum.cpp
OuadiElfarouki: Well spotted, it should be destroyed indeed!

s-Nick (approved):

PR #385, opened 2023-03-10
Adding USM support for Level 1 operators in SYCL-BLAS

muhammad-tanvir-1211:
This PR adds USM support in SYCL-BLAS. Currently, only Level 1 operators are supported. All Level 1 benchmarks and tests now use USM and Buffers.


src/views/view_sycl_usm.hpp
```diff
@@ -0,0 +1,292 @@
+/***************************************************************************
```
muhammad-tanvir-1211, line 1: Remove this file and use `view.hpp`

benchmark/syclblas/blas1/axpy.cpp
```diff
@@ -96,26 +111,35 @@ void run(benchmark::State& state, blas::SB_Handle* sb_handle_ptr, index_t size,
...
   auto axpy_params = blas_benchmark::utils::get_blas1_params(args);
 
-  for (auto size : axpy_params) {
-    auto BM_lambda = [&](benchmark::State& st, blas::SB_Handle* sb_handle_ptr,
-                         index_t size, bool* success) {
-      run<scalar_t>(st, sb_handle_ptr, size, success);
-    };
-    benchmark::RegisterBenchmark(get_name<scalar_t>(size).c_str(), BM_lambda,
-                                 sb_handle_ptr, size, success);
+#define REGISTER_MEM_TYPE_BENCHMARKS(MEMORY, NAME)                             \
```
muhammad-tanvir-1211, line 124: Remove macro, create a function that accepts `AllocType` as template parameter and then call this function twice in the loop.

include/interface/blas1_interface.h
```diff
@@ -491,55 +524,60 @@ typename sb_handle_t::event_t _dot(sb_handle_t &sb_handle, index_t _N,
...
+  return internal::_sdsdot(sb_handle, _N, sb, _vx, _incx, _vy, _incy, _rs,
+                           _dependencies);
 }
 
 /**
  * \brief ASUM Takes the sum of the absolute values
  * @param sb_handle SB_Handle
  * @param _vx BufferIterator
  * @param _incx Increment for the vector X
+ * @param _dependencies Vector of events
```
s-Nick, line 545: @muhammad-tanvir-1211  thank you for your great work!
 I think it is missing the `rs` description in the documentation tag

include/blas_meta.h
```diff
@@ -184,6 +184,9 @@ struct is_sycl_scalar
 template <>
 struct is_sycl_scalar<cl::sycl::half> : std::true_type {};
 
+template <>
+struct is_sycl_scalar<float *> : std::false_type {};
+
```
pgorlani, line 189: Thanks for your PR, @muhammad-tanvir-1211.

What about the case of doubles?

 ```suggestion
template <>
struct is_sycl_scalar<double *> : std::false_type {};
```

File: include/sycl_blas_helper.h [hunk 5]
```diff
@@ -32,6 +32,91 @@
...
+  using tuple_type = std::tuple<value_t *, cl::sycl::event>;
+};
+
+template <typename value_t>
+struct AllocHelper<value_t, AllocType::buffer> {
+  using type = blas::BufferIterator<value_t>;
+  using tuple_type = std::tuple<BufferIterator<value_t>, cl::sycl::event>;
+};
+
+#ifdef SB_ENABLE_USM
```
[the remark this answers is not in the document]
muhammad-tanvir-1211: I had to put a guard around these functions to get the ComputeCpp CI to pass.

include/blas_meta.h
muhammad-tanvir-1211: Fixed this.

include/sb_handle/sycl_blas_handle.h
```diff
@@ -127,14 +133,16 @@ class SB_Handle {
 
   inline void wait(cl::sycl::event ev) { cl::sycl::event::wait({ev}); }
 
-  /*  @brief waiting for a list of sycl events
- @param first_event  and next_events are instances of sycl::sycl::event
+  /*  @brief waiting for a list of sycl dependencies
+ @param first_event  and next_dependencies are instances of sycl::sycl::event
```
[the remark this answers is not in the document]
muhammad-tanvir-1211: Fixed this.

src/sb_handle/sycl_blas_handle.hpp
```diff
@@ -130,20 +134,40 @@ inline typename SB_Handle::event_t SB_Handle::execute(
...
   } while (_N > 1);
+
+  // Need to add this guard since the enqueue_deallocate
+  // function requires a host_task which throws a runtime
+  // exception when used with the enable_profiling{} queue
+  // property. We need to create all intermediate memory
+  // before launching the kernel to avoid running into this
+  // issue.
+  // Enabling this code only for DEFAULT_CPU backend to get the
+  // CI to pass.
```
aacostadiaz, line 159: Would it make sense to add this to all backends? What would happen if CI happen to test a different backend (or all of them)?

test/unittest/blas1/blas1_rotg_test.cpp
```diff
@@ -83,8 +96,29 @@ void run_test(const combination_t<scalar_t> combi) {
...
+  std::tie(alloc, api, a_input, b_input) = combi;
+
+  if (alloc == "usm") {  // usm alloc
+#ifdef SB_ENABLE_USM
+    run_test<scalar_t, helper::AllocType::usm>(combi);
+#endif
+  } else {  // buffer alloc
+    run_test<scalar_t, helper::AllocType::buffer>(combi);
+  }
+}
```
aacostadiaz, line 117: What do you think about moving this code to a common place (util_test of something)? I have seen this same block of code for all tests.

src/views/view_sycl.hpp
```diff
@@ -100,6 +100,14 @@ struct VectorView<
         disp_((strd > 0) ? disp : disp + (size_ - 1) * (-strd)),
         stride_(strd) {}
 
+  SYCL_BLAS_INLINE VectorView(BufferIterator<scalar_t> data, index_t disp,
+                              increment_t strd, index_t size)
+      : data_{data.template get_range_accessor<
+            cl::sycl::access::mode::read_write>()},
```
pgorlani, line 106: Basically, are we are specializing the `VectorView` for storing the accessor rather than the buffer with this PR?
So `data_` will contain an accessor that will be in any case `read_write`.

muhammad-tanvir-1211: I actually had to add this new constructor in this PR to make sure that we are able to allocate the right memory (usm or buffer) without adding anything extra in the `sycl_blas_helper.h` file. The only reason for having this constructor here is to manage any intermediate memory allocation performed by the operator. This constructor is not in any of the `tests` or `benchmarks` or `samples`, but it is required to make the code independent of the helper functions.

src/sb_handle/sycl_blas_handle.hpp
muhammad-tanvir-1211: I have tested these changes with `DEFAULT_CPU` and `INTEL_GPU` backends and both of them seem to be working fine. I am going to test the `NVIDIA_GPU` and the `AMD_GPU` backends as well to make sure there is no issue there as well.

test/unittest/blas1/blas1_rotg_test.cpp
muhammad-tanvir-1211: I couldn't move this code to a more common place since the `combi` object for each test doesn't necessarily need to be the same. That's the reason I decided to keep this code in every test separately to keep the code consistent.

s-Nick (approved):

src/views/view_sycl.hpp
muhammad-tanvir-1211: I was able to remove this constructor altogether by making the changes we discussed earlier.

include/sycl_blas_helper.h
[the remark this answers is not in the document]
pgorlani: I understand. Maybe, we could use `#ifndef __COMPUTECPP__`  for making the reason clearer.

src/sb_handle/kernel_constructor.hpp
```diff
@@ -237,6 +235,9 @@ static SYCL_BLAS_INLINE cl::sycl::event execute_tree(queue_t q_,
   cl::sycl::event ev;
   try {
     auto cg1 = [=](cl::sycl::handler &h) mutable {
+#ifdef SB_ENABLE_USM
```
pgorlani, line 238: If we are adding the event dependency also for the kernels that use buffer, it would be better to remove this pragma.

muhammad-tanvir-1211: This is just to keep the call to `depends_on` for `DPC++`. (USM and Buffers)

pgorlani: Yes, but I think we are introducing the dependencies also for ComputeCPP in case of buffer.

muhammad-tanvir-1211: Yes, but that's only at the api level. We don't use those dependencies anywhere in the code.

include/views/view.h
```diff
@@ -53,15 +53,13 @@ struct VectorView {
   using index_t = view_index_t;
   using increment_t = view_increment_t;
   using self_t = VectorView<value_t, container_t, index_t, increment_t>;
-  container_t &data_;
+  container_t data_;
```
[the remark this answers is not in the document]
pgorlani: I understand, thanks!

pgorlani (approved):
It looks good to me, but I might be missing something.

Probably a review from @mehdi-goli could be advisable.

PR #386, opened 2023-03-15
Added RocBLAS benchmarks for Level 2 operators

OuadiElfarouki:
This PR adds BLAS level 2 operators benchmarks targetting AMD GPUs through rocBLAS routines.

- gemv
- gbmv
- sbmv
- tbmv
- tbsv
- trsv
- spr
- spr2
- symv
- ger
- syr
- syr2
- tpmv
- tpsv
- trmv
- spmv

### DIFF
--- a/benchmark/rocblas/blas2/gbmv.cpp
+++ b/benchmark/rocblas/blas2/gbmv.cpp
@@ -174,6 +174,9 @@ void run(benchmark::State& state, rocblas_handle& rb_handle, int ti, index_t m,
     }
 
     blas_benchmark::utils::calc_avg_counters(state);
+
+    CHECK_HIP_ERROR(hipEventDestroy(start));
+    CHECK_HIP_ERROR(hipEventDestroy(stop));
   }  // release device memory via utils::DeviceVector destructors
 }
 
--- a/benchmark/rocblas/blas2/gbmv.cpp
+++ b/benchmark/rocblas/blas2/gbmv.cpp
@@ -146,7 +146,6 @@ void run(benchmark::State& state, rocblas_handle& rb_handle, int ti, index_t m,
     auto blas_warmup = [&]() -> void {
       rocblas_gbmv_f<scalar_t>(rb_handle, transA, m, n, kl, ku, &alpha, m_a_gpu,
                                lda, v_x_gpu, incX, &beta, v_y_gpu, incY);
-      CHECK_HIP_ERROR(hipStreamSynchronize(NULL));
       return;
     };
 
@@ -165,6 +164,7 @@ void run(benchmark::State& state, rocblas_handle& rb_handle, int ti, index_t m,
 
     // Warmup
     blas_benchmark::utils::warmup(blas_warmup);
+    CHECK_HIP_ERROR(hipStreamSynchronize(NULL));
 
     blas_benchmark::utils::init_counters(state);
 
--- a/benchmark/rocblas/blas2/gbmv.cpp
+++ b/benchmark/rocblas/blas2/gbmv.cpp
@@ -76,13 +76,12 @@ void run(benchmark::State& state, rocblas_handle& rb_handle, int ti, index_t m,
       (t_str[0] == 'n' ? n_d : m_d) * (kl_d + ku_d + 1.0) -
       0.5 * (kl_d * (kl_d + 1.0)) - 0.5 * (ku_d * (ku_d + 1.0));
 
-  {
-    double nflops_AtimesX = 2.0 * A_validVal;
-    double nflops_timesAlpha = ylen;
-    double nflops_addBetaY = (beta != scalar_t{0}) ? 2 * ylen : 0;
-    state.counters["n_fl_ops"] =
-        nflops_AtimesX + nflops_timesAlpha + nflops_addBetaY;
-  }
+  double nflops_AtimesX = 2.0 * A_validVal;
+  double nflops_timesAlpha = ylen;
+  double nflops_addBetaY = (beta != scalar_t{0}) ? 2 * ylen : 0;
+  double nflops_tot = nflops_AtimesX + nflops_timesAlpha + nflops_addBetaY;
+  state.counters["n_fl_ops"] = nflops_tot;
+
   {
     double mem_readA = A_validVal;
     double mem_readX = xlen;
@@ -177,6 +176,8 @@ void run(benchmark::State& state, rocblas_handle& rb_handle, int ti, index_t m,
       // Report
       blas_benchmark::utils::update_counters(state, times);
     }
+
+    state.SetItemsProcessed(state.iterations() * nflops_tot);
 
     blas_benchmark::utils::calc_avg_counters(state);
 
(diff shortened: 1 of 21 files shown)


benchmark/rocblas/blas2/gbmv.cpp
```diff
@@ -0,0 +1,216 @@
...
+  }
+
+  // Matrix options (rocBLAS)
+  const rocblas_operation transA =
+      t_str[0] == 'n' ? rocblas_operation_none : rocblas_operation_transpose;
+
+  // Data sizes
+  const int m_size = lda * n;
+  const int v_x_size = 1 + (xlen - 1) * incX;
+  const int v_y_size = 1 + (ylen - 1) * incY;
```
s-Nick, line 102: Thanks for your patch @OuadiElfarouki
I see `incX` and `incY` are set to 1 in this benchmark at line 59 and line 60, is this computation necessary?

benchmark/rocblas/blas2/gbmv.cpp
```diff
@@ -0,0 +1,216 @@
...
+      const std::string& err_str = err_stream.str();
+      state.SkipWithError(err_str.c_str());
+      *success = false;
+    };
+#endif
+
+    auto blas_warmup = [&]() -> void {
+      rocblas_gbmv_f<scalar_t>(rb_handle, transA, m, n, kl, ku, &alpha, m_a_gpu,
+                               lda, v_x_gpu, incX, &beta, v_y_gpu, incY);
+      CHECK_HIP_ERROR(hipStreamSynchronize(NULL));
```
s-Nick, line 149: Could you move this `hipStreamSynchronize` to line 168 as in level1 operators?

OuadiElfarouki: Yes missed that one! Thanks.

benchmark/rocblas/blas2/gbmv.cpp
OuadiElfarouki: Exactly, they're not really necessary as long as the values are set to 1. I'm just keeping it this explicit in case we parametrize the increment values in the future.

benchmark/rocblas/blas2/ger.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+    CHECK_HIP_ERROR(hipEventDestroy(start));
+    CHECK_HIP_ERROR(hipEventDestroy(stop));
+  }  // release device memory via utils::DeviceVector destructors
+}
+
+template <typename scalar_t>
+void register_benchmark(blas_benchmark::Args& args, rocblas_handle& rb_handle,
+                        bool* success) {
+  auto ger_params = blas_benchmark::utils::get_ger_params<scalar_t>(args);
```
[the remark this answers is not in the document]
s-Nick: I think we can do something similar to `trsv` at line 190-191, where the `k` param is not needed and if `k != 1` it continues the loop. This way we should not have double benchmarks.
``` C++
    {...}
    std::string ts;
    index_t m, n;
    scalar_t alpha, beta;
    std::tie(ts, m, n, alpha, beta) = p;

    // Repurpose general blas2 params
    if (ts != "n") continue;
    {...}
```
 How does it look to you?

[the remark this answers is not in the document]
OuadiElfarouki: Yes it makes sense this way. Will try it out thank you!

benchmark/rocblas/blas2/gbmv.cpp
```diff
@@ -0,0 +1,217 @@
...
+void register_benchmark(blas_benchmark::Args& args, rocblas_handle& rb_handle,
+                        bool* success) {
+  auto gbmv_params = blas_benchmark::utils::get_gbmv_params<scalar_t>(args);
+
+  for (auto p : gbmv_params) {
+    std::string ts;
+    index_t m, n, kl, ku;
+    scalar_t alpha, beta;
+    std::tie(ts, m, n, kl, ku, alpha, beta) = p;
+    int t = static_cast<int>(blas_benchmark::utils::to_transpose_enum(ts));
```
s-Nick, line 199: I did the same in `cublas` PR, so this is a question for both of us. 
`ts` is casted to `int` passed to the lambda and casted back to string in the `run` function(line 51), why don't we just pass it as a string or a char?

benchmark/rocblas/blas2/spr.cpp
```diff
@@ -0,0 +1,189 @@
...
+  state.counters["incX"] = incX;
+
+  double nflops_tot = 2.0 * size_d + n_d;
+  state.counters["n_fl_ops"] = nflops_tot;
+
+  {
+    double mem_readA = size_d;
+    double mem_readX = n_d;
+    state.counters["bytes_processed"] =
+        (mem_readA + mem_readX) * sizeof(scalar_t);
```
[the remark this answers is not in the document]
OuadiElfarouki: yes indeed. Thanks!

benchmark/rocblas/blas2/spmv.cpp
```diff
@@ -0,0 +1,204 @@
...
+  // The counters are double. We convert n to double to avoid
+  // integer overflows for n_fl_ops and bytes_processed
+  double n_d = static_cast<double>(n);
+
+  state.counters["n"] = n_d;
+
+  // Compute the number of A non-zero elements.
+  const double A_validVal = .5 * n_d * (n_d + 1);
+
+  double nflops_AtimesX = 2.0 * A_validVal;
```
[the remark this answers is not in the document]
OuadiElfarouki: yes it is a full matrix multiplication so it is correct indeed. Thanks!

benchmark/rocblas/blas2/gbmv.cpp
OuadiElfarouki: Yes indeed. As discussed, this will be subject of another PR.

benchmark/rocblas/blas2/trsv.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+  // Input matrix/vector, output vector.
+  std::vector<scalar_t> m_a(m_size);
+  std::vector<scalar_t> v_x =
+      blas_benchmark::utils::random_data<scalar_t>(v_size);
+
+  // Populate the main diagonal with larger values.
+  for (index_t i = 0; i < n; ++i)
+    for (index_t j = 0; j < n; ++j)
+      m_a[(i * lda) + i] = (i == j) ? blas_benchmark::utils::random_scalar(
```
s-Nick, line 98: `m_a[( i * lda) + i]` seems to override diagonal value once they are set. Can you check that this is what we want? 
I think that something like `m_a[(i * lda) + j]`  would update only the main diagonal.

Ouadio: Of course thanks for spotting that!

s-Nick (approved):

muhammad-tanvir-1211 (approved):

PR #387, opened 2023-03-17
Add cuBLAS benchmarks for level 2 operators

s-Nick:
This PR adds BLAS level 2 operators benchmarks using cuBLAS implementation targeting NVIDIA GPUs.

List of benchmarks added:
- gbmv
- gemv
- ger
- sbmv
- spmv
- spr
- spr2
- symv
- syr
- syr2
- tbmv
- tbsv
- tpmv
- tpsv
- trmv
- trsv

### DIFF
--- a/benchmark/cublas/blas2/gbmv.cpp
+++ b/benchmark/cublas/blas2/gbmv.cpp
@@ -106,24 +106,23 @@ void run(benchmark::State& state, cublasHandle_t* cuda_handle_ptr, int ti,
   std::vector<scalar_t> v_y =
       blas_benchmark::utils::random_data<scalar_t>(ylen);
 
-  scalar_t* alpha_d = nullptr;
-  scalar_t* beta_d = nullptr;
+  scalar_t* d_alpha = nullptr;
+  scalar_t* d_beta = nullptr;
   scalar_t* m_a_gpu = nullptr;
   scalar_t* v_x_gpu = nullptr;
   scalar_t* v_y_gpu = nullptr;
   CUDA_CHECK(cudaMalloc(&m_a_gpu, (lda*n)*sizeof(scalar_t)));
   CUDA_CHECK(cudaMalloc(&v_x_gpu, xlen*sizeof(scalar_t)));
   CUDA_CHECK(cudaMalloc(&v_y_gpu, ylen*sizeof(scalar_t)));
-  CUDA_CHECK(cudaMalloc(&alpha_d, sizeof(scalar_t)));
-  CUDA_CHECK(cudaMalloc(&beta_d, sizeof(scalar_t)));
+  CUDA_CHECK(cudaMalloc(&d_alpha, sizeof(scalar_t)));
+  CUDA_CHECK(cudaMalloc(&d_beta, sizeof(scalar_t)));
 
   CUDA_CHECK(cudaMemcpy(m_a_gpu, m_a.data(), (lda*n)*sizeof(scalar_t), cudaMemcpyHostToDevice));
   CUDA_CHECK(cudaMemcpy(v_x_gpu, v_x.data(), xlen*sizeof(scalar_t), cudaMemcpyHostToDevice));
   CUDA_CHECK(cudaMemcpy(v_y_gpu, v_y.data(), ylen*sizeof(scalar_t), cudaMemcpyHostToDevice));
-  CUDA_CHECK(cudaMemcpy(alpha_d, &alpha, sizeof(scalar_t), cudaMemcpyHostToDevice));
-  CUDA_CHECK(cudaMemcpy(beta_d, &beta, sizeof(scalar_t), cudaMemcpyHostToDevice));
-  CUDA_CHECK(cudaDeviceSynchronize());
-  cublasSetPointerMode(cuda_handle, CUBLAS_POINTER_MODE_DEVICE);
+  CUDA_CHECK(cudaMemcpy(d_alpha, &alpha, sizeof(scalar_t), cudaMemcpyHostToDevice));
+  CUDA_CHECK(cudaMemcpy(d_beta, &beta, sizeof(scalar_t), cudaMemcpyHostToDevice));
+  CUBLAS_CHECK(cublasSetPointerMode(cuda_handle, CUBLAS_POINTER_MODE_DEVICE));
 
 #ifdef BLAS_VERIFY_BENCHMARK
   // Run a first time with a verification of the results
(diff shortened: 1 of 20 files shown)


benchmark/cublas/blas2/symv.cpp
```diff
@@ -0,0 +1,196 @@
...
+  // integer overflows for n_fl_ops and bytes_processed
+  double n_d = static_cast<double>(n);
+
+  state.counters["n"] = n_d;
+
+  // Compute the number of A non-zero elements.
+  const double A_validVal = (n_d * (n_d + 1) / 2);
+
+  {
+    double nflops_AtimesX = 2 * n_d;
```
OuadiElfarouki, line 70: Did you rather mean `2*A_validVal` here? I believe it's a `2*n_d*n_d` because even-though only half the values are being **used**(read), it's still a full symmetrical matrix-vector operation... But yes for the memory accesses it's indeed half the count.

s-Nick: No, that is my mistake. I meant `2*n_d*n_d` but forgot one. Thank you for spotting it!

benchmark/cublas/blas2/gbmv.cpp
```diff
@@ -0,0 +1,202 @@
...
+  }
+  {
+    double mem_readA = A_validVal;
+    double mem_readX = xlen;
+    double mem_writeY = ylen;
+    double mem_readY = (beta != scalar_t{0}) ? ylen : 0;
+    state.counters["bytes_processed"] =
+        (mem_readA + mem_readX + mem_writeY + mem_readY) * sizeof(scalar_t);
+  }
+
```
muhammad-tanvir-1211, line 94: Can you also add this line (and line 126) in these Level 2 benchmarks? This basically reports the `items_processed_per_second` which we use to check the performance for the operator on the hardware.
https://github.com/codeplaysoftware/sycl-blas/blob/adc04a93803199669afe29dae1bdda010da9d00e/benchmark/syclblas/blas3/gemm.cpp#L133

benchmark/cublas/blas2/ger.cpp
```diff
@@ -0,0 +1,191 @@
...
+
+  // Input matrix/vector, output vector.
+  std::vector<scalar_t> m_a =
+      blas_benchmark::utils::random_data<scalar_t>(m * n);
+  std::vector<scalar_t> v_x =
+      blas_benchmark::utils::random_data<scalar_t>(xlen);
+  std::vector<scalar_t> v_y =
+      blas_benchmark::utils::random_data<scalar_t>(ylen);
+
+  cublasSetPointerMode(cuda_handle, CUBLAS_POINTER_MODE_HOST);
```
muhammad-tanvir-1211, line 89: Why do we need to set the `cublasSetPointerMode` here?

benchmark/cublas/blas2/syr2.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+  CUDA_CHECK(cudaEventDestroy(start));
+  CUDA_CHECK(cudaEventDestroy(stop));
+}
+
+template <typename scalar_t>
+void register_benchmark(blas_benchmark::Args& args,
+                        cublasHandle_t* cuda_handle_ptr, bool* success) {
+  // syr2 use the same parameters so reuse spr function
+  auto syr2_params = blas_benchmark::utils::get_syr_params<scalar_t>(args);
```
muhammad-tanvir-1211, line 164: Based on the comment above, shouldn't this be `get_spr_params`?

s-Nick: I checked, this is an old comment. Even though `spr` and `syr` share the same parameters for operators call the implementation of `get_spr_params` generates also `incX` and multiple values for `alpha` and in this case I didn't want to benchmark these changes, so I created a new function.
 I will remove the comment, but I think we should try to have a single function that generates parameters for many operators where possible instead of `gen_<operator_name>_params` for each operator.

benchmark/cublas/blas2/ger.cpp
s-Nick: We don't. Thank you for spotting it.

benchmark/cublas/blas2/syr2.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+  index_t lda = n;
+  index_t incX = 1;
+  index_t incY = 1;
+
+  // The counters are double. We convert m and n to double to avoid
+  // integer overflows for n_fl_ops and bytes_processed
+  scalar_t size_d = static_cast<scalar_t>(n * (n + 1) / 2);
+
+  state.counters["n"] = static_cast<double>(size_d);
```
OuadiElfarouki, line 60: this should be `static_cast<double>(n)` right?

benchmark/cublas/blas2/spr2.cpp
```diff
@@ -0,0 +1,185 @@
...
+void run(benchmark::State& state, cublasHandle_t* cuda_handle_ptr, char uplo,
+         int size, scalar_t alpha, int incX, bool* success) {
+  // The counters are double. We convert size to double to avoid
+  // integer overflows for n_fl_ops and bytes_processed
+  double n_d = static_cast<double>(size * (size + 1) / 2);
+
+  // Temporarely set incY to 1 for all benchmarks
+  index_t incY = 1;
+
+  state.counters["n_d"] = n_d;
```
OuadiElfarouki, line 56: state.counters["n"]

benchmark/cublas/blas2/spr2.cpp
```diff
@@ -0,0 +1,185 @@
...
+  state.counters["alpha"] = static_cast<double>(alpha);
```
OuadiElfarouki, line 57: We're not used to store the alpha param counter as it doesn't really affect the 'computations' _(even-though it got actually merged with the spr PR within sycl-blas & clblast benchs)_. What do you think @muhammad-tanvir-1211 ? I think we should standardize the choice of parameters to store as counters in benchmarks, namely keep the ones that do affect computations (dimensions, transposition etc..).

benchmark/cublas/blas2/spr2.cpp
```diff
@@ -0,0 +1,186 @@
...
+  state.counters["incX"] = incX;
+  state.counters["incY"] = incY;
+
+  const double nflops_XtimesX = 2.0 * n_d;
+  const double nflops_tot = size + nflops_XtimesX;
+  state.counters["n_fl_ops"] = nflops_tot;
```
OuadiElfarouki, line 63: There are additional operations & memory accesses in `spr2` compared to `spr`. Can you double check them and update these ? I've checked the theoretical values document and they seem correct.

benchmark/cublas/blas2/syr2.cpp
s-Nick: I am following the same convention of `spr` in syclblas, where it uses `size_d` as counter. We could decide to use `n` in all symmetric operators or use the actual `size_d` and set it as standard, ok?

benchmark/cublas/blas2/spr2.cpp
s-Nick: You are absolutely right! Thank you for spotting it

muhammad-tanvir-1211 (approved):

OuadiElfarouki (approved):

PR #391, opened 2023-03-30
Adding USM support for Level 3 operators in SYCL-BLAS

muhammad-tanvir-1211:
This PR extends the USM support in SYCL-BLAS for Level 3 operators. All Level 3 benchmarks and tests now use USM and Buffers.

FYI: This PR is blocked by #385 and #388

### DIFF
--- a/src/operations/blas3/gemm_local.hpp
+++ b/src/operations/blas3/gemm_local.hpp
@@ -278,12 +278,9 @@ class Gemm<input_t, output_t, DoubleBuffer, NbcA, NbcB, ClSize, TileType,
     const index_t b_size = trans_b ? ldb * k : n * ldb;
     const index_t c_size = ldc * n;
 
-    auto ptr_A = a_.get_data().get_pointer() + a_.get_access_displacement() +
-                 (wg_batch_id * stridea_);
-    auto ptr_B = b_.get_data().get_pointer() + b_.get_access_displacement() +
-                 (wg_batch_id * strideb_);
-    auto ptr_C = c_.get_data().get_pointer() + c_.get_access_displacement() +
-                 (wg_batch_id * stridec_);
+    auto ptr_A = a_.get_pointer() + (wg_batch_id * stridea_);
+    auto ptr_B = b_.get_pointer() + (wg_batch_id * strideb_);
+    auto ptr_C = c_.get_pointer() + (wg_batch_id * stridec_);
 
     const index_t item_id = id.get_local_id(0);
     const index_t tile_id = wg_id / tile_size;
--- a/test/unittest/blas3/blas3_trsm_test.cpp
+++ b/test/unittest/blas3/blas3_trsm_test.cpp
@@ -111,6 +111,8 @@ void run_test(const combination_t<scalar_t> combi) {
   if (alloc == "usm") {
 #ifdef SB_ENABLE_USM
     run_test<scalar_t, helper::AllocType::usm>(combi);
+#else
+    GTEST_SKIP();
 #endif
   } else {
     run_test<scalar_t, helper::AllocType::buffer>(combi);
--- a/test/unittest/blas3/blas3_trsm_test.cpp
+++ b/test/unittest/blas3/blas3_trsm_test.cpp
@@ -24,11 +24,12 @@
 #include "blas_test.hpp"
 
 template <typename scalar_t>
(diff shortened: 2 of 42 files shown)


abagusetty:
@muhammad-tanvir-1211  Thanks for the USM support. I was wondering if there is any timeline on this PR, as I am mostly looking forward to integrate this repo into an application for L3 APIs to benchmark at scale on all the different vendor GPUs.

mehdi-goli (approved):
LGTM! Just check it doesn't break auto tuner as well.

src/operations/blas3/gemm_local.hpp
```diff
@@ -278,12 +278,9 @@ class Gemm<input_t, output_t, DoubleBuffer, NbcA, NbcB, ClSize, TileType,
     const index_t b_size = trans_b ? ldb * k : n * ldb;
     const index_t c_size = ldc * n;
 
-    auto ptr_A = a_.get_data().get_pointer() + a_.get_access_displacement() +
-                 (wg_batch_id * stridea_);
-    auto ptr_B = b_.get_data().get_pointer() + b_.get_access_displacement() +
-                 (wg_batch_id * strideb_);
-    auto ptr_C = c_.get_data().get_pointer() + c_.get_access_displacement() +
-                 (wg_batch_id * stridec_);
+    auto ptr_A = a_.get_pointer() + (wg_batch_id * stridea_);
```
pgorlani, line 281: Thanks for your PR, @muhammad-tanvir-1211.

Why did you remove the `get_access_displacement`? I think it should be available also for USM.

mehdi-goli: `get_access_displacement` is not needed for the USM. It has been added because buffer could not handle capturing arithmetic operation on the device side from the host side.

mehdi-goli: yes

pgorlani:
For avoid confusion it is better to remove the `offset` argument from `make_*_view` in case of USM pointers. Moreover, the presence of this argument breaks the interface of `make_*_view` for buffers.

https://github.com/codeplaysoftware/sycl-blas/blob/5e0656c0de259578fe845ef601dc5e7eea3d2c57/include/views/view.h#L332-L334

https://github.com/codeplaysoftware/sycl-blas/blob/5e0656c0de259578fe845ef601dc5e7eea3d2c57/include/views/view.h#L340-L342

test/unittest/blas3/blas3_trsm_test.cpp
```diff
@@ -69,31 +70,64 @@ void run_test(const combination_t<scalar_t> combi) {
...
+  scalar_t ldaMul;
+  scalar_t ldbMul;
+  scalar_t unusedValue;
+  std::tie(alloc, m, n, trans, side, diag, uplo, alpha, ldaMul, ldbMul,
+           unusedValue) = combi;
+
+  if (alloc == "usm") {
+#ifdef SB_ENABLE_USM
+    run_test<scalar_t, helper::AllocType::usm>(combi);
+#endif
```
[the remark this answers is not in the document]
muhammad-tanvir-1211: I have addressed this change in the latest commit. Thanks.

muhammad-tanvir-1211:
> 

The offset has been removed from the USM view class and the `make_*_view` methods for USM. I have also removed the `get_access_displacement()` method for both `USM` and `Buffers`.

mehdi-goli:
> > 
> 
> The offset has been removed from the USM view class and the `make_*_view` methods for USM. I have also removed the `get_access_displacement()` method for both `USM` and `Buffers`.

You need to remove it from the buffer make_*_veiw as well and set it based on the buffer_iterator input. otherwise you create inconsistent make_view operator between buffer and usm

muhammad-tanvir-1211:
> > > 
> > 
> > 
> > The offset has been removed from the USM view class and the `make_*_view` methods for USM. I have also removed the `get_access_displacement()` method for both `USM` and `Buffers`.
> 
> You need to remove it from the buffer make_*_veiw as well and set it based on the buffer_iterator input. otherwise you create inconsistent make_view operator between buffer and usm

We don't have it for buffers in the `make_*_view` methods. I added the `offset` argument only for `USM` in the early stages and then forgot to remove it afterwards.

pgorlani (approved):

PR #394, opened 2023-04-10
Refactored Level 1 benchmarks to use the same counters across all libraries

muhammad-tanvir-1211:
This PR modifies the Level 1 operator benchmarks across all supported libraries to use the same `state.counters` making it easier to compare the performance across libraries.


aacostadiaz (approved):
Nice!! :D

common/include/common/state_counters.hpp
```diff
@@ -0,0 +1,137 @@
...
+  iamax = 3,
+  iamin = 4,
+  nrm2 = 5,
+  rotm = 6,
+  rotmg = 7,
+  scal = 8,
+  sdsdot = 9
+};
+
+template <Level1Op op, typename scalar_t, typename index_t>
```
aacostadiaz, line 45: What do you think about adding some unit tests for all these functions?

OuadiElfarouki (approved):
Thanks. Looks good to me!

s-Nick (approved):

PR #396, opened 2023-04-10
Refactored Level 2 benchmarks to use the same counters across all libraries

muhammad-tanvir-1211:
This PR modifies the Level 2 operator benchmarks across all supported libraries to use the same `state.counters` making it easier to compare the performance across libraries.


common/include/common/state_counters.hpp
```diff
@@ -0,0 +1,267 @@
+/***************************************************************************
```
aacostadiaz, line 1: Would it make sense to separate the states counters for different blas levels in different files? For example `blas1_state_counters.hpp`, `blas2_state_counters.hpp` and `blas3_state_counters.hpp`. We have this separation elsewhere in the codebase.

benchmark/clBench/clblast/blas2/gbmv.cpp
```diff
@@ -49,38 +49,9 @@ void run(benchmark::State& state, ExecutorType* executorPtr, int ti, index_t m,
...
-        nflops_AtimesX + nflops_timesAlpha + nflops_addBetaY;
-  }
-  {
-    double mem_readA = A_validVal;
-    double mem_readX = xlen;
-    double mem_writeY = ylen;
-    double mem_readY = (beta != scalar_t{0}) ? ylen : 0;
-    state.counters["bytes_processed"] =
-        (mem_readA + mem_readX + mem_writeY + mem_readY) * sizeof(scalar_t);
-  }
```
aacostadiaz, line 83: Love this.

aacostadiaz (approved):

s-Nick (approved):

common/include/common/blas2_state_counters.hpp
```diff
@@ -0,0 +1,267 @@
...
+                          index_t kl = 0) {
+  // Google-benchmark counters are double.
+  double n_d = static_cast<double>(n);
+  state.counters["n"] = n_d;
+  // Compute the number of A non-zero elements.
+  const double A_validVal = (n_d * (n_d + 1) / 2);
+
+  const double nflops_AtimesX = 2 * n_d * n_d;
+  const double nflops_timesAlpha = n_d;
+  const double nflops_addBetaY = (beta != scalar_t{0}) ? 2 * n_d : 0;
```
OuadiElfarouki, line 177: Thanks for the refactoring @muhammad-tanvir-1211 ! 
I was just wondering why not add beta values to the state counters for benchmarks _(applies to other operators as well)_? especially that they're affecting computations & other state counters (flops & bytes processed). This might be relevant as two benchmarks for instance might happen to have the same nflops &/or bytes processed AND still have different beta values (one being zero) : in such case the benchmarking would compare two different operations (one with an addition beta!=0 and another without) but still will consider it as a unique operation given that beta value is absent from the record. What do you think?

muhammad-tanvir-1211: That's a good point. Changing the value of `beta` does have an impact on the overall performance as well since we may or may not need to read the output from global memory within the kernel. I'll add `beta` to the counters to reflect it's impact. Thanks for brining this up.

muhammad-tanvir-1211: Done.

OuadiElfarouki (approved):

PR #399, opened 2023-04-12
Fixed the gemm_batched_strided benchmark

muhammad-tanvir-1211:
This PR adds the `gemm_batched_strided` benchmark in SYCL-BLAS. This benchmark was accidentally pushed in a previous PR in it's incomplete state, but it is now fixed and working correctly.

### DIFF
--- a/benchmark/syclblas/blas3/gemm_batched_strided.cpp
+++ b/benchmark/syclblas/blas3/gemm_batched_strided.cpp
@@ -193,6 +193,7 @@ void register_benchmark(blas_benchmark::Args& args,
 
     auto BM_lambda = [&](benchmark::State& st, blas::SB_Handle* sb_handle_ptr,
                          int t1, int t2, index_t m, index_t k, index_t n,
+                         index_t stride_a, index_t stride_b, index_t stride_c,
                          scalar_t alpha, scalar_t beta, index_t batch_size,
                          bool* success) {
       run<scalar_t>(st, sb_handle_ptr, t1, t2, m, k, n, stride_a, stride_b,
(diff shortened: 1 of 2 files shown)


OuadiElfarouki (commented):
Thanks for the fix! Overall it looks good to me if we follow the default strides approach. I've added and described the other options in this [PR](https://github.com/codeplaysoftware/sycl-blas/pull/401). If we are sure that the custom strides don't affect the benchmarking directly _(that performance only depends on batch sizes and matrix dimensions)_ then we can adopt this one I guess.

benchmark/syclblas/blas3/gemm_batched_strided.cpp
```diff
@@ -186,17 +187,21 @@ void register_benchmark(blas_benchmark::Args& args,
     int t1 = static_cast<int>(blas_benchmark::utils::to_transpose_enum(t1s));
     int t2 = static_cast<int>(blas_benchmark::utils::to_transpose_enum(t2s));
 
+    index_t stride_a = m * k;
+    index_t stride_b = k * n;
+    index_t stride_c = m * n;
+
     auto BM_lambda = [&](benchmark::State& st, blas::SB_Handle* sb_handle_ptr,
```
OuadiElfarouki, line 194: Looks like strides are not captured/passed properly inside the lambda function since it's not evaluated in-place within `RegisterBenchmark`. Passing the strides explicitly as arguments _(to be added in line 203 for as well)_ so that they're properly picked by `run<scalar>(..).` solved the observed segfaults and invalid strides.

OuadiElfarouki (approved):

PR #401, opened 2023-04-13
Add rocBLAS benchmarks for level3 operators and blas-like extensions

OuadiElfarouki:
This PR extends rocBLAS benchmarks to level 3 operators & batched extensions including : 

- gemm
- symm
- syrk
- syr2k
- trmm
- trsm
- trsm_batched
- gemm_batched
- gemm_batched_strided

### DIFF
--- a/benchmark/rocblas/blas3/gemm.cpp
+++ b/benchmark/rocblas/blas3/gemm.cpp
@@ -44,14 +44,14 @@ static inline void rocblas_gemm_f(args_t&&... args) {
 }
 
 template <typename scalar_t>
-void run(benchmark::State& state, rocblas_handle& rb_handle, int t_ai, int t_bi,
-         index_t m, index_t k, index_t n, scalar_t alpha, scalar_t beta,
-         bool* success) {
+void run(benchmark::State& state, rocblas_handle& rb_handle, int t_a_i,
+         int t_b_i, index_t m, index_t k, index_t n, scalar_t alpha,
+         scalar_t beta, bool* success) {
   // Standard test setup.
   std::string t_a = blas_benchmark::utils::from_transpose_enum(
-      static_cast<blas_benchmark::utils::Transposition>(t_ai));
+      static_cast<blas_benchmark::utils::Transposition>(t_a_i));
   std::string t_b = blas_benchmark::utils::from_transpose_enum(
-      static_cast<blas_benchmark::utils::Transposition>(t_bi));
+      static_cast<blas_benchmark::utils::Transposition>(t_b_i));
   const char* t_a_str = t_a.c_str();
   const char* t_b_str = t_b.c_str();
 
@@ -123,10 +123,8 @@ void run(benchmark::State& state, rocblas_handle& rb_handle, int t_ai, int t_bi,
     // Rocblas verification gemm
     std::vector<scalar_t> c_temp = c;
     {
-      // Temp result on device (copied back to Host upon destruction)
       blas_benchmark::utils::HIPVector<scalar_t, true> c_temp_gpu(
           c_size, c_temp.data());
-      // rocBLAS function call
       rocblas_gemm_f<scalar_t>(rb_handle, trans_a_rb, trans_b_rb, m, n, k,
                                &alpha, a_gpu, lda, b_gpu, ldb, &beta,
                                c_temp_gpu, ldc);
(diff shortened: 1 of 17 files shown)


OuadiElfarouki:
Two approaches dealing with the `strides` parametrizations in batched (strided) operations are implemented here for discussion : 
- [trsm_batched.cpp](https://github.com/codeplaysoftware/sycl-blas/pull/401/files#diff-d3cde8d5e1e29b1143522f84e765a1af6b497803923d0bbbdcd38ed756d52fe7) : Strides are used / passed as they are to the benchmark and used directly by the operators. This can be bulky during benchmark results analysis/interpretation as these strides are included in counters.
- [gemm_batched_strided.cpp](https://github.com/codeplaysoftware/sycl-blas/pull/401/files#diff-3841847483edb733397e8f956027bc3e82465eb953d9f02076bc972da2ab60f3) : Instead of explicit strides, stride_multipliers are passed to the benchmark for clarity/simplification purposes, for instance `stride_a = 1` refers to a stride equal to matrix a size. This can be more meaningful and concise when examining the benchmark results and also while passing custom parameters for benchmarks. The same approach is currently adopted in the `gemm_batched_strided` [unittests](https://github.com/codeplaysoftware/sycl-blas/blob/master/test/unittest/blas3/blas3_gemm_batched_test.cpp#L117)  

A third option discussed in this [PR](https://github.com/codeplaysoftware/sycl-blas/pull/399) can be followed as well : hide the strides from the benchmarking and use default values.

common/include/common/common_utils.hpp
```diff
@@ -78,8 +83,9 @@ using tbmv_param_t =
 using trsv_param_t = std::tuple<std::string, std::string, std::string, index_t>;
 
 template <typename scalar_t>
-using trsm_batched_param_t = std::tuple<char, char, char, char, index_t,
-                                        index_t, scalar_t, index_t, int>;
+using trsm_batched_param_t =
```
OuadiElfarouki, line 86: This change will break the cuBLAS `trsm_batched` [benchmark](https://github.com/codeplaysoftware/sycl-blas/blob/master/benchmark/cublas/blas3/trsm_batched.cpp#L223) if merged.

benchmark/rocblas/blas3/symm.cpp
```diff
@@ -0,0 +1,206 @@
...
+      blas_benchmark::utils::random_data<scalar_t>(c_size);
+
+  {
+    // Device memory allocation & H2D copy
+    blas_benchmark::utils::HIPVector<scalar_t> a_gpu(a_size, a.data());
+    blas_benchmark::utils::HIPVector<scalar_t> b_gpu(b_size, b.data());
+    blas_benchmark::utils::HIPVector<scalar_t> c_gpu(c_size, c.data());
+
+    CHECK_ROCBLAS_STATUS(
+        rocblas_set_pointer_mode(rb_handle, rocblas_pointer_mode_host));
```
s-Nick, line 111: Thank you for your work @OuadiElfarouki ! 
Do we need to set the pointer-mode for  this operator?

benchmark/rocblas/blas3/gemm.cpp
```diff
@@ -0,0 +1,214 @@
...
+    double total_mem =
+        (mem_readA + mem_readB + mem_readC + mem_writeC) * sizeof(scalar_t);
+    state.counters["bytes_processed"] = total_mem;
+    state.SetBytesProcessed(state.iterations() * total_mem);
+
+    double nflops_AtimesB = (2 * k_d) * m_d * n_d;
+    double nflops_addBetaC = (beta != scalar_t{0}) ? 2 * m_d * n_d : 0;
+    double nflops_tot = nflops_AtimesB + nflops_addBetaC;
+    state.counters["n_fl_ops"] = nflops_tot;
+    state.SetItemsProcessed(state.iterations() * nflops_tot);
```
s-Nick, line 86: Did you check if using `state.iterations()` here work? I think this and *line 80* should be moved after the loop *line 168-175*, otherwise the number of iteration could be 0

benchmark/rocblas/blas3/trsm.cpp
```diff
@@ -0,0 +1,211 @@
...
+  blas_benchmark::utils::fill_trsm_matrix(a, k, lda, uplo, diag_value,
+                                          scalar_t{0});
+
+  {
+    // Device memory allocation & H2D copy
+    blas_benchmark::utils::HIPVector<scalar_t> a_gpu(a_size, a.data());
+    blas_benchmark::utils::HIPVector<scalar_t> b_gpu(b_size, b.data());
+
+    CHECK_ROCBLAS_STATUS(
+        rocblas_set_pointer_mode(rb_handle, rocblas_pointer_mode_host));
```
s-Nick, line 115: is this necessary?

OuadiElfarouki: It looks like it's not. Couldn't find in the doc if it's set to `rocblas_pointer_mode_host`by default but results show no difference between enabling/disabling it. Will remove it then thanks!

benchmark/rocblas/blas3/gemm.cpp
OuadiElfarouki: Correct! I've just checked the doc and the results. Will fix these thank you.

OuadiElfarouki:
> Two approaches dealing with the `strides` parametrizations in batched (strided) operations are implemented here for discussion :
> 
>     * [trsm_batched.cpp](https://github.com/codeplaysoftware/sycl-blas/pull/401/files#diff-d3cde8d5e1e29b1143522f84e765a1af6b497803923d0bbbdcd38ed756d52fe7) : Strides are used / passed as they are to the benchmark and used directly by the operators. This can be bulky during benchmark results analysis/interpretation as these strides are included in counters.
> 
>     * [gemm_batched_strided.cpp](https://github.com/codeplaysoftware/sycl-blas/pull/401/files#diff-3841847483edb733397e8f956027bc3e82465eb953d9f02076bc972da2ab60f3) : Instead of explicit strides, stride_multipliers are passed to the benchmark for clarity/simplification purposes, for instance `stride_a = 1` refers to a stride equal to matrix a size. This can be more meaningful and concise when examining the benchmark results and also while passing custom parameters for benchmarks. The same approach is currently adopted in the `gemm_batched_strided` [unittests](https://github.com/codeplaysoftware/sycl-blas/blob/master/test/unittest/blas3/blas3_gemm_batched_test.cpp#L117)
> 
> 
> A third option discussed in this [PR](https://github.com/codeplaysoftware/sycl-blas/pull/399) can be followed as well : hide the strides from the benchmarking and use default values.

The first approach is deemed to verbose and was replaced with the second one (stride multipliers instead of explicit strides)

muhammad-tanvir-1211 (approved):

PR #411, opened 2023-04-25
Enabled vectorization for axpy

OuadiElfarouki:
Axpy vectorization through : 
- vectorized ScalOp
- vectorized BinaryOp
- vectorized Assign

### DIFF
--- a/src/operations/blas1_trees.hpp
+++ b/src/operations/blas1_trees.hpp
@@ -314,12 +314,32 @@ SYCL_BLAS_INLINE typename ScalarOp<operator_t, scalar_t, rhs_t>::index_t
 ScalarOp<operator_t, scalar_t, rhs_t>::get_size() const {
   return rhs_.get_size();
 }
+
+template <typename operator_t, typename scalar_t, typename rhs_t>
+template <int vector_size>
+SYCL_BLAS_INLINE typename ScalarOp<operator_t, scalar_t, rhs_t>::index_t
+ScalarOp<operator_t, scalar_t, rhs_t>::get_size_vector() const {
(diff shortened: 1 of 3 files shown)


src/operations/blas1_trees.hpp
```diff
@@ -374,6 +465,25 @@ SYCL_BLAS_INLINE typename BinaryOp<operator_t, lhs_t, rhs_t>::value_t
...
   return BinaryOp<operator_t, lhs_t, rhs_t>::eval(ndItem.get_global_id(0));
 }
+
+template <typename operator_t, typename lhs_t, typename rhs_t>
+template <int vector_size>
+SYCL_BLAS_INLINE cl::sycl::vec<
+    typename BinaryOp<operator_t, lhs_t, rhs_t>::value_t, vector_size>
+BinaryOp<operator_t, lhs_t, rhs_t>::eval_vector(
+    typename BinaryOp<operator_t, lhs_t, rhs_t>::index_t i) {
+  return operator_t::eval(lhs_.template eval_vector<vector_size>(i),
```
aacostadiaz, line 475: Should it be eval_vector ?

OuadiElfarouki: @aacostadiaz From what I saw the [Operators](https://github.com/codeplaysoftware/sycl-blas/blob/master/src/operations/blas_operators.hpp#L225) don't seem to need a specialization for the `eval` functions as they're already templated by the `lhs` and `rhs` types and in the case of vector inputs they'll handle them  (as the `sycl::vec` interface already overloads the main operators used here). What do you think?

muhammad-tanvir-1211:
Closing this PR as the performance of vectorization doesn't look promising.

PR #416, opened 2023-05-04
Add SPMV Level 2 BLAS Operator

pgorlani:
This patch adds an implementation of SPMV Level 2 operator compatible with `TUNING_TARGET` specializations.


benchmark/syclblas/blas2/spmv.cpp
```diff
@@ -0,0 +1,146 @@
...
+  str << "BM_Spmv<" << blas_benchmark::utils::get_type_name<scalar_t>() << ">/"
+      << uplo << "/" << n << "/" << alpha << "/" << beta;
+  return str.str();
+}
+
+template <typename scalar_t>
+void run(benchmark::State& state, blas::SB_Handle* sb_handle_ptr,
+         std::string uplo, index_t n, scalar_t alpha, scalar_t beta,
+         bool* success) {
+  // Standard test setup.
```
muhammad-tanvir-1211, line 40: Is this a misplaced comment here?

src/interface/blas2/backend/amd_gpu.hpp
```diff
@@ -86,6 +86,22 @@ typename SB_Handle::event_t inline _sbmv(SB_Handle& sb_handle, index_t _N,
 }  // namespace backend
 }  // namespace sbmv
 
+namespace spmv {
+namespace backend {
+template <uplo_type uplo, typename SB_Handle, typename index_t,
+          typename element_t, typename container_t0, typename container_t1,
+          typename increment_t, typename container_t2>
+typename SB_Handle::event_t inline _spmv(SB_Handle& sb_handle, index_t _N,
```
muhammad-tanvir-1211, line 94: Maybe we can use `SYCL_BLAS_INLINE` here (and in the other places) to match with the changes inside `operations/blas2/spmv.hpp` file.

src/operations/blas2/spmv.hpp
```diff
@@ -0,0 +1,284 @@
...
+ *  @filename spmv.hpp
+ *
+ **************************************************************************/
+
+#ifndef SPMV_HPP
+#define SPMV_HPP
+#include "operations/blas2_trees.h"
+#include "operations/blas_operators.hpp"
+#include "views/view_sycl.hpp"
+#include <stdexcept>
```
muhammad-tanvir-1211, line 31: Does the implementation need the `stdexcept` and `vector` includes here?

src/operations/blas2/spmv.hpp
```diff
@@ -0,0 +1,284 @@
...
+      alpha_(_alpha),
+      beta_(_beta) {}
+
+template <typename lhs_t, typename matrix_t, typename vector_t,
+          uint32_t local_range_x, uint32_t local_range_y, bool is_upper>
+SYCL_BLAS_INLINE typename Spmv<lhs_t, matrix_t, vector_t, local_range_x,
+                               local_range_y, is_upper>::index_t
+Spmv<lhs_t, matrix_t, vector_t, local_range_x, local_range_y,
+     is_upper>::get_size() const {
+  return matrix_.get_size();
```
muhammad-tanvir-1211, line 61: Shouldn't we be returning the `output` size here?

src/operations/blas2/spmv.hpp
```diff
@@ -0,0 +1,284 @@
...
+
+  const index_t N = lhs_.get_size();
+  const index_t nblock = (N + x_range - 1) / x_range;
+
+  value_t *const loc_x = shrMem.localAcc.get_pointer();
+  value_t *const loc_A = loc_x + x_range;
+
+  // ------------------------------------------------------------------------ //
+
+  auto _mat_J_offset = [&N](const index_t &_J) {
```
muhammad-tanvir-1211, line 98: I think it's better to use `if` conditions instead of these `lambda` functions, the compiler may be able to flatten the code out the `if` condition since the code is dependent on a template parameter but the lambda object may not go away.

src/operations/blas2/spmv.hpp
```diff
@@ -0,0 +1,284 @@
...
+    for (index_t _j = 0; _j < y_range; ++_j) res += loc_A[loc_lda * _j + l_idx];
+
+    const index_t lhs_idx = gid * x_range + l_idx;
+    if (lhs_idx < N)
+      lhs_.eval(lhs_idx) =
+          AddOperator::eval(ProductOperator::eval(alpha_, res),
+                            ProductOperator::eval(beta_, lhs_.eval(lhs_idx)));
+  }
+
+  return 0;
```
muhammad-tanvir-1211, line 261: I think it would be better if we return the actual output here instead of `0`. If someone decides to use this node in a tree than the logic will break as we will get only `0` as output for all the work items involved.

test/unittest/blas2/blas2_spmv_test.cpp
```diff
@@ -0,0 +1,116 @@
...
+ *  SYCL-BLAS: BLAS implementation using SYCL
+ *
+ *  @filename blas2_spmv_test.cpp
+ *
+ **************************************************************************/
+
+#include "blas_test.hpp"
+
+template <typename T>
+using combination_t = std::tuple<int, T, T, bool, int, int>;
```
muhammad-tanvir-1211, line 29: Minor suggestion, but maybe we can use `index_t` instead of `int` here.

test/unittest/blas2/blas2_spmv_test.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+  const bool isAlmostEqual = utils::compare_vectors(y_v_gpu_result, y_v_cpu);
+
+  ASSERT_TRUE(isAlmostEqual);
+}
+
+#ifdef STRESS_TESTING
+template <typename scalar_t>
+const auto combi =
+    ::testing::Combine(::testing::Range(1, 999),                    // n
```
muhammad-tanvir-1211, line 86: Maybe we can reduce the no. of tests here? Waiting ~5 minutes for 1 test may not be ideal on every machine.

benchmark/syclblas/blas2/spmv.cpp
```diff
@@ -0,0 +1,146 @@
...
+    std::tie(uplos, n, alpha, beta) = p;
+
+    auto BM_lambda = [&](benchmark::State& st, blas::SB_Handle* sb_handle_ptr,
+                         std::string uplos, index_t n, scalar_t alpha,
+                         scalar_t beta, bool* success) {
+      run<scalar_t>(st, sb_handle_ptr, uplos, n, alpha, beta, success);
+    };
+    benchmark::RegisterBenchmark(
+        get_name<scalar_t>(uplos, n, alpha, beta).c_str(), BM_lambda,
+        sb_handle_ptr, uplos, n, alpha, beta, success);
```
s-Nick, line 137: Thank you @pgorlani.
Missing `->UseRealTime()` when register the benchmark, can you add it?
```suggestion
        sb_handle_ptr, uplos, n, alpha, beta, success)
        ->UseRealTime();
```

pgorlani: Thanks for your review, @s-Nick.

Done.

benchmark/syclblas/blas2/spmv.cpp
pgorlani: Thanks for your review, @muhammad-tanvir-1211.

I removed this comment from all `spmv` benchmarks.

src/operations/blas2/spmv.hpp
s-Nick: Maybe we could use `if constexpr` to be sure the compiler selects the correct one, couldn't we?

test/unittest/blas2/blas2_spmv_test.cpp
pgorlani: Yes, I agree with you.

This is the reason why the above test combination is not run by default, but just if `STRESS_TESTING` is defined.

src/operations/blas2/spmv.hpp
pgorlani: No, I removed them.

test/unittest/blas2/blas2_spmv_test.cpp
pgorlani: Done

src/operations/blas2/spmv.hpp
pgorlani: Done.

src/operations/blas2/spmv.hpp
pgorlani: I introduced these lambda functions for the sake of making this kernel more readable/maintainable by avoiding the code duplication, and potentially (one day with a similar approach) share the common code among different kernels.

So far, these lambdas are handled correctly by the compiler. They are inlined automatically and the ternary operator that they contain is simplified at compile time since depends on a constant. This basic optimization is performed easily by clang. I understand your concerns related to the possibility that those lambdas could not be inlined, but for now the inline issues we encountered are related to functions having many instructions/arguments.

s-Nick (approved):

benchmark/cublas/blas2/spmv.cpp
```diff
@@ -63,13 +62,13 @@ void run(benchmark::State& state, cublasHandle_t* cuda_handle_ptr,
 
   // Input matrix/vector, output vector.
   std::vector<scalar_t> m_a =
-      blas_benchmark::utils::random_data<scalar_t>(n * n);
```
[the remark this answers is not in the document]
pgorlani: I was not aware of this. I'm having no issues with OpenBLAS.

If you want I can revert it, but probably it is better to use the `n*n` size just for the problematic cases.

muhammad-tanvir-1211 (approved):
LGTM. Thank you for addressing all the comments.

PR #417, opened 2023-05-11
Add TPMV Level 2 BLAS Operator

pgorlani:
This patch adds an implementation of TPMV Level 2 operator compatible with `TUNING_TARGET` specializations for `INTEL_GPU`, `AMD_GPU` and `NVIDIA_GPU`.

This depends on https://github.com/codeplaysoftware/sycl-blas/pull/416.


src/interface/blas2_interface.hpp
```diff
@@ -551,14 +553,18 @@ typename sb_handle_t::event_t _spmv_impl(sb_handle_t& sb_handle, index_t _N,
   auto vx = make_vector_view(_vx, _incx, vector_size);
   auto vy = make_vector_view(_vy, _incy, vector_size);
 
-  auto spmv = make_spmv<local_range_x, local_range_y, uplo == uplo_type::Upper>(
-      _alpha, mA, vx, _beta, vy);
+  auto spmv =
+      make_xpmv<local_range_x, local_range_y, /*is_symmetric=*/true, is_upper,
+                /*is_transposed=*/false, /*is_unit=*/false>(_alpha, mA, vx,
```
s-Nick, line 558: Thank you @pgorlani  for you work. I think these comments inside the template could be removed even if they increase readability.

src/operations/blas2/xpmv.hpp
```diff
@@ -176,12 +190,13 @@ Spmv<lhs_t, matrix_t, vector_t, local_range_x, local_range_y, is_upper>::eval(
 
   // VERTICAL
 
-  {
+  if (is_symmetric || is_transposed) {
```
aacostadiaz, line 193: I think you can use `if constexpr (is_symmetric || is_transposed)` to move this check to compile time.

aacostadiaz (approved):
Nice generalisation in Xpmv

src/interface/blas2_interface.hpp
```diff
@@ -598,6 +604,51 @@ typename sb_handle_t::event_t _tbmv_impl(sb_handle_t& sb_handle, index_t _N,
...
+
+  typename container_t0::scalar_t unused;
+
+  auto tpmv =
+      make_xpmv<local_range_x, local_range_y, /*is_symmetric=*/false, is_upper,
+                is_transposed, is_unit>(unused, mA, vx, unused, vres);
+
+  const index_t loc_mem_leading_dim = local_range_x + 1;
+
+  auto tpmvEvent = sb_handle.execute(
```
muhammad-tanvir-1211, line 642: Can we avoid making multiple calls to `sb_handle.execute()` here? This would lead to 2 independent `queue.submit()` calls and will also create a dependency between the two kernels as well which will need to be captured for the `usm` case.

muhammad-tanvir-1211 (commented):
Thanks for the PR Paolo. It looks good, just that it would be good if we could find a way around multiple calls to `sb_handle.execute()`.

src/operations/blas2/xpmv.hpp
pgorlani: Thanks for your review, @aacostadiaz.

I understand your concerns, but the fact that `is_simmetric` and `is_transposed` are two template parameters is enough for allowing clang to evaluate these if statements on compile time.

src/interface/blas2_interface.hpp
pgorlani: Thanks your review, @muhammad-tanvir-1211.

Unfortunately, I cannot. Basically, I'm forced to have an assign operation for `tpmv` since the routine result is stored in the same vector that contain the routine input, i.e. `vx`. For such a reason, I need a temporary vector (`vres`) for storing the solution that will be assigned to the input/output vector (`vx`) passed as routine argument.

src/interface/blas2_interface.hpp
pgorlani: Thanks for your review, @s-Nick.

I prefer to keep these comments for having a clearer code.

mehdi-goli: If the `make_xpmv` has a proper documentation, keeping this just create boiler plates and is not necessary.

pgorlani: Hi @mehdi-goli,

I think that `make_xpmv` is self-explanatory due to the template parameter names and types. This is just my opinion, I'm open to suggestions. That is the reason I think that recalling the template parameters here can be helpful considering also the fact that `make_xpmv` is defined on another file.

I don't think these three little comments will bloat the code, but I can remove them if you think so.

mehdi-goli: In SYCL-BLAS we provide doxygen for all the functions and classes we generate. So we must follow that. and one place will capture all these info regarding the function declaration. It is not only this function, This is for the repetition of such functions so imagine everywhere such similar functions used, we put multiple comments there. So if we have proper doxygen for the function, the generator provide proper documentation. hence, no need for such comments

muhammad-tanvir-1211:
Can you also update the operator table in the README to show that TPMV has been added into the library as part of this PR?

src/interface/blas2_interface.hpp
pgorlani: https://github.com/codeplaysoftware/sycl-blas/pull/417/commits/4dff277d2af77538c07f40e4dad3ad681f1c704a updates the documentation.

pgorlani:
> Can you also update the operator table in the README to show that TPMV has been added into the library as part of this PR?

Done in https://github.com/codeplaysoftware/sycl-blas/pull/417/commits/9c78b834d552f4f1fb26d04fc4e850245e1c11c6.

PR #418, opened 2023-05-12
Fixed string/char type issue in `spr2` rocblas benchmark

OuadiElfarouki:
Fixed string/char type issue in `spr2` operator rocBLAS benchmark causing compilation issue.


muhammad-tanvir-1211 (approved):
LGTM, Thanks.

s-Nick (approved):

PR #420, opened 2023-05-17
 Added device/compiler info and commit hash to state label in cublas benchmarks

muhammad-tanvir-1211:
This PR adds `device` info, `compiler` info and `commit` info to the `state` object in the `cublas` benchmarks and sets the `state` object's `label` accordingly.


aacostadiaz (approved):
Nice, really easy to review!

OuadiElfarouki (approved):

PR #422, opened 2023-05-22
Build fixes when using -DTUNING_TARGET=POWER_VR and -DIMGDNN_DIR=...

robquill:
Fixes the build when using: -DTUNING_TARGET=POWER_VR on it's own and also when using -DTUNING_TARGET=POWER_VR -DIMGDNN_DIR=...


pgorlani (approved):
Thanks for your patch, @robquill!

`_tbsv` looks good to me.

PR #423, opened 2023-05-25
Added USM support in portBLAS

muhammad-tanvir-1211:
This PR introduces USM support in portBLAS for all supported BLAS operators.

This PR is blocked on #424 and #425


aacostadiaz (approved):
LGTM!

include/views/view.h
```diff
@@ -130,79 +146,113 @@ struct MatrixView {
...
-  /*! get_access_device.
-   * @brief Access on the Device (e.g CPU: Row, GPU: Column).
-   */
-  int get_access_device() const;
+  SYCL_BLAS_INLINE index_t get_size_col() const;
 
   /*! adjust_access_displacement.
    * @brief set displacement from the origin.
+   * For USM case, this method is not going to do anything as the library
+   * doesn't allow pointer manipulation.
```
pgorlani, line 202: I suggest integrating the comment as follows for making things clearer.
```suggestion
   * This method allows to have a pointer arithmetic semantics for buffers
   * in the host code. The end result of the pointer arithmetic is passed  
   * as an access displacement for the buffer.
   * In the USM case, this method is not going to do anything since the
   * pointer arithmetic is performed implicitly.
```

include/sycl_blas_helper.h
```diff
@@ -32,6 +32,89 @@
...
+  return event;
+}
+
+template <AllocType alloc, typename container_t>
+typename std::enable_if<alloc == AllocType::buffer, cl::sycl::event>::type
+enqueue_deallocate(std::vector<cl::sycl::event>, container_t mem,
+                   cl::sycl::queue q) {
+  cl::sycl::event event;
+#ifdef SB_ENABLE_USM
+  event = q.submit([&](cl::sycl::handler &cgh) { cgh.host_task([=]() {}); });
```
pgorlani, line 112: What is the reason of spawning a `host_task` that does nothing? Can we remove it?

src/interface/blas2_interface.hpp
```diff
@@ -119,16 +123,26 @@ typename sb_handle_t::event_t _gemv_impl(sb_handle_t& sb_handle, index_t _M,
...
     }
 
+#ifdef DEFAULT_CPU
+    auto free_dt_buffer =
+        blas::helper::enqueue_deallocate < is_usm
+            ? helper::AllocType::usm
+            : helper::AllocType::buffer >
+                  (ret, dot_products_buffer, sb_handle.get_queue());
+    ret =
+        concatenate_vectors(ret, typename sb_handle_t::event_t{free_dt_buffer});
```
pgorlani, line 144: In order to remove the error due to queue profiling that you mentioned [here](https://github.com/codeplaysoftware/sycl-blas/blob/60a808d30b33dbd0efe4a9fcbe6d07396c98c1a0/include/sycl_blas_helper.h#L86), you can remove  `free_dt_buffer` from the event list that you return.

The SYCL [specification](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:interface.queue.class) guarantee us that the corresponding operation will be executed in any case.
```
A SYCL queue may be destroyed even when there are uncompleted commands that have been submitted to the queue. Doing so does not block. Instead, any commands that have been submitted to the queue begin execution when their requisites are satisfied, just as they would had the queue not been destroyed. Any event objects for those commands are signaled in the normal manner when the command completes. Resources associated with the queue will be freed by the time the last command completes.
```

muhammad-tanvir-1211: This change has been addressed in https://github.com/codeplaysoftware/sycl-blas/pull/436

cmake/CmakeFunctionHelper.cmake
```diff
@@ -93,6 +93,9 @@ set(LOCATION "${SYCLBLAS_GENERATED_SRC}/${blas_level}/${func}/")
 foreach(data ${data_list})
   cpp_type(cpp_data ${data})
   set(container_list "BufferIterator<${cpp_data}>")
+  if(is_dpcpp)
```
pgorlani, line 96: I think that the generation of the USM routine version should depend on a variable like `SB_ENABLE_USM` rather than `dpcpp` itself.

Moreover, it must be noted that mixing USM and Buffer arguments is possible just in case of header only compilation. Which is fine, but probably it needs to be stated somewhere.

src/interface/blas2_interface.hpp
```diff
@@ -678,19 +779,37 @@ typename sb_handle_t::event_t _tbsv_impl(sb_handle_t& sb_handle, index_t _N,
...
-  auto sync_buffer =
-      blas::make_sycl_iterator_buffer<int32_t>(sync_vec, sync_vec.size());
-  auto sync = make_vector_view(sync_buffer, one::value(), sync_vec.size());
+  constexpr bool is_usm = std::is_pointer<container_t0>::value;
+  auto queue = sb_handle.get_queue();
+
+  auto sync_buffer = blas::helper::allocate < is_usm
+                         ? blas::helper::AllocType::usm
+                         : blas::helper::AllocType::buffer,
+       index_t > (sync_vec.size(), queue);
```
pgorlani, line 788:
```suggestion
       int32_t > (sync_vec.size(), queue);
```

src/interface/blas2_interface.hpp
```diff
@@ -678,19 +779,37 @@ typename sb_handle_t::event_t _tbsv_impl(sb_handle_t& sb_handle, index_t _N,
...
+  auto copy_sync = blas::helper::copy_to_device<index_t>(
```
pgorlani, line 789:
```suggestion
  auto copy_sync = blas::helper::copy_to_device<int32_t>(
```

src/interface/blas2_interface.hpp
```diff
@@ -346,19 +384,37 @@ typename sb_handle_t::event_t _trsv_impl(sb_handle_t& sb_handle, index_t _N,
...
 
-  auto sync_buffer =
-      blas::make_sycl_iterator_buffer<int32_t>(sync_vec, sync_vec.size());
-  auto sync = make_vector_view(sync_buffer, one::value(), sync_vec.size());
+  auto queue = sb_handle.get_queue();
+  constexpr bool is_usm = std::is_pointer<container_t0>::value;
+  auto sync_buffer = blas::helper::allocate < is_usm
+                         ? blas::helper::AllocType::usm
+                         : blas::helper::AllocType::buffer,
+       index_t > (sync_vec.size(), queue);
```
pgorlani, line 392:
```suggestion
      int32_t > (sync_vec.size(), queue);
```

src/interface/blas2_interface.hpp
```diff
@@ -346,19 +384,37 @@ typename sb_handle_t::event_t _trsv_impl(sb_handle_t& sb_handle, index_t _N,
...
+  auto copy_sync = blas::helper::copy_to_device<index_t>(
```
pgorlani, line 393:
```suggestion
  auto copy_sync = blas::helper::copy_to_device<int32_t>(
```

src/interface/blas2_interface.hpp
mehdi-goli: I would strongly suggest not to hardcode index type. Intel in onemkl uses `int64_t` for the index, so we need to let the index type to be adopted to what has been chosen through template

src/interface/blas2_interface.hpp
mehdi-goli: Same as above

src/interface/blas2_interface.hpp
pgorlani: As you prefer, @mehdi-goli.

But, not all the devices support 64 bit atomic operations. Moreover, specifying `int32_t` will guarantee that this will have 32 bits no matter what host platform we are using.

mehdi-goli: If the device does not support we can set the index through template to be 32. But when we put this code as a backend of onemkl intel all the index must be 64. The point is when we have it in index we can control it for all the devices. But in onemkl case that can lead to index overflow as they are using very large index hence they force it to be 64

include/views/view.h
muhammad-tanvir-1211: This has been addressed in https://github.com/codeplaysoftware/sycl-blas/pull/441

cmake/CmakeFunctionHelper.cmake
muhammad-tanvir-1211: This has been addressed in https://github.com/codeplaysoftware/sycl-blas/pull/441

src/interface/blas2_interface.hpp
muhammad-tanvir-1211: This and other suggestions below have been addressed in https://github.com/codeplaysoftware/sycl-blas/pull/441

muhammad-tanvir-1211: We decided to stick to `int32_t` for the above.

OuadiElfarouki (approved):

PR #425, opened 2023-05-25
Add gtest_skip for USM tests

muhammad-tanvir-1211:
This PR adds `GTEST_SKIP()` for tests when USM is not supported for better visibility.


pgorlani (approved):

aacostadiaz (approved):
LGTM! 

TOL: We should add a job in CI to run USM tests

PR #426, opened 2023-05-26
Fixed rotmg test instantiation

muhammad-tanvir-1211:
Fixed a minor bug in the `rotmg` tests instantiation.

### DIFF
--- a/benchmark/syclblas/blas1/copy.cpp
+++ b/benchmark/syclblas/blas1/copy.cpp
@@ -26,14 +26,15 @@
 #include "../utils.hpp"
 
 template <typename scalar_t>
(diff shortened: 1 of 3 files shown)


aacostadiaz:
The benchmark for this operator is also failing. Could it be due to the same issue?

muhammad-tanvir-1211:
> The benchmark for this operator is also failing. Could it be due to the same issue?

I made a silly mistake when putting up this PR, which is the reason why the CI was failing for ComputeCpp. It should be fixed now.

The benchmarks are not failing on my machine, they pass for both `usm` and `buffers`. Can you try running them again?

pgorlani (approved):

aacostadiaz (approved):

PR #428, opened 2023-05-30
Add omatcopy and omatcopy2 operators support

s-Nick:
This PR add two blas extension operators.
The operators are quite similar and the implementation in splitted in two cases, one for non-transpose input matrix and one for transpose input matrix.

Co-authored-by: Ouadie El Farouki <ouadie.elfarouki@codeplay.com>


src/views/view_sycl.hpp
```diff
@@ -218,17 +218,23 @@ struct MatrixView<
...
   container_t data_;
   // Information related to the operation
-  const index_t sizeR_;  // number of rows
-  const index_t sizeC_;  // number of columns
-  const index_t sizeL_;  // size of the leading dimension
-  const index_t disp_;   // displacementt od the first element
+  const index_t sizeR_;   // number of rows
+  const index_t sizeC_;   // number of columns
+  const index_t sizeL_;   // size of the leading dimension
+  const index_t stride_;  // internal stride between same row/column elements
```
muhammad-tanvir-1211, line 224: See my comment above regarding `stride` vs `inc`. I think it would be good if we can replace `stride_` with `inc_` here and all other instances where we have used `stride_` as it may be misleading.

test/unittest/extension/omatcopy2_test.cpp
```diff
@@ -0,0 +1,136 @@
...
+  sb_handle.wait(event);
+
+  // Validate the result
+  const bool isAlmostEqual = utils::compare_vectors(B, B_ref);
+  ASSERT_TRUE(isAlmostEqual);
+}
+
+template <typename scalar_t>
+const auto combi = ::testing::Combine(
+    ::testing::Values<char>('n', 't'), ::testing::Values<index_t>(64, 129, 255),
```
muhammad-tanvir-1211, line 121: Maybe we can add comments in front of all these `::testing::Values()` to explain what they do. It's the same as what we do for all other operators.

test/unittest/extension/omatcopy2_test.cpp
```diff
@@ -0,0 +1,136 @@
...
+    ::testing::Values<index_t>(64, 129, 255), ::testing::Values<scalar_t>(0, 2),
+    ::testing::Values<index_t>(1, 7), ::testing::Values<index_t>(1, 3),
+    ::testing::Values<index_t>(1, 7), ::testing::Values<index_t>(1, 3));
+
```
muhammad-tanvir-1211, line 125: It would also be worth adding the `stress tests` for `omatcopy2` as well.

test/unittest/extension/omatcopy_test.cpp
```diff
@@ -0,0 +1,96 @@
...
+      sb_handle.get_queue(), matrix_out, B.data(), size_b);
+  sb_handle.wait(event);
+
+  // Validate the result
+  const bool isAlmostEqual = utils::compare_vectors(B, B_ref);
+  ASSERT_TRUE(isAlmostEqual);
+}
+
+template <typename scalar_t>
+const auto combi = ::testing::Combine(::testing::Values<char>('n', 't'),
```
muhammad-tanvir-1211, line 80: See my comment in the other test file regarding adding `comments` and `stress tests`.

muhammad-tanvir-1211:
Can you also update the operator table in the README to show that the `omatcopy*` operators have been added into the library as part of this PR?

include/interface/extension_interface.h
```diff
@@ -23,10 +23,11 @@
  *
  **************************************************************************/
 
-#ifndef SYCL_BLAS_REDUCTION_INTERFACE_H
-#define SYCL_BLAS_REDUCTION_INTERFACE_H
+#ifndef SYCL_BLAS_EXTENSION_INTERFACE_H
+#define SYCL_BLAS_EXTENSION_INTERFACE_H
 
 #include "operations/extension/reduction.h"
+#include "operations/extension/transpose.h"
```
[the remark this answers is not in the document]
OuadiElfarouki: Sounds like a plan, working on it!

external/cblas/include/cblas.h
```diff
@@ -562,6 +562,16 @@ void cblas_zher2k(const enum CBLAS_ORDER Order, const enum CBLAS_UPLO Uplo,
 
 void cblas_xerbla(int p, const char *rout, const char *form, ...);
 
+// blas-like extensions
+void cblas_somatcopy(const enum CBLAS_ORDER CORDER,
```
pgorlani, line 566: Thanks for your patch, @s-Nick.

Not all the BLAS implementations support these functions, luckily OpenBLAS does. However, CMake looks for the system BLAS library during configuration, which could not support them, e.g., in the case of BLAS reference implementation.

aacostadiaz (approved):
LGTM! I only have two minor comments

include/views/view.h
```diff
@@ -122,7 +122,7 @@ struct VectorView {
 @tparam container_t Type of the container.
  */
 template <typename view_value_t, typename view_container_t,
-          typename view_index_t, typename layout>
+          typename view_index_t, typename layout, bool is_inc = false>
```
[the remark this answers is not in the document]
aacostadiaz: Thanks for addressing this! :D

benchmark/syclblas/extension/omatcopy.cpp
```diff
@@ -0,0 +1,161 @@
...
+  const char* t_str = ts.c_str();
+
+  const auto lda = lda_mul * m;
+  const auto ldb = (*t_str == 't') ? ldb_mul * n : ldb_mul * m;
+
+  const auto size_a = lda * n;
+  const auto size_b = ldb * ((*t_str == 't') ? m : n);
+
+  blas_benchmark::utils::init_level_1_counters<
+      blas_benchmark::utils::Level1Op::copy, scalar_t>(state, 2 * m * n);
```
aacostadiaz, line 55: Add a new Level1 operator for omatcopy

external/cblas/include/cblas.h
s-Nick: Thank you for highlighting it.

As we discussed offline, I implemented an host version for these operators not to force dependencies on a specific BLAS implementation. Changes were commited in 49f48a5

benchmark/syclblas/extension/omatcopy.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+  state.counters["n_fl_ops"] = static_cast<double>(m * n);
```
aacostadiaz, line 57: Move this to a function in `common/include/common/blas1_state_counters.hpp`

aacostadiaz (commented):
Thanks for adding the benchmark.

aacostadiaz (approved):

src/operations/extension/transpose.hpp
```diff
@@ -0,0 +1,221 @@
...
+Transpose<in_place, Tile_size, wg_size, cl_size, local_memory, in_t, out_t,
+          element_t>::eval(cl::sycl::nd_item<1> id) {
+  index_t idx = id.get_global_linear_id();
+
+  index_t in_index, out_index, i_id, j_id;
+
+  get_indices(id, in_index, out_index, i_id, j_id);
+
+  if (i_id < ((M_ - 1) * inc_a_ + 1)) {
+    auto A = A_.get_data().get_pointer();
```
pgorlani, line 120:
```suggestion
    auto A = A_.get_pointer();
```
This should work for both USM and buffer.

*Apply this to all `.get_data()` occurrences.*

test/unittest/extension/omatcopy_test.cpp
```diff
@@ -0,0 +1,108 @@
...
+ *
+ *  @filename omatcopy_test.cpp
+ *
+ **************************************************************************/
+
+#include "blas_test.hpp"
+#include "extension_reference.hpp"
+
+using index_t = int;
+
```
pgorlani, line 30:
```suggestion
```
I think there is no need of this.

*The same applies for the other tests.*

test/unittest/CMakeLists.txt
```diff
@@ -55,6 +55,10 @@ set(SYCL_UNITTEST_SRCS
   ${SYCLBLAS_UNITTEST}/blas3/blas3_gemm_batched_test.cpp
   ${SYCLBLAS_UNITTEST}/blas3/blas3_trsm_test.cpp
   ${SYCLBLAS_UNITTEST}/blas3/blas3_symm_test.cpp
+  # Blas extension
+  ${SYCLBLAS_UNITTEST}/extension/omatcopy_test.cpp
+  ${SYCLBLAS_UNITTEST}/extension/omatcopy2_test.cpp
+  ${SYCLBLAS_UNITTEST}/extension/transpose_test.cpp
```
pgorlani, line 61: I think they should be moved below, inside the `if(${BLAS_ENABLE_EXTENSIONS})` statement, since they belong to the extensions.

OuadiElfarouki: Thanks for highlighting this @pgorlani. We haven't fully included these operators in the "Extensions" group in SYCL BLAS because it doesn't match by definition the BLAS extensions defined by oneMKL spec or cuBLAS/rocBLAS. So far it has been limited to the reduction operator which is not an explicit operator in BLAS extensions as defined in the reference specs. So it's a bit confusing I agree to re-conciliate the extensions option in SYCL BLAS & the recently added extension operators, especially that there is no such flag in other libraries I guess. 
But anyways, we can put them all together (and disable their generation/build alongside tests & benchs as it's done with reduction) or separate the reduction operator flag if it has any property different from the extension ops I'm not aware of.

benchmark/syclblas/extension/omatcopy.cpp
```diff
@@ -0,0 +1,155 @@
+/**************************************************************************
```
pgorlani, line 1: What about the `omatcopy2` benchmark? Don't we need it?

test/unittest/extension/transpose_test.cpp
```diff
@@ -0,0 +1,134 @@
...
+void Transpose(const T* in, const index_t& ld_in, T* out, const index_t& ld_out,
+               const index_t& M, const index_t& N) {
+  for (index_t i = 0; i < M; i++) {
+    for (index_t j = 0; j < N; j++) {
+      out[j + i * ld_out] = in[i + j * ld_in];
+    }
+  }
+}
+
+}  // namespace reference_blas
```
pgorlani, line 63: Could we move it to `test/unittest/extension/extension_reference.hpp`?

test/unittest/extension/extension_reference.hpp
```diff
@@ -0,0 +1,94 @@
...
+ * @param trans tranpose input matrix or not
+ * @param m,n matrices dimensions
+ * @param alpha scalar factor for input matrix
+ * @param A input matrix
+ * @param lda leading dimension of input matrix
+ * @param B output matrix
+ * @param ldb leading dimension of output matrix
+ */
+template <typename index_t, typename scalar_t>
+void omatcopy_ref(char trans, const index_t m, const index_t n,
```
pgorlani, line 43: Do we need to append `_ref` since it already in the `reference_blas` namespace?

test/unittest/extension/transpose_test.cpp
```diff
@@ -0,0 +1,134 @@
...
+ *
+ *  @filename transpose.cpp
+ *
+ **************************************************************************/
+
+#include "blas_test.hpp"
+
+using index_t = int;
+
+enum trans_type : int { Inplace = 0, Outplace = 1 };
```
pgorlani, line 30: Probably, it is convenient to use a `char` (e.g. `i` and `o`) rather than an `enum`. This allows to print out a more understandable test-name, `tr_type_{0,1}` vs `tr_type_{i,o}`.

Moreover, in this way, you can also get rid of `dump_arg`.

test/unittest/extension/omatcopy_test.cpp
s-Nick: yes, it also breaks the test if `index_t` is not `int`. I will update it

benchmark/syclblas/extension/omatcopy.cpp
s-Nick: These benchmarks have been under discussion for some time. We decided to include benchmarks for `omatcopy` and other operators in the PR that builds upon it because cuBLAS and rocBLAS have a similar operator. However, we later realized that comparing `omatcopy` may not be appropriate due to slight differences in the operator. Therefore, we also omitted it for cuBLAS and rocBLAS.

As for `omatcopy2`, it is different from the others because there is no suitable point of comparison. Hence, we did not include it. Should we consider its inclusion?

src/interface/extension_interface.hpp
```diff
@@ -46,6 +50,61 @@ struct get_second_step_op<MeanOperator> {
   using type = AddOperator;
 };
 
+/**
+ * @brief Implementation of matrix copy operators for transpose cases with out
+ * of place memory copy.
+ */
+template <bool in_place, bool trans, typename sb_handle_t, typename element_t,
+          typename index_t, typename in_t, typename out_t>
+typename std::enable_if<trans && !in_place, typename sb_handle_t::event_t>::type
```
pgorlani, line 59: Did you consider unifying the `_matcopy_impl` and using `if` (this is normally done in `src/interface/blas2_interface.hpp`) rather than `enable_if`?

test/unittest/CMakeLists.txt
s-Nick: I agree with @OuadiElfarouki that we probably need another flag to split these extensions and the one already present, but we can start moving them all together and separate the two later.

benchmark/cublas/CMakeLists.txt
```diff
@@ -69,6 +69,8 @@ set(sources
   blas3/trsm.cpp
   blas3/trsm_batched.cpp
   blas3/trmm.cpp
+  # extension
+  extension/omatcopy.cpp
 ) 
 
 #if(${BLAS_ENABLE_EXTENSIONS})
```
pgorlani, line 76: At this point, I think is possible to remove these commented lines.

benchmark/syclblas/extension/omatcopy.cpp
pgorlani: Probably, we should include it because it is present in oneMKL.

benchmark/syclblas/CMakeLists.txt
```diff
@@ -63,10 +63,13 @@ set(sources
...
   blas3/trsm.cpp
   blas3/symm.cpp
+  # blas Extension
+  extension/omatcopy.cpp
+  extension/omatcopy2.cpp
 )
 
 if(${BLAS_ENABLE_EXTENSIONS})
-  list(APPEND sources "extension/reduction.cpp")
+  set(sources extension/reduction.cpp)
```
pgorlani, line 72:
```suggestion
  list(APPEND sources extension/reduction.cpp)
```
Otherwise, the other benchmarks get removed.

pgorlani (approved):

PR #430, opened 2023-05-31
Removed RCAR and ARM Mali backends

muhammad-tanvir-1211:
This PR removed the RCAR and ARM Mali backends from the library.


OuadiElfarouki (approved):
Looks neat!

pgorlani (approved):

PR #431, opened 2023-05-31
Add backend label for benchmarks

aacostadiaz:
This pull request adds a label to benchmarks to indicate the backend that is being used.


common/include/common/set_benchmark_label.hpp
```diff
@@ -143,7 +145,7 @@ inline void set_label(benchmark::State& state,
 
 template <typename scalar_t>
 inline void set_benchmark_label(benchmark::State& state,
-                                const cl::sycl::queue& q) {
+                                const cl::sycl::queue& q, BackendType backend) {
```
pgorlani, line 148: Thanks for your patch, @aacostadiaz.

Probably, it is better to create a specific `set_benchmark_label` for each backend type. Also considering the fact that for `cublas` and `rocblas` it is not possible to use a SYCL queue for querying the device name (you can use `{cuda,hip}GetDeviceProperties` instead).

muhammad-tanvir-1211: These changes have been addressed in [9550dff](https://github.com/codeplaysoftware/sycl-blas/pull/431/commits/9550dff64fdfba4d35c74e00f6abacd6db2764aa) and [739f91d](https://github.com/codeplaysoftware/sycl-blas/pull/431/commits/739f91d3a0f9d49ed3c379d1282f35b8846b0438). Thanks for detecting this.

s-Nick (approved):
LGTM! Thanks

PR #433, opened 2023-06-05
Added USM support for TPMV Level 2 BLAS Operator

muhammad-tanvir-1211:


benchmark/syclblas/blas2/tpmv.cpp
```diff
@@ -49,15 +51,23 @@ void run(benchmark::State& state, blas::SB_Handle* sb_handle_ptr,
...
   // Input matrix/vector, output vector.
   std::vector<scalar_t> m_a =
       blas_benchmark::utils::random_data<scalar_t>(((n + 1) * n) / 2);
   std::vector<scalar_t> v_x =
       blas_benchmark::utils::random_data<scalar_t>(xlen);
 
-  auto m_a_gpu = blas::make_sycl_iterator_buffer<scalar_t>(m_a, m_a.size());
-  auto v_x_gpu = blas::make_sycl_iterator_buffer<scalar_t>(v_x, xlen);
+  auto m_a_gpu = blas::helper::allocate<mem_alloc, scalar_t>(m_a.size(), q);
+  auto v_x_gpu = blas::helper::allocate<mem_alloc, scalar_t>(xlen, q);
```
OuadiElfarouki, line 63: Thanks for the TPMV USM extension! 
This is just a minor comment regarding variables affected by `incX` such as the size of `v_x`, it might be relevant to include it in the formulation of the size even-though it's hard-coded to 1, in order to avoid eventual confusion later if smth changes.

src/interface/blas2_interface.hpp
```diff
@@ -728,10 +731,21 @@ typename sb_handle_t::event_t _tpmv_impl(sb_handle_t& sb_handle, index_t _N,
...
-      static_cast<index_t>(local_range_x * (loc_mem_leading_dim + 2)));
+      static_cast<index_t>(local_range_x * (loc_mem_leading_dim + 2)),
+      _dependencies);
 
   auto assignOp = make_op<Assign>(vx, vres);
-  return concatenate_vectors(tpmvEvent, sb_handle.execute(assignOp));
+  auto ret =
+      concatenate_vectors(tpmvEvent, sb_handle.execute(assignOp, tpmvEvent));
+
+#ifdef DEFAULT_CPU
```
pgorlani, line 741: Thanks for your patch, @muhammad-tanvir-1211.

It looks good to me. As discuss earlier, we should deallocate the temporary arrays also for the other backends in order to avoid memory leaks.

pgorlani (approved):

src/interface/blas2_interface.hpp
mehdi-goli: Yes, otherwise we get memory leak. You can create a issue track in dpcpp for this error. Also you if profiling is enabled you can query the queue and do wait on dependencies here and deallocate the memory, otherwise do what you are doing. And remove the default_CPU as it should be for all the cases.

benchmark/syclblas/blas2/tpmv.cpp
muhammad-tanvir-1211: Hi @OuadiElfarouki 
I have addressed this in [6417e3a363](https://github.com/codeplaysoftware/sycl-blas/pull/433/commits/6417e3a363ee90073af935a40e9a1c66c3a7e07a).

src/interface/blas2_interface.hpp
muhammad-tanvir-1211: Hi @mehdi-goli and @pgorlani,
I am going to address this in another PR as this change is going to effect multiple operators. Thanks.

muhammad-tanvir-1211: This change has been addressed in https://github.com/codeplaysoftware/sycl-blas/pull/436

aacostadiaz (approved):
LGTM

OuadiElfarouki (approved):

PR #435, opened 2023-06-06
Adjust benchmark parameters

aacostadiaz:
This PR modify the benchmark parameters to: 

- Increase problem size to enable meaningful GPU benchmarks.
- Reduce variations to ensure timely completion of all benchmarks.


common/include/common/common_utils.hpp
```diff
@@ -523,35 +502,17 @@ get_gemm_batched_strided_params(Args& args) {
     warning_no_csv();
     std::vector<gemm_batched_strided_param_t<scalar_t>>
         gemm_batched_strided_default;
-    constexpr index_t dmin = 64, dmax = 1024;
+    constexpr index_t dmin = 1024, dmax = 8192;
     constexpr index_t stride_a_mul_min = 0, stride_a_mul_max = 2;
     constexpr index_t stride_b_mul_min = 0, stride_b_mul_max = 2;
     constexpr index_t stride_c_mul_min = 1, stride_c_mul_max = 2;
```
OuadiElfarouki, line 508: stride multipliers variables can be removed in this case.

common/include/common/common_utils.hpp
```diff
@@ -237,7 +237,7 @@ static inline std::vector<blas1_param_t> get_blas1_params(Args& args) {
   if (args.csv_param.empty()) {
     warning_no_csv();
     std::vector<blas1_param_t> blas1_default;
-    for (index_t size = 4096; size <= 1048576; size *= 2) {
+    for (index_t size = 1048576; size <= 32 * 1048576; size *= 2) {
```
OuadiElfarouki, line 240: This brings up the number of generated sizes to only 5, is this what we want or it doesn't matter?

common/include/common/common_utils.hpp
```diff
@@ -597,12 +558,10 @@ get_trsm_batched_params(Args& args) {
...
         for (char trans : {'n', 't'}) {
           for (char diag : {'u', 'n'}) {
-            for (index_t m = dmin; m <= dmax; m *= 2) {
-              for (index_t n = dmin; n <= dmax; n *= 2) {
-                trsm_batched_default.push_back(
-                    std::make_tuple(side, uplo, trans, diag, m, n, alpha,
-                                    batch_size, stride_a_mul, stride_b_mul));
-              }
+            for (index_t n = dmin; n <= dmax; n *= 2) {
+              trsm_batched_default.push_back(
```
muhammad-tanvir-1211, line 562: I don't think this `trsm_batched` is going to be used anywhere, since this operator is not available in SYCL-BLAS at the moment. Could you please double check this?

muhammad-tanvir-1211: Disregard this comment please. I forgot that we use `trsm_batched` for `cublas/rocblas` benchmarks.

pgorlani:
Thanks for your patch, @aacostadiaz.

I suggest updating this README [section](https://github.com/codeplaysoftware/sycl-blas/blob/6ec4f00e654570dc21aa00f61d711f46751db829/benchmark/README.md?plain=1#L183) accordingly.

common/include/common/common_utils.hpp
```diff
@@ -805,20 +711,13 @@ static inline std::vector<gbmv_param_t<scalar_t>> get_gbmv_params(Args& args) {
   if (args.csv_param.empty()) {
     warning_no_csv();
     std::vector<gbmv_param_t<scalar_t>> gbmv_default;
-    constexpr index_t dmin = 64, dmax = 1024;
-    constexpr index_t kmin = 1;
+    constexpr index_t dmin = 1024, dmax = 16384;
+    constexpr index_t kmin = 2048;
```
pgorlani, line 715: I think that `kmin < dmin`, see
https://github.com/codeplaysoftware/sycl-blas/blob/6ec4f00e654570dc21aa00f61d711f46751db829/src/interface/blas2_interface.hpp#L478-L480

I suggest setting the minimum `k` as a function of the actual problem size, e.g., `m/32`

*Same applies for `tbmv` and `sbmv`.*

common/include/common/common_utils.hpp
```diff
@@ -237,7 +237,7 @@ static inline std::vector<blas1_param_t> get_blas1_params(Args& args) {
   if (args.csv_param.empty()) {
     warning_no_csv();
     std::vector<blas1_param_t> blas1_default;
-    for (index_t size = 1048576; size <= 67108864; size *= 2) {
+    for (index_t size = 1048576; size <= 134217728; size *= 2) {
```
pgorlani, line 240: Since we are working with large positive integer numbers that are power-of-two, we could represent them using the shift operator for increasing clarity, e.g., `1l<<20`.

common/include/common/common_utils.hpp
aacostadiaz: I updated the ranges. Now we run 8 different samples, and Google Benchmark will run multiple iterations for each sample until statistically significant results are obtained.

common/include/common/common_utils.hpp
```diff
@@ -717,17 +682,12 @@ static inline std::vector<syrk_param_t<scalar_t>> get_syrk_params(Args& args) {
...
       for (char trans : {'n', 't'}) {
-        for (index_t n = dmin; n <= dmax; n *= 2) {
-          for (index_t k = dmin; k <= dmax; k *= 2) {
-            for (scalar_t beta : {0, 1}) {
-              syrk_default.push_back(
-                  std::make_tuple(uplo, trans, n, k, alpha, beta));
-            }
-          }
+        for (index_t n = dmin; n <= dmax; n *= 4) {
+          syrk_default.push_back(std::make_tuple(uplo, trans, n, n, alpha, 0));
```
s-Nick, line 690: Thank you for you patch @aacostadiaz.
For this benchmark's parameter I see we are not testing for `beta != 0`. I think this may reduce the number of operations carried out by the routine, I suggest you to set `beta` to 1 as it is for `alpha`

common/include/common/common_utils.hpp
```diff
@@ -805,19 +763,13 @@ static inline std::vector<gbmv_param_t<scalar_t>> get_gbmv_params(Args& args) {
...
-        for (index_t n = dmin; n <= dmax; n *= 2) {
-          for (index_t kl = kmin; kl <= m / 4; kl *= 2) {
-            for (index_t ku = kmin; ku <= n / 4; ku *= 2) {
-              gbmv_default.push_back(
-                  std::make_tuple(t, m, n, kl, ku, alpha, beta));
-            }
-          }
+      for (index_t m = dmin; m <= dmax; m *= 4) {
+        for (index_t kl = m / 32; kl <= m / 4; kl *= 2) {
+          gbmv_default.push_back(std::make_tuple(t, m, m, kl, kl, alpha, beta));
```
s-Nick, line 772: Same thought about `beta = 0` as above

common/include/common/common_utils.hpp
```diff
@@ -423,15 +421,15 @@ static inline std::vector<blas3_param_t<scalar_t>> get_blas3_params(
   if (args.csv_param.empty()) {
     warning_no_csv();
     std::vector<blas3_param_t<scalar_t>> blas3_default;
-    constexpr index_t dmin = 64, dmax = 1024;
+    constexpr index_t dmin = 32, dmax = 8192;
     std::vector<std::string> dtranspose = {"n", "t"};
     scalar_t alpha = 1;
     scalar_t beta = 0;
```
OuadiElfarouki, line 427: All looks good to me, I was just wondering if we want to keep `beta = 0` here or change it to 1 as we did for level 2?

aacostadiaz: Ooh, I missed that one, sorry. It is fixed now. Thanks for bringing it up.

mehdi-goli (approved):

PR #446, opened 2023-06-29
Add omatadd_batch operator

s-Nick:
This patch add the blas extension operator omatadd_batch.

Co-authored-by: @OuadiElfarouki [ouadie.elfarouki@codeplay.com](mailto:ouadie.elfarouki@codeplay.com)


OuadiElfarouki:
Squashed all commits history for convenience when merging. PR still stacked on top of [matcopy_batch_pr](https://github.com/codeplaysoftware/sycl-blas/tree/matcopy_batch_pr)

README.md
```diff
@@ -321,14 +321,17 @@ matrices in a batched entry for inputs/outputs A, B and C.
 * `M` and `N` are the dimensions of the matrices.
 * `alpha` and `beta` are scaling scalars.
 * `batch_size` is the number of batch matrices.
+* `inc_a` and `inc_b` are integers. The distance between element in the same column.
 
 | operation | arguments | description |
 |---|---|---|
 | `_omatcopy` | `sb_handle`, `transa`, `M`, `N`, `alpha`, `A`, `lda`, `B`, `ldb`  | Perform an out-of-place scaled matrix transpose or copy operation using a general dense matrix. |
 | `_omatcopy2`| `sb_handle`, `transa`, `M`, `N`, `alpha`, `A`, `lda`, `inc_a`, `B`, `ldb`, `inc_b`  | Computes two-strided scaling and out-of-place transposition or copying of general dense matrices. |
+| `_omatcopy_batch` | `sb_handle`, `transa`, `M`, `N`, `alpha`, `A`, `lda`, `stride_a`, `B`, `ldb`, `stride_b`, `batch_size` | Perform an out-of-place scaled batched-strided matrix transpose or copy operation using a general dense matrix. |
```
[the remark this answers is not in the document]
s-Nick: I noticed that too, I just removed it. Thank you @muhammad-tanvir-1211

test/unittest/extension/transpose_test.cpp
```diff
@@ -26,6 +26,9 @@
 #include "blas_test.hpp"
 #include "extension_reference.hpp"
 
+using index_t = int;
+enum trans_type : int { Inplace = 0, Outplace = 1 };
```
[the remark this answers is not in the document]
s-Nick: I don't think it is. We switched to use `char` instead of this one. @OuadiElfarouki, can you confirm?

muhammad-tanvir-1211 (approved):
LGTM, Thank you for addressing all the comments.

src/interface/extension_interface.hpp
```diff
@@ -445,24 +477,79 @@ typename sb_handle_t::event_t _omatadd(
     element_t alpha, container_0_t a, index_t lda, element_t beta,
     container_1_t b, index_t ldb, container_2_t c, index_t ldc,
     const typename sb_handle_t::event_t& _dependencies) {
+  // bail out early if the leading dimensions are not correct
+  if (ldc < m || lda < (trans_a == 't' ? n : m) ||
+      ldb < (trans_b == 't' ? n : m)) {
+    typename sb_handle_t::event_t ret;
+    return ret;
+  }
```
pgorlani, line 485: Probably it is better to make aware the user that there is an error with an exception, otherwise it is totally silent.

test/unittest/extension/omatadd_batched_test.cpp
```diff
@@ -0,0 +1,191 @@
...
+  index_t m, n, ld_a_mul, ld_b_mul, ld_c_mul, stride_a_m, stride_b_m,
+      stride_c_m, batch_size;
+  scalar_t alpha, beta;
+
+  std::tie(alloc, trans_a, trans_b, m, n, alpha, beta, ld_a_mul, ld_b_mul,
+           ld_c_mul, stride_a_m, stride_b_m, stride_c_m, batch_size) = combi;
+
+  if ((trans_a == 'n') && (trans_b == 'n')) {
+    return;
+  }
```
pgorlani, line 47: Please, could you state the reason for skipping silently the test in this case?

OuadiElfarouki: @pgorlani this was actually left from the early tests as we focused mainly on the transposed cases, as the non transposed ones became trivial at a certain point in time. Will remove this skip though as it is still a case that needs to be tested!

src/interface/extension_interface.hpp
OuadiElfarouki: will add a `throw std::invalid_argument ` here

common/include/common/common_utils.hpp
```diff
@@ -1199,49 +1255,58 @@ static inline std::vector<omatadd_param_t<scalar_t>> get_omatadd_params(
...
+        for (index_t m = dmin; m <= dmax; m *= 2) {
+          for (index_t n = dmin; n <= dmax; n *= 2) {
+            for (index_t lda_mul = 1; lda_mul < 2; ++lda_mul) {
+              for (index_t ldb_mul = 1; ldb_mul < 2; ++ldb_mul) {
+                for (index_t ldc_mul = 1; ldc_mul < 2; ++ldc_mul) {
+                  omatadd_batch_default.push_back(
+                      std::make_tuple(trans_a, trans_b, m, n, alpha, beta,
+                                      lda_mul, ldb_mul, ldc_mul, stride_a_mul,
+                                      stride_b_mul, stride_c_mul, batch_size));
+                }
```
pgorlani, line 1287: I think you need to exclude the configurations that do not fit the conditions for `ld{a,b}`. I'm getting `Invalid lda` for `bench_omatadd_batched`.

OuadiElfarouki: Thanks a lot for pointing to this! Actually it was caused by a typo in the `omatadd_batched.cpp` benchmark files itself, it computed the `lda` based on `trans_b` and now it's fixed. The generated configurations would not cause any error as they only contain leading dimension multipliers(>=1) and will thus always result in valid ld relatively to the m & n parameters.

pgorlani (approved):

PR #434, opened 2023-06-05
Add TPSV, refactor TRSV and TBSV Level 2 BLAS Operators

pgorlani:
This patch adds an implementation of `TPSV` Level 2 operator compatible with `TUNING_TARGET` specializations for `INTEL_GPU`, `AMD_GPU` and `NVIDIA_GPU`.

Moreover, this patch refactors and unifies the implementations of `TRSV` and `TBSV` with `TPSV`.

PR #447, opened 2023-07-05
Update cmake to support hipSYCL

s-Nick:
This PR resolves a compilation error that occurs when targeting NVIDIA GPUs and using hipSYCL as the SYCL compiler. Additionally, it re-enables benchmarking with hipSYCL.

It's important to note that the fix for this compiler issue with all targets will be complete once PR #434 is merged. This is because PR #434 includes essential safeguards to prevent compilation failures, which are necessary due to a missing feature in hipSYCL.


aacostadiaz (approved):
Great job on discovering and fixing the issue!

PR #449, opened 2023-07-17
Update `gemv` tuning target parameters

pgorlani:
This patch adjusts the `gemv` tuning target parameters and removes unnecessary headers for the generation of matrix-vector multiplication routines.


aacostadiaz (approved):
Ship it!

PR #450, opened 2023-07-28
Fixes issue with operators using negative increments with USM

aacostadiaz:
This PR fixes an issue with operators that use negative increments with USM. Additionally, the test framework has been updated to allow for tests with negative increments.

### DIFF
--- a/include/views/view.h
+++ b/include/views/view.h
@@ -97,13 +97,17 @@ struct VectorView {
   template <bool use_as_ptr = false>
   SYCL_BLAS_INLINE typename std::enable_if<!use_as_ptr, value_t &>::type eval(
       index_t i) {
-    return (strd_ == 1) ? *(data_ + i) : *(data_ + i * strd_);
+    return (strd_ == 1)  ? *(data_ + i)
+           : (strd_ > 0) ? *(data_ + i * strd_)
+                         : *(data_ + (size_ * -strd_) + ((i + 1) * strd_));
   }
 
   template <bool use_as_ptr = false>
   SYCL_BLAS_INLINE typename std::enable_if<!use_as_ptr, value_t>::type eval(
       index_t i) const {
-    return (strd_ == 1) ? *(data_ + i) : *(data_ + i * strd_);
+    return (strd_ == 1)  ? *(data_ + i)
+           : (strd_ > 0) ? *(data_ + i * strd_)
+                         : *(data_ + (size_ * -strd_) + ((i + 1) * strd_));
   }
 
   SYCL_BLAS_INLINE value_t &eval(cl::sycl::nd_item<1> ndItem) {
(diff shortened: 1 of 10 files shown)


include/views/view.h
```diff
@@ -97,13 +97,17 @@ struct VectorView {
   template <bool use_as_ptr = false>
   SYCL_BLAS_INLINE typename std::enable_if<!use_as_ptr, value_t &>::type eval(
       index_t i) {
-    return (strd_ == 1) ? *(data_ + i) : *(data_ + i * strd_);
+    return (strd_ == 1)  ? *(data_ + i)
+           : (strd_ > 0) ? *(data_ + i * strd_)
+                         : *(data_ + (size_ * -strd_) + ((i + 1) * strd_));
```
muhammad-tanvir-1211, line 102: Maybe, we could add `size_ * -strd_` to the pointer at the construction of the of `VectorView` once or store this value in the `struct` as an `offset` variable rather than performing this multiplication every time we call the `eval` function.

muhammad-tanvir-1211 (commented):
Thank you for finding and fixing this bug @aacostadiaz.

include/views/view.h
mehdi-goli: This will introduce two nested if statement which affect the performance.

include/views/view.h
```diff
@@ -97,13 +97,17 @@ struct VectorView {
...
   }
 
   template <bool use_as_ptr = false>
   SYCL_BLAS_INLINE typename std::enable_if<!use_as_ptr, value_t>::type eval(
       index_t i) const {
-    return (strd_ == 1) ? *(data_ + i) : *(data_ + i * strd_);
+    return (strd_ == 1)  ? *(data_ + i)
+           : (strd_ > 0) ? *(data_ + i * strd_)
```
mehdi-goli, line 109: same here

test/unittest/CMakeLists.txt
```diff
@@ -30,6 +30,7 @@ include_directories(${SYCLBLAS_TEST} ${BLAS_INCLUDE_DIRS})
 # compiling tests
 set(SYCL_UNITTEST_SRCS
   # Blas 1 tests
+  ${SYCLBLAS_UNITTEST}/blas1/blas1_asum_test.cpp
```
muhammad-tanvir-1211, line 33: Do we want to enable this test given that we are not enabling the rest?

muhammad-tanvir-1211 (approved):

test/unittest/CMakeLists.txt
aacostadiaz: The more tests we enable, the better. I'm enabling asum because I used it to fix the issue with negative increments, and we want to keep a test checking that it works properly. For the rest, we should enable all of them.

pgorlani (approved):

PR #451, opened 2023-07-28
Fixes issue with dependencies in nrm2

aacostadiaz:
This PR fixes an issue with dependencies in nrm2. Previously, the operations could finish without computing the sqrt operation. This was because the dependency tree was not correctly constructed. This PR fixes the issue by correctly constructing the dependency tree, which ensures that all operations are computed in the correct order.

### DIFF
--- a/test/unittest/blas1/blas1_nrm2_test.cpp
+++ b/test/unittest/blas1/blas1_nrm2_test.cpp
@@ -83,8 +83,6 @@ void run_test(const combination_t<scalar_t> combi) {
     out_s = _nrm2(sb_handle, size, gpu_x_v, incX, {copy_x});
   }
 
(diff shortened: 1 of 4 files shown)


test/unittest/blas1/blas1_nrm2_test.cpp
```diff
@@ -71,6 +83,8 @@ void run_test(const combination_t<scalar_t> combi) {
     out_s = _nrm2(sb_handle, size, gpu_x_v, incX, {copy_x});
   }
 
+  std::cout << "SYCL " << out_s << " vs ref " << out_cpu_s << std::endl;
```
muhammad-tanvir-1211, line 86: Do we need this `cout` statement here?

PR #459, opened 2023-08-31
Fix building with different BLAS_INDEX_TYPES

s-Nick:
Currently if trying to build with `BLAS_INDEX_TYPES` different from `int` or with multiple values for the index type i.e. `int;long` the compilation fails.
This PR addresses this problem.


muhammad-tanvir-1211 (approved):
LGTM, Thanks.

pgorlani (approved):

PR #460, opened 2023-08-31
Add Benchmark CSV files for Language Models

muhammad-tanvir-1211:
This PR adds the benchmark csv files for `gemm` and `gemm_batched_strided` for some of the latest `transformer` based language models. The oneDNN matmul inputs for language models are used as reference for these csv files. The reference files can be found here: https://github.com/oneapi-src/oneDNN/tree/master/tests/benchdnn/inputs/matmul

For all cases, the oneDNN inputs use `bias` as an extra input to `gemm` but I have omitted this argument from the csv files as this argument is not yet supported in portBLAS `gemm`.


OuadiElfarouki (approved):
Looks good, thanks for gathering all these!

s-Nick (approved):
LGTM, Thank you @muhammad-tanvir-1211

PR #462, opened 2023-09-11
Enabed Complex data type for Gemm

OuadiElfarouki:
This PR extends support to sycl::complex<float/double> data types to GEMM operators (gemm & batched extensions).
Main changes : 
- Added a wrapper around sycl::complex data type to be used within fully & partially vectorized gemm kernels to mimic sycl::vec (as we don't have an implementation for it in DPCPP yet). 
- Changes so far rely on the experimental support of sycl::complex in sycl/llvm branch, which is used through the extension header : `ext/oneapi/experimental/sycl_complex.hpp` 
- Gemm configurations for complex type are put separately from float/double types as they might expect custom optimal values (to be tuned later once benchmarking is ready)
- Added tests for GEMM & batched derivatives, tested passing on intel, amd and Nvidia GPUs as well as intel CPU.


include/blas_meta.h
```diff
@@ -190,6 +195,29 @@ struct is_sycl_scalar<float *> : std::false_type {};
...
+
+template <class type>
+struct is_complex_sycl
+    : std::integral_constant<bool,
+                             std::is_same_v<type, complex_sycl<double>> ||
+                                 std::is_same_v<type, complex_sycl<float>>> {};
+
+// STD Complex type alias
+template <typename T>
+using complex_std = typename std::complex<T>;
```
s-Nick, line 211: It seems to me that redefine `std::complex` to `complex_std` doesn't have any particular utility. It is used only in `common_utils.hpp` and tests, but you can use  directly `std::complex`. Am I missing something? Do you have any particular reason in mind?

common/include/common/common_utils.hpp
```diff
@@ -1307,6 +1307,37 @@ static inline std::vector<scalar_t> random_data(size_t size) {
   return v;
 }
 
+#ifdef BLAS_ENABLE_COMPLEX
+template <typename scalar_t>
+static inline complex_std<scalar_t> random_scalar() {
+  scalar_t rl = 1e-3 * ((rand() % 2000) - 1000);
+  scalar_t im = 1e-3 * ((rand() % 2000) - 1000);
+  return complex_std<scalar_t>({rl, im});
```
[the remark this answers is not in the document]
OuadiElfarouki: Indeed! It even caused issues in some places, will fix these thanks!

include/blas_meta.h
OuadiElfarouki: Hmm yes indeed it looks like I forgot to reuse it in the tests (where i was planning to), and I'm reusing it heavily in benchmarks. Idk I was thinking about having it distinguishable from sycl::complex that's why. But yes I'll take note of it and discuss with the rest of the team if it's worth keeping!

src/operations/blas1_trees.hpp
```diff
@@ -108,6 +108,17 @@ struct DetectScalar<std::complex<double>> {
   static element_t get_scalar(element_t &scalar) { return scalar; }
 };
 
+#ifdef BLAS_ENABLE_COMPLEX
+/*! DetectScalar (for sycl::complex<value_t>)
+ * @brief See Detect Scalar.
+ */
+template <typename value_t>
+struct DetectScalar<complex_sycl<value_t>> {
```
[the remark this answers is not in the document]
OuadiElfarouki: yes `sycl::complex` one is needed as it is actually the one used in kernels, not the std one.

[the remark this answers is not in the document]
muhammad-tanvir-1211: Can we remove the other one than? Or do we need that for some other part of the code?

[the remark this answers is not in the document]
OuadiElfarouki: Will double check it & delete it if needed, as there are a couple of complex related utilities _(namely this DetectScalar as well as constant scalar partial-specialization)_ that have been implemented a long time ago and don't seem to be used anywhere.

src/operations/blas3/gemm_load_store.hpp
```diff
@@ -186,5 +186,149 @@ struct Packetize {
   }
 };
 
+#ifdef BLAS_ENABLE_COMPLEX
```
muhammad-tanvir-1211, line 189: I think it's best if we move this to a separate file to avoid confusion between the normal load/store.

test/unittest/blas3/blas3_gemm_common.hpp
```diff
@@ -383,4 +396,327 @@ static std::string generate_batched_strided_name(
...
+  for (int i = 0; i < batch; ++i) {
+    reference_blas::cgemm<scalar_t>(
+        ta_str, tb_str, m, n, k, reinterpret_cast<const void*>(&alpha),
+        reinterpret_cast<const void*>(a_m.data() + i * size_a + offset), lda,
+        reinterpret_cast<const void*>(b_m.data() + i * size_b + offset), ldb,
+        reinterpret_cast<const void*>(&beta),
+        reinterpret_cast<void*>(c_m_cpu.data() + i * size_c + offset), ldc);
+  }
+
+  if (batch > 1 && batch_type == gemm_batch_type_t::interleaved) {
```
muhammad-tanvir-1211, line 459: We could move this if condition to before the call to `reference_blas` to avoid running the `cgemm` unnecessarily.

src/operations/blas3/gemm_load_store.hpp
OuadiElfarouki: Moved the complex section to` gemm_load_store_complex.hpp`

test/unittest/blas3/blas3_gemm_common.hpp
OuadiElfarouki: Yes done!

muhammad-tanvir-1211 (approved):

PR #467, opened 2023-09-28
Fixed get_name in cublas and rocblas benchmarks

muhammad-tanvir-1211:


muhammad-tanvir-1211:
These changes have already been addressed in #446

PR #470, opened 2023-10-05
Fix gemm bug for `sm_80` `NVIDIA_GPU` target

pgorlani:
For `sm_80`, global memory data is `round_to_tf32` in any case even if a non-joint matrix version of `_gemm` is selected. This rounding causes test failings.

This patch splits the class for loading global memory values for `gemm_local` and `gemm_local_joint_matrix` fixing the issue.


muhammad-tanvir-1211 (approved):
LGTM, Thanks.

OuadiElfarouki (approved):

PR #471, opened 2023-10-12
Refactored blas1 dot & sdsdot operators

OuadiElfarouki:
This PR follows the refactoring approach of reduction based operators _(asum, nrm2)_ for the dot _(& sdsdot)_, namely the reduction through the `sycl::reduce_over_group` encapsulated in the blas::WGAtomicReduction kernel. 
Also, per-target device tuned _wg size_ & _local memory_ configurations are added for the dot operator to be selected depending on tuning target (at compile time) & input size (at run-time). 
The changes affect the `dot` and `sdsdot` **async** operators directly and their **synchronous** _(with output return)_ versions indirectly. The async version of the two operators API **expect** the input/output to be initialized to zero, while the sync version handles the initialization to zero internally.


src/interface/blas1_interface.hpp
```diff
@@ -764,16 +813,21 @@ typename ValueType<container_0_t>::type _sdsdot(
     const typename sb_handle_t::event_t &_dependencies) {
   constexpr bool is_usm = std::is_pointer<container_0_t>::value;
   using element_t = typename ValueType<container_0_t>::type;
-  element_t res{};
+  element_t res = element_t(0);
```
s-Nick, line 816: For future reference. I don't think this change is necessary. `element_t res{}` would initialize to zero or in any case `element_t res{0}` would do it.

test/unittest/blas1/blas1_dot_test.cpp
```diff
@@ -26,7 +26,8 @@
 #include "blas_test.hpp"
 
 template <typename scalar_t>
-using combination_t = std::tuple<std::string, api_type, int, int, int>;
+using combination_t =
+    std::tuple<std::string, api_type, int, int, int, scalar_t>;
```
s-Nick, line 30: Could you use `index_t` instead of `int` here? I think that keep it generic would help us to enable tests for different index type in the future, if necessary.

src/interface/blas1_interface.hpp
```diff
@@ -165,16 +151,22 @@ typename sb_handle_t::event_t _sdsdot(
...
                              static_cast<index_t>(1));
-
-  dot_event =
-      internal::_dot(sb_handle, _N, _vx, _incx, _vy, _incy, _rs, _dependencies);
-  auto addOp = make_op<ScalarOp, AddOperator>(sb, rs);
-  auto assignOp2 = make_op<Assign>(rs, addOp);
-  auto ret2 = sb_handle.execute(assignOp2, dot_event);
-  return blas::concatenate_vectors(dot_event, ret2);
+  if (!_N) {
+    auto addOp = make_op<ScalarOp, AddOperator>(sb, rs);
```
s-Nick, line 157: if `_N` is zero this is a copy operation. Did you try to copy directly the value instead of adding and then assigning it? Would make sense to try it?

test/unittest/blas1/blas1_dot_test.cpp
OuadiElfarouki: hmm fair enough but we usually don't template the test parameter combination_t with the `index_t`, I think it's better to have it in a separate PR in case we want to as it has to be reflected on all other operators tests.

s-Nick (approved):
Thank you for addressing all the comments, LGTM!

src/interface/blas1_interface.hpp
```diff
@@ -165,16 +151,21 @@ typename sb_handle_t::event_t _sdsdot(
...
-
-  dot_event =
-      internal::_dot(sb_handle, _N, _vx, _incx, _vy, _incy, _rs, _dependencies);
-  auto addOp = make_op<ScalarOp, AddOperator>(sb, rs);
-  auto assignOp2 = make_op<Assign>(rs, addOp);
-  auto ret2 = sb_handle.execute(assignOp2, dot_event);
-  return blas::concatenate_vectors(dot_event, ret2);
+  if (!_N) {
+    sb_handle.wait(_dependencies);
+    auto ret = blas::helper::copy_to_device(sb_handle.get_queue(), &sb, _rs, 1);
```
muhammad-tanvir-1211, line 156: The copy_to_device function definition assumes that source and destination containers have the same data type, so this copy introduces an error when `_rs` is of type `double`.

OuadiElfarouki: Thanks for the highlight @muhammad-tanvir-1211, I've fixed the issue with a `reinterpret_cast` on the `sb` reference to match the output (float or double), even-though the oneMKL spec only assumes float types for the `sdsdot` operator _(so we could have disabled the double support for it all-together as well)._

test/unittest/blas1/blas1_sdsdot_test.cpp
```diff
@@ -35,7 +35,7 @@ template <typename scalar_t, helper::AllocType mem_alloc>
 void run_test(const combination_t<scalar_t> combi) {
   std::string alloc;
   index_t N;
-  float sb;
+  scalar_t sb;
```
muhammad-tanvir-1211, line 38: I think this still needs to be `float` instead of `scalar_t` since this argument is of type `float` in oneapi specification for `sdsdot`.

test/unittest/blas1/blas1_sdsdot_test.cpp
```diff
@@ -104,7 +104,7 @@ void run_test(const combination_t<scalar_t> combi) {
 
   std::string alloc;
   index_t N;
-  float sb;
+  scalar_t sb;
```
muhammad-tanvir-1211, line 107: Same here.

test/unittest/blas1/blas1_sdsdot_test.cpp
```diff
@@ -149,7 +149,7 @@ static std::string generate_name(
     const ::testing::TestParamInfo<combination_t<T>>& info) {
   std::string alloc;
   int size, incX, incY;
-  float sb;
+  T sb;
```
muhammad-tanvir-1211, line 152: Same here.

test/unittest/blas1/blas1_sdsdot_test.cpp
OuadiElfarouki: Indeed! Will fix that.

muhammad-tanvir-1211 (approved):
LGTM, Thanks.

PR #472, opened 2023-10-12
Fix omatadd_batch interface

s-Nick:
`omatadd_batch` interface is broken for `usm`. It is not possible passing `const` ptr and `non-const` ptr because the input type is expected to be the same.
This patch fix this issue.


muhammad-tanvir-1211 (approved):
LGTM, Thanks.

OuadiElfarouki (approved):

PR #475, opened 2023-10-27
Reduce .so library size

OuadiElfarouki:
This PR exploits the internal cmake function `add_gemm_configuration` that helps generate template instantiations of different gemm kernels configurations, by dropping a number of unused cases :
- symm_a==true and symm_b==true
- symm_a==true and trans_b==true
- symm_b==true and trans_a==true

This patch reduces the final .so library file by 25 to 50% across all supported devices.


muhammad-tanvir-1211 (approved):
LGTM, Thanks.

s-Nick (approved):
LGTM!

PR #478, opened 2023-11-07
Reuse temporary memory allocations

pgorlani:
This patch introduces a pool of temporary memory allocations within the `sb_handler`, which allows reusing them in order to save re-allocation time.

### DIFF
--- a/include/sb_handle/portblas_handle.h
+++ b/include/sb_handle/portblas_handle.h
@@ -22,7 +22,6 @@
  *  @filename portblas_handle.h
  *
  **************************************************************************/
-
 #ifndef PORTBLAS_HANDLE_H
 #define PORTBLAS_HANDLE_H
 #include "blas_meta.h"
@@ -31,6 +30,9 @@
 #include "operations/blas3_trees.h"
 #include "operations/extension/reduction.h"
 #include "portblas_helper.h"
+#include <map>
+#include <mutex>
+
 namespace blas {
 
 /** SB_Handle.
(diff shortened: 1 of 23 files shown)


src/sb_handle/temp_memory_pool.hpp
```diff
@@ -0,0 +1,118 @@
...
+  temp_buffer_map_mutex_.lock();  // lock
+  if (temp_buffer_map_tot_byte_size_ + byteSize <= max_size_temp_mem_) {
+    temp_buffer_map_tot_byte_size_ += byteSize;
+    temp_buffer_map_.emplace(byteSize, rebuff);
+  }
+  temp_buffer_map_mutex_.unlock();  // unlock
+}
+
+template <typename container_t>
+typename Temp_Mem_Pool::event_t Temp_Mem_Pool::release_buff_mem(
```
muhammad-tanvir-1211, line 52: Could you please explain why we need a host task for releasing buffers here?

pgorlani: Thanks for your review, @muhammad-tanvir-1211.

We  call the `release_buff_mem_`  via a `host_task` because we need to insert the buffer within the map just when the involved computation is finished. Otherwise, a buffer still in use could be passed to another operation creating unnecessary dependencies between unrelated operations.

muhammad-tanvir-1211 (approved):
LGTM. Thanks.

s-Nick (commented):
Thank you for you work @pgorlani it looks very good to me!
I have some minor comments on it.

src/sb_handle/temp_memory_pool.hpp
```diff
@@ -0,0 +1,118 @@
...
+      reinterpret_cast<temp_usm_size_map_t::key_type>(mem));
+  const size_t byteSize = found->second;
+  if (temp_usm_map_tot_byte_size_ + byteSize > max_size_temp_mem_) {
+    temp_usm_size_map_.erase(found);
+    temp_usm_map_mutex_.unlock();  // unlock
+    cl::sycl::free(mem, q_);
+  } else {
+    temp_usm_map_tot_byte_size_ += byteSize;
+    temp_usm_map_.emplace(byteSize,
+                          reinterpret_cast<temp_usm_map_t::mapped_type>(mem));
```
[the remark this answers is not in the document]
pgorlani: Thanks for your review, @s-Nick.

Please note that `temp_usm_map_` is a `multimap` and not `map`, this means that is possible to have multiple elements with the same key. For this reason, the described situation cannot happen. The `emplace` method in `multimap` returns just an iterator without any `bool` because it will result in an insertion in any case.

include/sb_handle/temp_memory_pool.h
```diff
@@ -0,0 +1,112 @@
...
+
+  template <typename container_t>
+  typename Temp_Mem_Pool::event_t release_usm_mem(
+      const typename Temp_Mem_Pool::event_t&, const container_t&);
+#endif
+
+ private:
+  static_assert(sizeof(temp_buffer_map_t::mapped_type::value_type) == 1);
+
+  static constexpr size_t max_size_temp_mem_ = 1e9;
```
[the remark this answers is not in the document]
pgorlani: That is an arbitrary value for limiting the number of the allocated bytes. I could also write `1<<9`, but I think that scientific notation is clearer despite being a double literal.

src/sb_handle/temp_memory_pool.hpp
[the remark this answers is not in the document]
s-Nick: Thank you for the explanation @pgorlani, now it makes perfectly sense to me.

s-Nick (approved):

include/sb_handle/portblas_handle.h
```diff
@@ -47,11 +49,49 @@ class SB_Handle {
  public:
   using event_t = std::vector<cl::sycl::event>;
   inline SB_Handle(queue_t q)
-      : q_(q),
+      : tempMemPool_(NULL),
```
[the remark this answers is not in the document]
pgorlani: Why not, https://github.com/codeplaysoftware/portBLAS/pull/478/commits/9b4ca547900fe6f0f87676ebed8e349a6e818d08.

include/sb_handle/temp_memory_pool.h
```diff
@@ -0,0 +1,112 @@
...
+  void release_usm_mem_(const container_t& mem);
+
+#ifdef SB_ENABLE_USM
+  std::mutex temp_usm_map_mutex_;
+  size_t temp_usm_map_tot_byte_size_;
+  temp_usm_map_t temp_usm_map_;
+  temp_usm_size_map_t temp_usm_size_map_;
+
+  template <typename container_t>
+  void release_buff_mem_(const container_t& mem);
```
OuadiElfarouki, line 108: Sorry for the late comment. I think this one should be swapped with the `release_usm_mem_` above.

include/sb_handle/temp_memory_pool.h
```diff
@@ -0,0 +1,112 @@
...
+  using event_t = std::vector<cl::sycl::event>;
+  using temp_usm_map_t = std::multimap<size_t, void*>;
+  using temp_usm_size_map_t = std::map<void*, size_t>;
+  using temp_buffer_map_t = std::multimap<size_t, cl::sycl::buffer<int8_t, 1>>;
+
+ public:
+  Temp_Mem_Pool(queue_t q)
+      : q_(q),
+        temp_buffer_map_tot_byte_size_(0),
+        temp_usm_map_tot_byte_size_(0) {}
```
OuadiElfarouki, line 43: the `temp_usm_map_tot_byte_size_` is undefined if SB_ENABLE_USM is not defined.

pgorlani: Fixed in https://github.com/codeplaysoftware/portBLAS/commit/aa1fdf1de65df04b8034d4949d3ba948a33fe2a6.

PR #480, opened 2023-11-15
Fix type comparison for joint_matrix

muhammad-tanvir-1211:
This patch fixes a bug identified when passing `const float` as input container type for using `joint_matrix` api in portBLAS instead of `float`. Also, I removed a few lines from `.clang-format` file which were causing error while formatting the source code.


aacostadiaz (approved):
LGTM!

s-Nick (approved):

PR #481, opened 2023-11-17
Fixed is_device_copyable error for IndexValueTuple

muhammad-tanvir-1211:
This PR fixes the `is_device_copyable` error for `IndexValueTuple` in portBLAS. It also removes an extra file that wasn't removed previously. Applied `clang-format` to `blas2_interface.hpp` as well.

### DIFF
--- a/include/operations/blas_constants.h
+++ b/include/operations/blas_constants.h
@@ -265,4 +265,16 @@ struct constant_pair {
 
 }  // namespace blas
 
+#define PORTBLAS_DEVICE_COPY(T1, T2)                             \
+  template <>                                                    \
+  struct sycl::is_device_copyable<blas::IndexValueTuple<T1, T2>> \
+      : std::true_type {};
+
+PORTBLAS_DEVICE_COPY(int32_t, float)
+PORTBLAS_DEVICE_COPY(int32_t, double)
+PORTBLAS_DEVICE_COPY(int64_t, float)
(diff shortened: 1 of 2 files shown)


include/operations/blas_constants.h
```diff
@@ -265,4 +265,16 @@ struct constant_pair {
...
+  template <>                                                    \
+  struct sycl::is_device_copyable<blas::IndexValueTuple<T1, T2>> \
+      : std::true_type {};
+
+PORTBLAS_DEVICE_COPY(int32_t, float)
+PORTBLAS_DEVICE_COPY(int32_t, double)
+PORTBLAS_DEVICE_COPY(int64_t, float)
+PORTBLAS_DEVICE_COPY(int64_t, double)
+
+#undef PORTBLAS_DEVICE_COPY
```
s-Nick, line 278: Thank you for the fix @muhammad-tanvir-1211 
I have just a minor consideration, did you think about using only a template to fix the issue? I think that something like the following should work:

```C++
template<typename ind_t, typename val_t>
struct sycl::is_device_copyable<blas::IndexValueTuple<ind_t, val_t>>
      : std::true_type {};
```

muhammad-tanvir-1211: Thank you for the suggestion @s-Nick. I have applied the patch in the latest commit.

OuadiElfarouki (approved):

s-Nick (approved):
LGTM!

PR #482, opened 2023-11-27
Refactor iamax and iamin api to match oneapi spec

muhammad-tanvir-1211:
This PR refactors the existing `iamax` and `iamin` operator apis' to match the oneapi specification and introduces a `backend` specific single step or two step reduction based on the input size.


src/interface/blas1_interface.hpp
```diff
@@ -246,6 +246,108 @@ typename sb_handle_t::event_t _asum_impl(
...
+    if constexpr (localMemSize == 0) {
+      auto q = sb_handle.get_queue();
+      const index_t sg_size = static_cast<index_t>(
+          q.get_device()
+              .template get_info<sycl::info::device::sub_group_sizes>()[0]);
+      ret = sb_handle.execute(op, sg_size, sg_size, _dependencies);
+    } else {
+      ret =
+          sb_handle.execute(op, static_cast<index_t>(localSize),
+                            _nWG * static_cast<index_t>(localSize),
```
pgorlani, line 303: Thanks for your patch, @muhammad-tanvir-1211 .

I think that `_nWG` must be `1` since `single` is true or I'm missing something.

In that case, I would add a comment explaining it and 
```suggestion
                            static_cast<index_t>(localSize),
```

src/interface/blas1_interface.hpp
```diff
@@ -246,6 +246,108 @@ typename sb_handle_t::event_t _asum_impl(
...
+        make_vector_view(gpu_res, static_cast<increment_t>(1), memory_size);
+    auto step0 = make_index_max_min<is_max, true>(gpu_res_vec, tupOp);
+    auto step1 = make_index_max_min<is_max, false>(rs, gpu_res_vec);
+    if constexpr (localMemSize == 0) {
+      const scalar_t val = is_max ? std::numeric_limits<scalar_t>::min()
+                                  : std::numeric_limits<scalar_t>::max();
+      const index_t idx = std::numeric_limits<index_t>::max();
+      const tuple_t init{idx, val};
+      ret = typename sb_handle_t::event_t{
+          helper::fill(q, gpu_res, init, memory_size, _dependencies)};
```
pgorlani, line 330: I'm missing the reason why we need to initilize `gpu_res` just in case of `localMemSize == 0`.

muhammad-tanvir-1211: When `localMemSize == 0`, the implementation assumes that it is going to be executed on the CPU. The size of `gpu_res` is dependant on the `sg_size` variable since we want every `sub_group` to write one intermediate output to global memory. Since the `sub_group` size (to be used by the kernel) is not known at compile time, I allocated a larger memory using the smallest `sub_group` size supported by the CPU. This means that there could be a case where the actual `sub_group` size chosen by the compiler is larger than the one used to calculate the size of `gpu_res`, and this would mean that there could be locations inside `gpu_res` with garbage value that would effect the kernel output in the final reduction. That's why we have to initialize the intermediate memory for the CPU case only. For GPUs, every work group writes 1 value of data to global memory, and since the no. of work groups are known, we don't need to initialize the memory beforehand.

muhammad-tanvir-1211: e.g. on the device I tested, smallest sub_group size supported was `4` whereas the sub_group size used by the kernel was `16`.

pgorlani: Ok, fine. Please, could you add some comments for explaining it and change the variable names accordingly? E.g., `sg_size -> min_sg_size`.

src/interface/blas1_interface.hpp
```diff
@@ -258,17 +360,14 @@ template <typename sb_handle_t, typename container_t, typename ContainerI,
...
-  auto rs = make_vector_view(_rs, static_cast<increment_t>(1),
-                             static_cast<index_t>(1));
-  const auto localSize = sb_handle.get_work_group_size();
-  const auto nWG = 2 * localSize;
-  auto tupOp = make_tuple_op(vx);
-  auto assignOp = make_assign_reduction<IMaxOperator>(rs, tupOp, localSize,
-                                                      localSize * nWG);
-  auto ret = sb_handle.execute(assignOp, _dependencies);
-  return ret;
+  if (_incx < 0) {
```
s-Nick, line 363: Thank you for you work @muhammad-tanvir-1211 
oneMKL specification says: *"If either n or incx is not positive, the routine returns 0, regardless of the base of the index selected."* 
So I think we should add a check also for `_N` here and at line `385`

include/portblas_helper.h
```diff
@@ -138,22 +138,24 @@ inline size_t get_num_compute_units(cl::sycl::queue &q) {
...
-inline cl::sycl::event copy_to_device(cl::sycl::queue q, const element_t *src,
-                                      BufferIterator<element_t> dst,
-                                      size_t size) {
+inline cl::sycl::event copy_to_device(
+    cl::sycl::queue q, const element_t *src, BufferIterator<element_t> dst,
+    size_t size, const std::vector<cl::sycl::event> &_dependencies = {}) {
   auto event = q.submit([&](cl::sycl::handler &cgh) {
     auto acc = dst.template get_range_accessor<cl::sycl::access::mode::write>(
         cgh, size);
+    cgh.depends_on(_dependencies);
```
s-Nick, line 147: Is this dependency enforcement necessary? The buffer/accessor runtime should handle dependency and keep it correct, shouldn't it?

muhammad-tanvir-1211: Ideally speaking, this is not required for buffers since the dependencies are handled by the runtime. I added it here for completeness sake since we pass the dependencies to `queue.memcpy` for USM.

s-Nick: I understand the completeness sake, but could this interfere with the runtime in some cases?

muhammad-tanvir-1211: I don't think so, no.

s-Nick (approved):
LGTM!

src/interface/blas1_interface.hpp
muhammad-tanvir-1211: Fixed this in the recent commit.

pgorlani (approved):

PR #485, opened 2023-12-13
Refined transpose kernel configurations for CPU target & down-sized some copy-op benchmarks

OuadiElfarouki:
This patch updates the kernel configurations of both  `_transpose_outplace` & `_transpose_add` for default CPU targets resulting in an overall increase in performance (benchmark numbers shared internally).
The PR also includes some refactoring of omatcopy/omatadd batched operators benchmarks to address the time consuming ones.


s-Nick (approved):
Thank you @OuadiElfarouki LGTM

PR #491, opened 2024-01-05
Fix joint_matrix implementation to match latest api

muhammad-tanvir-1211:
This PR updates the `joint_matrix` based `GEMM` implementation to match the api from `intel/llvm` (commit # `2a828f49283145433dc9bbbff74cefcb2d2b10dc`). 

* Removed the deprecated `get_wi_data()` calls and replaced them with `joint_matrix_apply()` and `joint_matrix_copy()` calls.
* Changed the leading dimensions for shared memory accesses to avoid excessive bank conflicts.
* Refactored the code to write output data to Global Memory in a coalesced way (this wasn't possible earlier due to the `joint_matrix_store()` operation performed on the Global pointer `C`)
* Fixed race condition for `batch size > 1`.


src/operations/blas3/gemm_local_joint_matrix.hpp
```diff
@@ -787,49 +772,58 @@ class Gemm<input_t, output_t, DoubleBuffer, NbcA, NbcB, ClSize, TileType,
...
     const index_t strideA = ldsa;
     const index_t strideB = ldsb;
 
     auto sg = id.get_sub_group();
 
 #pragma unroll
-    for (index_t frag = 0; frag < frags_per_sg; frag++) {
-      auto new_B = s2 + frag * tile_type::joint_matrix_N * ldsb;
-      auto new_A = s4;
+    for (index_t i = 0; i < cl_elems / tile_type::joint_matrix_K; i++) {
```
s-Nick, line 806: Thank @muhammad-tanvir-1211 for your work and for fixing the compilation compatibility issue.
LGTM, the only minor change is that I think this `for` condition could be moved at compile time as you did with other loops

s-Nick (approved):

OuadiElfarouki (approved):
LGTM!

test/unittest/blas3/blas3_trsm_test.cpp
```diff
@@ -84,7 +92,7 @@ void run_test(const combination_t<scalar_t> combi) {
       blas::helper::copy_to_host<scalar_t>(q, b_gpu, B.data(), B.size());
   sb_handle.wait(event);
 
-  bool isAlmostEqual = utils::compare_vectors(cpu_B, B);
+  bool isAlmostEqual = utils::compare_vectors(cpu_B, B, std::cerr, "", true);
```
pgorlani, line 95: Thanks for your PR, @muhammad-tanvir-1211.

```suggestion
  bool isAlmostEqual = utils::compare_vectors(cpu_B, B, std::cerr, "", en_joint_matrix != NULL && en_joint_matrix == '1');
```
I think it is better to increase the tolerance just in case of the joint matrix is used.

common/include/common/float_comparison.hpp
```diff
@@ -152,16 +162,27 @@ inline cl::sycl::half getRelativeErrorMargin<cl::sycl::half>() {
...
    * absolute differences of up to 0.0006 were observed for float
    */
-  return 0.001f;
+  scalar_t margin = 0.001f;
+  if (is_trsm) {
+    const char* en_joint_matrix = std::getenv("SB_ENABLE_JOINT_MATRIX");
+    if (en_joint_matrix != NULL && std::is_same<scalar_t, float>::value &&
+        *en_joint_matrix == '1') {
+      // increase error margin for mixed precision calculation
+      // for trsm operator.
```
pgorlani, line 176: I would move the environmental variable  within the tests, in order to make it clearer.

CMakeLists.txt
```diff
@@ -203,6 +203,7 @@ export(EXPORT portblas
 
 option(BLAS_ENABLE_TESTING "Whether to enable testing" ON)
 option(ENABLE_EXPRESSION_TESTS "Whether to build expression tree fusion tests" OFF)
+option(ENABLE_JOINTMATRIX_TESTS "Whether to build joint_matrix GEMM tests" OFF)
```
pgorlani, line 206: Thanks for your PR, @muhammad-tanvir-1211.

Probably this flag needs to be documented in the [README.md#cmake-options](https://github.com/codeplaysoftware/portBLAS/blob/master/README.md#cmake-options).

test/blas_test.hpp
```diff
@@ -223,6 +223,31 @@ static inline void fill_trsm_matrix(std::vector<scalar_t> &A, size_t k,
...
+ * @brief Set to zero the last n bits of a float.
+ *
+ * @param val input/output float value.
+ * @param nbits number of last bit set to zero. It is set by default to 13 since
+ * this is the difference of the number of bits of the mantissa between floats
+ * (23) and FP16 / NVIDIA TF32 (10). For bfloat16, this value needs to be set to
+ * 16 to get correct result.
+ */
+static inline void set_to_zero_last_nbits(float &val, int32_t nbits = 13) {
+  int32_t *int_pntr = reinterpret_cast<int32_t *>(&val);
```
pgorlani, line 236:
```suggestion
template <typename T>
void set_to_zero_last_nbits(T &val, int32_t nbits = 13) {
  static_assert(sizeof(T) <= 64);
  using integer_t =
      std::conditional_t<sizeof(T) == 64, int64_t,
                         std::conditional_t<sizeof(T) == 32, int32_t, int16_t>>;
  integer_t *int_pntr = reinterpret_cast<integer_t *>(&val);
```

We need to make this function generic  for compiling in case of `double`s even if we don't use them.

*Same below.*

test/blas_test.hpp
```diff
@@ -223,6 +223,31 @@ static inline void fill_trsm_matrix(std::vector<scalar_t> &A, size_t k,
...
+
+/**
+ * @brief Set to zero the last n bits of floats contained in a vector.
+ *
+ * @param val input/output float vector.
+ * @param nbits number of last bit set to zero.
+ */
+static inline void set_to_zero_last_nbits(std::vector<float> &vec,
+                                          int32_t nbits = 13) {
+  for (float &val : vec) set_to_zero_last_nbits(val, nbits);
```
pgorlani, line 248:
```suggestion
template <typename T>
void set_to_zero_last_nbits(std::vector<T> &vec, int32_t nbits = 13) {
  for (T &val : vec) set_to_zero_last_nbits(val, nbits);
```

test/unittest/blas3/blas3_trsm_test.cpp
```diff
@@ -84,7 +92,12 @@ void run_test(const combination_t<scalar_t> combi) {
       blas::helper::copy_to_host<scalar_t>(q, b_gpu, B.data(), B.size());
   sb_handle.wait(event);
 
-  bool isAlmostEqual = utils::compare_vectors(cpu_B, B);
+  bool isAlmostEqual = utils::compare_vectors(
+      cpu_B, B, std::cerr, "",
```
pgorlani, line 96:
```suggestion
      cpu_B, B, std::cerr, "\n",
```
You need to pass `"\n"` as an argument to `compare_vector` otherwise the error output will be malformed.

*Apply this to all the occurrences of `compare_vectors` in which the 'error multiplier' argument is present.*

CMakeLists.txt
muhammad-tanvir-1211: This has been added in the README.

PR #492, opened 2024-01-16
Fixes for SPR/SPR2 BLAS 2 routines

pgorlani:
This patch fixes various aspects of the SPR/SPR2 BLAS2 routines:

 * fix the mapping of the global work-item index to the triangular matrix coordinates by using only integer arithmetic;
 * move method definitions to the proper file;
 * remove unnecessary increment arguments;
 * improve unit tests by randomly generating the matrix and removing the unnecessary memory layout argument;
 * improve documentation.

### DIFF
--- a/test/unittest/blas2/blas2_spr_test.cpp
+++ b/test/unittest/blas2/blas2_spr_test.cpp
@@ -26,8 +26,7 @@
 #include "blas_test.hpp"
 
 template <typename scalar_t>
-using combination_t =
-    std::tuple<std::string, char, char, index_t, scalar_t, index_t>;
+using combination_t = std::tuple<std::string, char, index_t, scalar_t, index_t>;
 
 template <typename scalar_t, helper::AllocType mem_alloc>
 void run_test(const combination_t<scalar_t> combi) {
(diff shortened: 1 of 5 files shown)


test/unittest/blas2/blas2_spr_test.cpp
```diff
@@ -26,18 +26,17 @@
...
-    std::tuple<std::string, char, char, index_t, scalar_t, index_t>;
+using combination_t = std::tuple<std::string, char, index_t, scalar_t, index_t>;
 
 template <typename scalar_t, helper::AllocType mem_alloc>
 void run_test(const combination_t<scalar_t> combi) {
   std::string alloc;
   index_t n;
   index_t lda_mul;
   index_t incX;
-  char layout, uplo;
```
muhammad-tanvir-1211, line 38: I think we can keep the `layout` option here, since the code practically works for both `row major` and `col major` layouts. The only reason why this option was kept here and not introduced in the library was because we don't have a dedicate support for `row major` layout yet.

muhammad-tanvir-1211 (approved):
LGTM, apart from the comment regarding `layout` in tests. Thanks.

OuadiElfarouki (approved):
Looks good thank you!

PR #493, opened 2024-01-17
HipSYCL to AdaptiveCpp update & fixes

OuadiElfarouki:
This patch updates the hipSYCL compilation in portBLAS including :
- Necessary naming updates hipSYCL -> AdaptiveCpp)
- Cmake changes to account for AdaptiveCpp compilation approaches
- Relevant compatibility considerations of SYCL 2020 features and other extensions still not supported by AdaptiveCpp.
- Updated documentation/readme.

Compilation flows / Device targets compiled successfully at this stage : 
- intel CPU : omp default flow (host), and generic  SPIRV/OpenCL flow.
- intel GPU : generic OpenCL/SPIRV flow, and generic SPIRV/LevelZero flow.
- Nvidia GPU : generic PTX/cuda flow.
- AMD GPU : generic HIP/amdgcn flow. (Tested with rocm 5.6.1)

Tests and benchmarks are conditionally disabled for some operators depending on the target device and AdaptiveCpp flow. Remaining tests (in each case) are passing for all tested devices above.


benchmark/portblas/CMakeLists.txt
```diff
@@ -76,12 +76,29 @@ if(${BLAS_ENABLE_EXTENSIONS})
   list(APPEND sources extension/reduction.cpp)
 endif()
 
+# Skip these benchmarks for AdaptiveCpp for SPIRV/OpenCL targets
+# that use SYCL 2020 features like reduction or hang during execution
+# (https://github.com/AdaptiveCpp/AdaptiveCpp/issues/1309)
+set(ADAPTIVE_CPP_SKIP
```
muhammad-tanvir-1211, line 82: Why does this list not include `tbsv`, `trsv` and `tpsv` as mentioned in the `README.md`?

src/operations/blas1/IndexMaxMin.hpp
```diff
@@ -94,6 +94,7 @@ PORTBLAS_INLINE void IndexMaxMin<is_max, is_step0, lhs_t, rhs_t>::eval(
   using element_t =
       typename ResolveReturnType<op, rhs_t>::type::value_t::value_t;
 
+#ifndef __HIPSYCL__
```
muhammad-tanvir-1211, line 97: Do we need the macro here? I think this implementation is never going to be used with AdaptiveCpp because we have changed the `default_cpu.hpp` file.

benchmark/portblas/CMakeLists.txt
OuadiElfarouki: those are missing indeed, will ad them!

src/operations/blas1/IndexMaxMin.hpp
OuadiElfarouki: Yes indeed it's never going there, but for some reason the compiler fails if the `shuffle_down` in that block is exposed unconditionally. Unlike the `sycl::reduce_over_group`, the shuffle_down is not recognized by the compiler and thus the error shows at compile time not runtime.

src/interface/gemm_interface.hpp
```diff
@@ -28,9 +28,10 @@
 #include "interface/blas1_interface.h"
 #include "interface/blas3/backend/backend.hpp"
 #include "interface/blas3_interface.h"
+#include "interface/extension_interface.h"
```
muhammad-tanvir-1211, line 31: Do we need this header here?

test/unittest/blas3/blas3_gemm_batched_test.cpp
```diff
@@ -146,6 +146,27 @@ const auto AllStridedBatched =
     );
 GENERATE_GEMM_STRIDED_BATCHED_TEST(BatchStridedGemm, AllStridedBatched);
 
+template <typename scalar_t>
+const auto ZerosStridedBatched =
```
muhammad-tanvir-1211, line 150: I think we can merge this test combination with the `AllStridedBatched` one.

src/interface/gemm_interface.hpp
OuadiElfarouki: Typo left from some experiments, will remove it!

test/unittest/blas3/blas3_gemm_batched_test.cpp
OuadiElfarouki: Yes will add it ;)

s-Nick (commented):
Thank for your patch @OuadiElfarouki 
I point out other `hipsycl` reference that could be updated to `adaptiveCpp` everything else LGTM

include/sb_handle/temp_memory_pool.h
```diff
@@ -105,8 +109,19 @@ class Temp_Mem_Pool {
...
 };
 }  // namespace blas
+
+#else
+
+namespace blas {
+// Empty class to serve as a temporary Placeholder asthe feature
+// is not fully supported when using AdaptiveCpp
+class Temp_Mem_Pool;
+}  // namespace blas
```
pgorlani, line 123:
```suggestion
```
Are you sure this is needed? There should be no references to `Temp_Mem_Pool` if `__HIPSYCL__` is defined.

README.md
```diff
@@ -417,13 +417,38 @@ advisable for NVIDIA and **mandatory for AMD** to provide the specific device
...
+export CC=[path/to/system/clang]
+export CXX=[path/to/AdaptiveCpp/install/bin/syclcc]
+export HIPSYCL_TARGETS=[compilation_flow:target] # (e.g. cuda:sm_75)
+cmake -GNinja ../ -DAdaptiveCpp_DIR=/path/to/AdaptiveCpp/install/lib/cmake/AdaptiveCpp \
+      -DSYCL_COMPILER=adaptivecpp -DHIPSYCL_TARGETS=$HIPSYCL_TARGETS
 ninja
 ```
-To build for other than the default devices (`omp`), set the `HIPSYCL_TARGETS` environment variable or specify `-DHIPSYCL_TARGETS` as [documented](https://github.com/illuhad/hipSYCL/blob/develop/doc/using-hipsycl.md).
+To build for other than the default devices (`omp`), set the `HIPSYCL_TARGETS` environment
+variable or specify `-DHIPSYCL_TARGETS` as
```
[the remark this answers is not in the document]
OuadiElfarouki: Yes the new macros/variables work too so I'll update them as well. At this point people who want to use hipSYCL will definitely know it changed naming etc.. so no confusion hopefully.

include/sb_handle/temp_memory_pool.h
OuadiElfarouki: Yes  @pgorlani we can remove this but in that case we'll need to hide it conditionally in `portblas_handle.h` with a macro there so that we don't have an undefined attribute type.

include/operations/blas_constants.h
```diff
@@ -265,16 +265,18 @@ struct constant_pair {
 
 }  // namespace blas
 
+#ifndef __HIPSYCL__
```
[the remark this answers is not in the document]
OuadiElfarouki: same, thanks for the suggestion I'll update this as well.

README.md
```diff
@@ -425,13 +425,13 @@ by AdaptiveCpp.
 cd build
 export CC=[path/to/system/clang]
 export CXX=[path/to/AdaptiveCpp/install/bin/syclcc]
```
s-Nick, line 427: Running the tests, `syclcc` explicitly says it is deprecated and it suggests to use `acpp`. An AdaptiveCpp user should know about it, but I suggest to update this path:
```suggestion
export CXX=[path/to/AdaptiveCpp/install/bin/acpp]
```

s-Nick (approved):
A part that small suggestion, LGTM

muhammad-tanvir-1211 (approved):

PR #494, opened 2024-01-24
Update configuration for gemm on AMD GPUs

s-Nick:
Following preliminary investigation and tuning with the auto-tuner, these are the new configurations for `gemm` that provide the best performance. 
The selection of the configuration is now based on the arithmetic intensity and not only on `_M` and `_N` dimension.

### DIFF
--- a/src/interface/blas3/backend/amd_gpu.hpp
+++ b/src/interface/blas3/backend/amd_gpu.hpp
@@ -134,7 +134,31 @@ _gemm(sb_handle_t& sb_handle, index_t _M, index_t _N, index_t _K,
       // Folling configuration are taken using the auto tuner on amd-mi210
       // and divided following their arith_intensity or another ratio between _N
       // and _K input size
-      if (arith_intensity >= 360 || (_N >> 4) > _K) {
+      if ((_N >> 4) > _K) {
+        if (arith_intensity <= 100) {
+          return blas::Gemm_Launcher<
+              container_0_t, container_1_t, container_2_t, 256, false, false,
+              true, 64, Tile<4, 8, 16, 16>, _t_a, _t_b, s_a, s_b,
+              static_cast<int>(gemm_memory_t::local),
+              static_cast<int>(gemm_algorithm_t::standard),
+              static_cast<int>(gemm_vectorization_t::full), is_beta_zero, 1,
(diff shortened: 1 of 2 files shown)


src/interface/blas3/backend/amd_gpu.hpp
```diff
@@ -123,18 +131,79 @@ _gemm(sb_handle_t& sb_handle, index_t _M, index_t _N, index_t _K,
       }
     } else
 #endif  // GEMM_TALL_SKINNY_SUPPORT
-      if (_M * _N <= 65536) {
+      // Folling configuration are taken using the auto tuner on amd-mi210
```
OuadiElfarouki, line 134: Some typos in the comment ;)

src/interface/blas3/backend/amd_gpu.hpp
```diff
@@ -123,18 +131,79 @@ _gemm(sb_handle_t& sb_handle, index_t _M, index_t _N, index_t _K,
...
+      // and divided following their arith_intensity or another ratio between _N
+      // and _K input size
+      if (arith_intensity >= 360 || (_N >> 4) > _K) {
```
OuadiElfarouki, line 137: The `arithm_intensity` based if/else statements seem coherent, but combining them with the `(_N >> 4) > _K)` confuses me a little bit. If it has a direct impact _(independently or depending on the combination's `arithm_intensity`)_, isn't it better to put is a stand-alone if/else statement ? 
Something like : 
```
if( (_N >> 4) > _K))
// Config or Configs _(depending on arith_intensity - 2 in total if we follow the setup here : arith_intensity > or <= 100)_
else {
// Rest of `arith_intensity` conditional statements.
}
```

s-Nick: Thank you for your review @OuadiElfarouki 
I addressed it in commit 94d5704. Please let me know if it seems more readable to you now, otherwise I can revert it back to what it was.
The performance are not affected by this change.

OuadiElfarouki: Yes looks more readable now thanks a lot!

OuadiElfarouki (approved):
Looks good thank you!

PR #495, opened 2024-02-12
Revisited & Fixed Half (fp16) data support

OuadiElfarouki:
This PR is an update and extension of half data support in portBLAS and includes following changes : 
- half support is enabled using the cmake option `BLAS_ENABLE_HALF` and is only applied to operators meant to support half according to oneMKL spec _(so far in this PR axpy, scal and gemm)_
- unittests & benchmarksare extended to support mixed-precision comparison (reference BLAS libs only support float/double).
- Extended unittests for axpy, scal, and gemm (+gemm_batched) using half.
- Extended portblas, cublas & rocblas benchmarks for gemm (+gemm_batched).
- Separated gemm configurations when using half data type for each `TUNING_TARGET` from the float/double configurations.

Other notes : 
- half precision support is disabled when targetting `DEFAULT_CPU` due to lack of fp16 support.
- some legacy gemm configurations for intel GPU targets with sycl::half have been removed (not based on a tuning but rather a temporary reduction of generated kernels)


common/include/common/float_comparison.hpp
```diff
@@ -101,7 +101,72 @@ inline cl::sycl::half abs<cl::sycl::half>(cl::sycl::half value) noexcept {
...
+/**
+ * Float/Double to sycl::half cast utility function using the intermediate
+ * cast_to_half(T&).
+ */
+template <typename T>
+inline cl::sycl::half cast_to_sycl_half(T& val) {
+  uint16_t half_bin = cast_to_half(val);
+  cl::sycl::half result = *reinterpret_cast<cl::sycl::half*>(&half_bin);
+  return static_cast<cl::sycl::half>(result);
+}
```
pgorlani, line 168: Thanks for your patch, @OuadiElfarouki.

```suggestion
template <typename T>
inline cl::sycl::half cast_to_sycl_half(T& val) {
  return static_cast<cl::sycl::half>(val);
}
```
Do we really need a custom function (`cast_to_half`) for casting the halfs?

test/unittest/blas1/blas1_scal_test.cpp
```diff
@@ -30,33 +30,53 @@ using combination_t = std::tuple<std::string, int, scalar_t, int>;
...
+  }
+
+  if constexpr (is_sycl_half) {
+    // Float-type variables for reference ops
+    ref_scalar_t alpha_f = alpha;
+    // sycl::half to float reference type
+    std::transform(x_v.begin(), x_v.end(), x_cpu_v.begin(),
+                   [](scalar_t x) { return (static_cast<ref_scalar_t>(x)); });
+    // Reference implementation
+    reference_blas::scal(size, alpha_f, x_cpu_v.data(), incX);
```
pgorlani, line 60: Did you considered performing the type conversion within the `reference_blas` implementation rather than re-implementing every time for each unittest/benchmark? E.g.,
```
 -- common/include/common/system_reference_blas.hpp --
 template <typename scalar_t>
 void scal(const int n, const scalar_t alpha, scalar_t x[], const int incX) {
  if constexpr (!std::is_same_v<scalar_t, cl::sycl::half>) {
    auto func = blas_system_function<scalar_t>(&cblas_sscal, &cblas_dscal);
    func(n, alpha, x, incX);
  } else {
    std::vector<float> x_f( n * incX);
    for(int i = 0; i < n * incX; ++i)
      x_f[i] = static_cast<float>(x[i]);
    cblas_sscal(n, float(alpha), x_f.data(), incX);
    for(int i = 0; i < n * incX; ++i)
      x[i] = static_cast<cl::sycl::half>(x_f[i]);
  }
 }
```
This kind of solution will allow to simplify all the unittests/benchmarks of the operators supporting halfs. Moreover, this will remove the need for a specific `compare_vectors` function.

common/include/common/float_comparison.hpp
OuadiElfarouki: You're right it wasn't necessary. Got it removed!

benchmark/cublas/blas3/gemm.cpp
```diff
@@ -35,6 +35,11 @@ static inline void cublas_routine(args_t&&... args) {
   } else if constexpr (std::is_same_v<scalar_t, double>) {
     CUBLAS_CHECK(cublasDgemm(std::forward<args_t>(args)...));
   }
+#ifdef BLAS_ENABLE_HALF
+  else if constexpr (std::is_same_v<scalar_t, cl::sycl::half>) {
+    CUBLAS_CHECK(cublasHgemm(std::forward<args_t>(args)...));
+  }
+#endif
```
pgorlani, line 42:
```suggestion
  } else if constexpr (std::is_same_v<scalar_t, cl::sycl::half>) {
    CUBLAS_CHECK(cublasHgemm(std::forward<args_t>(args)...));
  }
```
I think this `#ifdef` is redundant. Same for the `rocblas` one.

benchmark/rocblas/blas3/gemm.cpp
```diff
@@ -91,81 +103,120 @@ void run(benchmark::State& state, rocblas_handle& rb_handle, int t_a_i,
...
+  std::vector<scalar_t> c_temp(n * m, 0);
+
+  if constexpr (is_half) {
+    // Float-type variables for reference ops
+    ref_scalar_t alpha_f = alpha;
+    ref_scalar_t beta_f = beta;
+    std::vector<ref_scalar_t> a_f(m * k);
+    std::vector<ref_scalar_t> b_f(k * n);
+
+    // sycl::half to float reference type
```
pgorlani, line 141: I think you forgot to remove the reference casting from the rocblas benchmarks.

test/blas_test.hpp
```diff
@@ -120,8 +120,16 @@ template <typename scalar_t>
...
+  if constexpr (std::is_same_v<scalar_t, cl::sycl::half>) {
+    float rangeMinF(rangeMin);
+    float rangeMaxF(rangeMax);
+    std::uniform_real_distribution<float> dis(rangeMinF, rangeMaxF);
+    float temp = dis(gen);
+    return (static_cast<cl::sycl::half>(temp));
+  } else {
+    std::uniform_real_distribution<scalar_t> dis(rangeMin, rangeMax);
+    return dis(gen);
+  }
```
pgorlani, line 132:
```suggestion
  using random_scalar_t =
      std::conditional_t<std::is_same_v<scalar_t, cl::sycl::half>, float, scalar_t>;
  std::uniform_real_distribution<random_scalar_t> dis(rangeMin, rangeMax);
  return dis(gen);
}
```

benchmark/rocblas/blas3/gemm.cpp
OuadiElfarouki: Yes will push these in a bit.

src/operations/blas3/gemm_common.hpp
```diff
@@ -42,12 +42,22 @@ mul_add(T a, T b, T c,
...
 
+#ifndef BLAS_ENABLE_HALF
 template <typename T>
 static PORTBLAS_INLINE T
 mul_add(T a, T b, T c,
         typename std::enable_if<is_sycl_scalar<T>::value>::type * = 0) {
   return (cl::sycl::mad(a, b, c));
 }
+#else
+template <typename T>
```
pgorlani, line 53: If you remove the `BLAS_ENABLE_HALF` guard from `is_half` in `include/blas_meta.h`, you can also remove this `#ifndef` pragma keeping just the `false` case.

src/interface/gemm_interface.hpp
```diff
@@ -55,6 +55,14 @@ inline typename std::enable_if<is_sycl_scalar<T>::value, bool>::type isZero(
   return (value == static_cast<T>(0));
 }
 
+#ifdef BLAS_ENABLE_HALF
```
pgorlani, line 58: If you remove the BLAS_ENABLE_HALF guard from is_half in include/blas_meta.h, you can also remove this one.

src/interface/blas3/backend/amd_gpu.hpp
```diff
@@ -33,8 +33,14 @@ namespace backend {
 template <bool _t_a, bool _t_b, bool s_a, bool s_b, bool is_beta_zero,
           typename sb_handle_t, typename container_0_t, typename container_1_t,
           typename container_2_t, typename element_t, typename index_t>
+#ifndef BLAS_ENABLE_HALF
 typename std::enable_if<is_sycl_scalar<element_t>::value,
                         typename sb_handle_t::event_t>::type
+#else
+typename std::enable_if<is_half<element_t>::value ||
```
pgorlani, line 40: If you remove the `BLAS_ENABLE_HALF` guard from `is_half` in `include/blas_meta.h`, you can also remove this `#ifndef` pragma keeping just the false case.

OuadiElfarouki: Yes that's a good idea thank you!

benchmark/cublas/utils.hpp
```diff
@@ -274,6 +276,23 @@ static inline std::tuple<double, double> timef_cuda(function_t func,
...
+  using type = T;
+};
+
+#ifdef BLAS_ENABLE_HALF
+// When T is sycl::half, use cuda's __cuda as type.
+template <typename T>
+struct CudaType<T, std::enable_if_t<std::is_same_v<T, cl::sycl::half>>> {
+  using type = __half;
+};
+#endif
```
pgorlani, line 294:
```suggestion
// When T is sycl::half, use cuda's __cuda as type.
template <typename T>
struct CudaType<T, std::enable_if_t<std::is_same_v<T, cl::sycl::half>>> {
  using type = __half;
};
```
I think it is possible to remove also this guard.

benchmark/rocblas/utils.hpp
```diff
@@ -373,6 +374,22 @@ static inline std::tuple<double, double> timef_hip(function_t func,
...
+  using type = T;
+};
+
+#ifdef BLAS_ENABLE_HALF
+// When T is sycl::half, use rocBLAS's rocblas_half as type.
+template <typename T>
+struct RocblasType<T, std::enable_if_t<std::is_same_v<T, cl::sycl::half>>> {
+  using type = rocblas_half;
+};
+#endif
```
pgorlani, line 392:
```suggestion
// When T is sycl::half, use rocBLAS's rocblas_half as type.
template <typename T>
struct RocblasType<T, std::enable_if_t<std::is_same_v<T, cl::sycl::half>>> {
  using type = rocblas_half;
};
```

common/include/common/common_utils.hpp
```diff
@@ -1626,6 +1626,13 @@ inline std::string get_type_name<double>() {
   return "double";
 }
 
+#ifdef BLAS_ENABLE_HALF
+template <>
+inline std::string get_type_name<cl::sycl::half>() {
+  return "half";
+}
+#endif
```
pgorlani, line 1634:
```suggestion
template <>
inline std::string get_type_name<cl::sycl::half>() {
  return "half";
}
```

common/include/common/common_utils.hpp
```diff
@@ -1656,8 +1661,16 @@ template <typename scalar_t>
 static inline scalar_t random_scalar(scalar_t rangeMin, scalar_t rangeMax) {
   static std::random_device rd;
   static std::default_random_engine gen(rd());
-  std::uniform_real_distribution<scalar_t> dis(rangeMin, rangeMax);
-  return dis(gen);
+  if constexpr (std::is_same_v<scalar_t, cl::sycl::half>) {
```
pgorlani, line 1664: I think that this function needs to be consistent with the `test/blas_test.hpp` one.

OuadiElfarouki: yes will update it accordingly.

common/include/common/float_comparison.hpp
```diff
@@ -32,7 +32,7 @@
 #include <complex>
 #endif
 
-#ifdef BLAS_DATA_TYPE_HALF
+#ifdef BLAS_ENABLE_HALF
```
pgorlani, line 35: I think that all the `BLAS_ENABLE_HALF` guards in this file can be safely removed.

OuadiElfarouki: Agree, there were some that were already there. I guess we can remove them as well.

SomeoneSerge: Whoopsie! Thanks!

pgorlani (approved):

s-Nick (approved):

PR #497, opened 2024-02-16
FIx error when building with oneAPI release compiler

muhammad-tanvir-1211:
This PR fixes an error seen from `googlebench` and `gtest` (warning in this case) that explicit comparison with NaN is not allowed in fast precision mode.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -83,6 +83,15 @@ if (MSVC)
   )
 endif()
 
+# set -ffp-model to precise to avoid getting error statement
+# in case of icpx compiler: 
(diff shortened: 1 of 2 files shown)


pgorlani:
If the purpose of the patch is removing the warning during the unit test/benchmark compilation, it would be better to set ` -Wno-tautological-constant-compare`  in `test/CMakeLists.txt` and `benchmark/portblas/CMakeLists.txt`.

`-ffp-model=precise` flag is an umbrella option that could have side consequences on many aspects since it "disables optimizations that are not value-safe on floating-point data".

pgorlani (approved):

PR #500, opened 2024-02-29
Extended Gemm interface to support mixed precision operations

OuadiElfarouki:
Extend Gemm operator interface to support mixed precision operations, namely by decoupling matrix A and B type `element_in_t` from output matrix C and scalars alpha and beta type `element_out_t`. 

Following oneMKL's spec notation for Gemm API : https://spec.oneapi.io/versions/latest/elements/oneMKL/source/domains/blas/gemm.html#onemkl-blas-gemm, this PR enables (Ta==Tb) to be set independently from (Tc==Ts). This feature has been enabled at a first stage for Ta=Tb=sycl::half and Tc=Ts=float. Thus enabling half support also enables the mixed precision case of (half, float) for gemm. 

Changes include:
- Updating different Gemm kernel implementations to account for the decoupled types.
- Necessary CMake and Kernel generation scripts updates to account for the **couple of types** instead of **single type** in gemm case.
- Necessary changes to unit-tests to account for this feature in the Gemm case.

Note : 
Following oneMKL expected Gemm API, Support of `bfloat16-float ` and `std::int8_t-float` would be straightforward afterwards, but the additional cases of `Ta==Tb==Tc` while `Ts` _(alpha & beta)_ is separate will require additional decoupling & re-design work..


src/operations/blas3/gemm_local_joint_matrix.hpp
```diff
@@ -62,6 +62,7 @@ namespace blas {
  * @tparam TransA  iff true, matrix A will be transposed on the fly
  * @tparam TransB  iff true, matrix B will be transposed on the fly
  * @tparam element_t  type of matrix elements
+ * @tparam element_out_t  type of output matrix elements (C) and scalars
```
hjabird, line 65: I'd also like to see `element_t` renamed to `element_in_t`.

src/sb_handle/portblas_handle.hpp
```diff
@@ -336,7 +339,7 @@ inline typename SB_Handle::event_t SB_Handle::execute(
   /* Create the cube buffer that will hold the output of the partial gemm */
   auto cube_buffer = acquire_temp_mem < is_usm ? helper::AllocType::usm
                                                : helper::AllocType::buffer,
-       element_t > (rows * cols * depth);
+       element_out_t > (rows * cols * depth);
```
hjabird, line 342: I know this code that's being updated, but it looks like a type is being compared to a value!? What's going on?

test/blas_test_macros.hpp
```diff
@@ -157,7 +204,8 @@
                                         name_generator);
 #endif  // BLAS_ENABLE_COMPLEX
 
-/** Registers test for all supported data types
+/** Registers test for all supported data
+ * types
```
hjabird, line 208: Why this change?

test/unittest/blas3/blas3_gemm_common.hpp
```diff
@@ -86,7 +86,14 @@ inline std::vector<scalar_t> interleaved_to_strided(
   return output;
 }
 
-template <typename scalar_t, helper::AllocType mem_alloc>
+/**
+ * @brief verify gemm correctness against reference BLAS.
+ *
+ * @tparam scalar_in_t  type of input matrices elements (A, B)
+ * @tparam scalar_t  type of output matrix elements (C) and scalars
+ * (gemm_arguments)
```
hjabird, line 94: `scalar_out_t`?

src/operations/blas3/gemm_interleaved.hpp
```diff
@@ -486,28 +493,41 @@ class Gemm<input_t, output_t, /* DoubleBuffer = */ false, /* NbcA = */ false,
...
       for (int j = 0; j < item_rows; ++j) {
 #pragma unroll
         for (int b = 0; b < item_batchs / VectorSize; ++b) {
 #ifdef __ADAPTIVECPP__
-          if constexpr (is_half<element_t>::value) {
-            *reg_res = reg_a[j * (item_batchs / VectorSize) + b] *
-                           reg_b[i * (item_batchs / VectorSize) + b] +
-                       *reg_res;
+          if constexpr (is_half<element_in_t>::value ||
+                        !std::is_same_v<element_in_t, element_out_t>) {
```
s-Nick, line 505: you check opposite condition, why not checking the same condition and keeping the order of if/else branch code. I think it would be more readable.

src/operations/blas3/gemm_interleaved.hpp
```diff
@@ -486,28 +493,41 @@ class Gemm<input_t, output_t, /* DoubleBuffer = */ false, /* NbcA = */ false,
...
+#pragma unroll
+            for (int v = 0; v < VectorSize; ++v) {
```
s-Nick, line 507: Can you elaborate why this loop is necessary?

s-Nick (commented):
Thank you for you work @OuadiElfarouki.
There is some confusion with `element_in_t` and `element_out_t` because you forget to update `element_t` to one or the other. Can you please fix it? @hjabird highlighted many of them. In general, I agree that something like `in_element_t` and `out_element_t` would be more readable, please consider it.

src/sb_handle/portblas_handle.hpp
hjabird: @s-Nick has explained this to me.

cmake/CmakeFunctionHelper.cmake
```diff
@@ -702,7 +722,17 @@ else() # default cpu backend
...
-  endforeach()
+  
+    if(BLAS_ENABLE_HALF)
+      add_gemm_configuration(
+        "half"  128 "false" "false" "false"
+        64 4 4 8 8 1 1 1 1 1 1 1 1 1 float float "no_local" "standard" "full" 1 "strided" "false" "false")
+      add_gemm_configuration(
+        "half" 64 "false" "false" "false"
+        64 2 2 4 4 1 1 1 1 4 4 1 1 1 float float "no_local" "standard" "full" 4 "interleaved" "false" "false")
+    endif()
```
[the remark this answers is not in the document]
hjabird: @s-Nick has explained this to me via other channels.

src/operations/blas3/gemm_interleaved.hpp
OuadiElfarouki: Agree @s-Nick. I avoided changing the scalar_t and element_t before to make less changes xD But I agree with you guys @hjabird it would improve readability to have same "format". So will push the renaming changes next.

src/operations/blas3/gemm_interleaved.hpp
OuadiElfarouki: @s-Nick AdaptiveCpp doesn't support sycl::vec<sycl::half> for mad operation, so had to break it down to its elements through the loop

include/operations/blas3_trees.h
```diff
@@ -169,7 +169,8 @@ struct Tile {
  * @tparam TransB  iff true, matrix B will be transposed on the fly
  * @tparam SymmA   whether the matrix A is a symmetric triangular matrix
  * @tparam SymmB   whether the matrix B is a symmetric triangular matrix
- * @tparam element_t  type of matrix elements
+ * @tparam element_in_t  type of input matrix elements (A & B)
+ * @tparam element_t  type of output matrix elements (C) and scaling parameters
```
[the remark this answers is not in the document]
OuadiElfarouki: Thanks for the hint @pgorlani, changes have been reduced now!

src/operations/blas3/gemm_common.hpp
```diff
@@ -42,13 +42,22 @@ mul_add(T a, T b, T c,
...
+                                is_sycl_scalar<Tin>::value>::type * = 0) {
   return (cl::sycl::mad(a, b, c));
 }
 
+template <typename Tin, typename Tout>
+static PORTBLAS_INLINE Tout
+mul_add(Tin a, Tin b, Tout c,
+        typename std::enable_if<!std::is_same<Tin, Tout>::value &&
+                                is_sycl_scalar<Tin>::value>::type * = 0) {
+  return static_cast<Tout>(a) * static_cast<Tout>(b) + c;
```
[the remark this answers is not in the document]
OuadiElfarouki: @hjabird yes indeed, it should be actually done in `Tin` but I had to fall back to `Tout` because gemm unit-tests failed on the oneMKL side (portBLAS backend) due to precision mismatches. This needs to be investigated though I agree, as casting to float beats the purpose of having half mul.

src/operations/blas3/gemm_local_joint_matrix.hpp
OuadiElfarouki: The extra template parameter has been removed.

include/operations/blas3_trees.h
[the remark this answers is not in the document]
pgorlani: Thank you so much, @OuadiElfarouki!

test/blas_test_macros.hpp
OuadiElfarouki: Formatting mistake, will revert it!

src/operations/blas3/gemm_common.hpp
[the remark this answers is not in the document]
hjabird: Do we have any ticket or issue to track that this needs to be investigated?

test/blas_test_macros.hpp
hjabird: Fixed in [c7aedc9](https://github.com/codeplaysoftware/portBLAS/pull/500/commits/c7aedc965dae663ae2ee9c171f69302c38a4d092).

test/unittest/blas3/blas3_gemm_common.hpp
hjabird: Appears to be done.

src/operations/blas3/gemm_common.hpp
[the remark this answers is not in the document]
OuadiElfarouki: @hjabird not that I know of no, I've put a placeholder [oneMKL PR](https://github.com/OuadiElfarouki/oneMKL/pull/1) locally that we can open once the portBLAS PRs are merged. Something similar was raised in this [PR](https://github.com/oneapi-src/oneMKL/pull/466/files#r1552523436) regarding the test precision tolerance. So basically if the precision on oneMKL tests is handled loosely for half/mixed operations, we can drop this static_cast on portBLAS side as it only causes failures on oneMKL tests not on our portBLAS unittests.

test/unittest/blas3/blas3_gemm_common.hpp
OuadiElfarouki: @hjabird yes done!

OuadiElfarouki:
@s-Nick @Rbiessy @hjabird I've rebased and fixed some bugs in the current PR after recent commits. Feel free to check it ! Thanks.

hjabird (approved):

PR #501, opened 2024-02-29
Minor Update to CMake configurations 

s-Nick:
This patch provides two updates to cmake configuration.

- Check if the repository was cloned correctly. This repository must be cloned *recursively* otherwise the project cannot be built and giving back a difficult to decode *compile-time* error. This patch adds a check for the submodule during the configuration, blocking the configuration process in case of missing submodule. Users are provided with a hint on how to fix the issue.

- RocBlas benchmark suite requires `libclang_rt.builtins` library to build. This library is not provided by default by `clang++` so it might be missing. This would block rocBlas benchmarks to build but no error is emitted while configuring the project. This patch adds a check for the library and it blocks the configuration if it is missing. Users can decide to build the library or use a compatible library already available in the system.

- Remove linking dependency to portBLAS from cuBLAS benchmark.


CMakeLists.txt
```diff
@@ -73,6 +73,10 @@ if (NOT INSTALL_HEADER_ONLY)
   include(ConfigurePORTBLAS)
   include(SYCL)
   find_package(PythonInterp 3 REQUIRED)
+  if(NOT EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/external/py_gen/README.md" )
```
aacostadiaz, line 76: Nice!!!

aacostadiaz (approved):
LGTM

muhammad-tanvir-1211 (approved):

PR #504, opened 2024-03-07
Fixed issues encountered through oneMKL portBLAS backend

OuadiElfarouki:
This patch includes following changes : 

- Re-enabled half data type support for CPUs the same way as the other devices _(half support no longer disabled when picking tuning_target=default_cpu)_
- Reverted changes (introduced in PR [493](https://github.com/codeplaysoftware/portBLAS/pull/493)) to the `BufferIterator<T>::get_range_accessor()` that caused some accessors to be **not initialized** (`sycl::property::no_init{}`) and thus yield erroneous test results. 
- Modify `WGAtomicReduction` signature to comply with different usm memory allocation and atomic initialization. Adding `usmManagedMem` to flag if memory is allocated using managed memory which requires `generic_space` in atomic initialization, especially on AMD hardware. Flag added  not to cause performance degradation to the operators that use this kernel.


samples/CMakeLists.txt
```diff
@@ -1,3 +1,7 @@
+cmake_minimum_required(VERSION 3.4.3)
```
muhammad-tanvir-1211, line 1: This file seems to be missing the license text, could you please add that as well?

include/interface/blas1_interface.h
```diff
@@ -136,9 +136,9 @@ typename sb_handle_t::event_t _asum(
  * \brief Prototype for the internal implementation of the ASUM operation. See
  * documentation in the blas1_interface.hpp file for details.
  */
-template <int localSize, int localMemSize, typename sb_handle_t,
-          typename container_0_t, typename container_1_t, typename index_t,
-          typename increment_t>
+template <int localSize, int localMemSize, bool usmManagedMem = false,
```
Rbiessy, line 139: Can you add a short doxygen documentation for `usmManagedMem` and why it's needed everywhere it's used?

src/interface/blas3/backend/default_cpu.hpp
```diff
@@ -120,6 +120,48 @@ _gemm(sb_handle_t& sb_handle, index_t _M, index_t _N, index_t _K,
...
+typename std::enable_if<is_half<element_t>::value,
+                        typename sb_handle_t::event_t>::type
+_gemm(sb_handle_t& sb_handle, index_t _M, index_t _N, index_t _K,
+      element_t _alpha, container_0_t _a, index_t _lda, index_t _stridea,
+      container_1_t _b, index_t _ldb, index_t _strideb, element_t _beta,
+      container_2_t _c, index_t _ldc, index_t _stridec, index_t batch_size,
+      gemm_batch_type_t batch_type,
+      const typename sb_handle_t::event_t& _dependencies) {
+  // Unused configuration cases
+  if constexpr (s_a || s_b) {
```
pgorlani, line 136: Thanks for your PR, @OuadiElfarouki.

I think you are excluding the symmetric matrices because they are not considered in case of the half datatype.

If this is the case, could you please add a comment specifying what are the supported routines?

OuadiElfarouki: Indeed  @pgorlani. We have the option to disable building those kernels for CPU which we know doesn't support `fp16`, but when using oneMKL we can't directly do that given the header only integration within oneMKL and the fact that `aspect::fp16` is only known at runtime. This means that the added Gemm configurations for the default CPU target are merely a temporary **placeholder** (until such support is brought to CPUs eventually), and they were consequently never tested as they're skipped by the unit-tests. This change comes in-line with the patch on oneMKL (not published yet) : https://github.com/OuadiElfarouki/oneMKL/pull/1

include/interface/blas1_interface.h
s-Nick: I added a small description for it in `src/interface/blas1_interface.hpp` along with the rest of the documentation.
Changes in f89fd4b55da6f4f92bb72c5db8f3b097cca7b7bd

src/interface/blas3/backend/nvidia_gpu.hpp
```diff
@@ -188,7 +188,8 @@ _gemm(sb_handle_t& sb_handle, index_t _M, index_t _N, index_t _K,
       container_2_t _c, index_t _ldc, index_t _stridec, index_t batch_size,
       gemm_batch_type_t batch_type,
       const typename sb_handle_t::event_t& _dependencies) {
-  // Unused configuration cases
+  // The symmetric matrice(s) cases are not enabled with half (fp16) as the Symm
+  // operator's specification doesn't include half floating point data.
   if constexpr (s_a && s_b || ((s_a && _t_b) || (s_b && _t_a))) {
```
pgorlani, line 193: Probably I missing something,  why do you want to enable the cases in which there is a symmetric matrix and a non-transposed one?

Shouldn't it be `s_a || s_b`?

src/interface/blas1_interface.hpp
```diff
@@ -219,16 +219,30 @@ typename sb_handle_t::event_t _asum(
...
+ *                       otherwise it will rely on pcie atomics
+ *                       which we cannot enforce, guarantee or check due to its
+ *                       hardware nature. Other targets do not have the same
+ *                       strong dependency and managed memory is handled
+ *                       correctly in any case by default. It is automatically
+ *                       initialized to false to reduce verbosity of
+ *                       initialization for many targets since only one of them,
+ *                       with specific allocation type, requires a different
+ *                       value. Having a default value allows the compiler to
+ *                       handle automatically other templates.
```
hjabird, line 241:
```suggestion
 * @tparam usmManagedMem Specifies if usm memory allocation is automatically
 *                       managed  or not. Automatically managed memory
 *                       requires that atomic address space is set to generic.
 *                       This is a strict requirement only for AMD GPUs, since
 *                       AMD's implementation of atomics may depend on specific
 *                       hardware configurations (PCIe atomics) that cannot be checked at runtime.
 *                       Other targets do not have the same
 *                       strong dependency and managed memory is handled
 *                       correctly by default.
```

I don't think you need to justify your choice of having a default in the documentations.

src/operations/blas1/WGAtomicReduction.hpp
```diff
@@ -119,22 +126,38 @@ WGAtomicReduction<operator_t, lhs_t, rhs_t>::eval(sharedT scratch,
...
+              lhs_.get_data()[0]);
+      atomic_res += val;
+    } else {
+      auto atomic_res =
+          cl::sycl::atomic_ref<value_t, cl::sycl::memory_order::relaxed,
+                               cl::sycl::memory_scope::device,
+                               cl::sycl::access::address_space::generic_space>(
+              lhs_.get_data()[0]);
+      atomic_res += val;
+    }
```
pgorlani, line 143: Maybe, we can simplify things as follows.

```suggestion
    constexpr cl::sycl::access::address_space addr_sp =
        usmManagedMem ? cl::sycl::access::address_space::generic_space
                      : cl::sycl::access::address_space::global_space;
    auto atomic_res =
        cl::sycl::atomic_ref<value_t, cl::sycl::memory_order::relaxed,
                             cl::sycl::memory_scope::device, addr_sp>(
            lhs_.get_data()[0]);
    atomic_res += val;
```

src/interface/blas3/backend/default_cpu.hpp
pgorlani: Hi @OuadiElfarouki, it seems that the portBLAS backend for oneMKL checks for `aspect::f16`, see https://github.com/oneapi-src/oneMKL/blob/3339418fa1318377c65815e6f85aced9efa46e9e/src/blas/backends/portblas/portblas_common.hpp#L196.

src/interface/blas1_interface.hpp
s-Nick: Thank you @hjabird I am still new to writing documentation. I will fix it.

src/operations/blas1/WGAtomicReduction.hpp
s-Nick: Thank you @pgorlani, written this way it looks cleaner.

src/interface/blas1/backend/amd_gpu.hpp
```diff
@@ -101,16 +134,43 @@ template <typename sb_handle_t, typename container_0_t, typename container_1_t,
...
+#ifdef SB_ENABLE_USM
+  bool usm_managed_mem{false};
+  if constexpr (std::is_pointer_v<decltype(_rs)>) {
+    usm_managed_mem =
+        sycl::usm::alloc::shared ==
+        sycl::get_pointer_type(_rs, sb_handle.get_queue().get_context());
+  }
+#else
+  constexpr bool usm_managed_mem{false};
+#endif
```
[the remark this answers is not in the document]
s-Nick: Thank you @hjabird for your review.
I agree with you that adding a single function to `blas::helper` makes the code cleaner. However, I have one concern regarding the documentation. I don't think I can use the same I used for `_asum` because in that case, I specifically discuss the `reduction operator`, and outside of that context, it seems out of place to me. Instead, I prefer to create the function elsewhere but leave a comment above this code snippet and reference `nrm2` and `dot` to it.

To clarify, the function will be something like:

```c++
template <typename sb_handle_t, typename containerT>
inline bool is_malloc_shared(sb_handle_t &sb_handle,
                                   const containerT _rs) {
  if constexpr (std::is_pointer_v<decltype(_rs)>) {
    return sycl::usm::alloc::shared ==
           sycl::get_pointer_type(_rs, sb_handle.get_queue().get_context());
  }else {
    return false;
  }
}
```

As you can see, this function is generic, so adding a comment referring to atomics and reduction doesn't seem appropriate to me. What do you think?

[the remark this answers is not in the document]
hjabird: Nice! I'd be tempted to have two implementations of the function, one of which is constexpr for when `SB_ENABLE_USM` is undefined.

I agree that the comment that refers to AMD's atomics would be out of place. Perhaps we can put it with the if statement where the difference is? I'm not sure how to avoid the repetition of the comment in a good way. If we repeat it, maybe it can be shortened.

src/interface/blas1/backend/amd_gpu.hpp
```diff
@@ -34,16 +34,49 @@ template <typename sb_handle_t, typename container_0_t, typename container_1_t,
...
+      constexpr index_t localSize = 1024;
+      const index_t number_WG = (_N + localSize - 1) / localSize;
+      return blas::internal::_asum_impl<static_cast<int>(localSize), 32, false>(
+          sb_handle, _N, _vx, _incx, _rs, number_WG, _dependencies);
+    } else {
+      constexpr int localSize = 512;
+      constexpr index_t number_WG = 256;
+      return blas::internal::_asum_impl<localSize, 32, false>(
+          sb_handle, _N, _vx, _incx, _rs, number_WG, _dependencies);
+    }
```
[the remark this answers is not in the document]
s-Nick: You are correct, apart from `true` and `false`, they are identical.
I don't think we can deduplicate it. We need to have both functions generated by the compiler because we need the kernel in both situations since only at runtime we discover how the memory was allocated and call the appropriate function. 
Ideally,  `usm_managed_mem` could be used, but if shared memory is enabled, its value is set at runtime so it cannot be used as a template argument. In my opinion, to deduplicate this we should move `usmManagedMem` from template parameter to function parameter, but doing this we end up with a branch inside kernel code.
If I am missing something about template instantiation, please point me in the right direction.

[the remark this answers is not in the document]
hjabird: For the reasons you say, I don't see a better solution than this either. Thanks for looking into it.

src/interface/blas1/backend/amd_gpu.hpp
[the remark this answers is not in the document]
s-Nick: I moved the function into `blas::helper` and update the comment in 5713422f46b3050266aaf6b705d42d47f72e7310

src/interface/blas1_interface.hpp
s-Nick: Fixed in 5713422f46b3050266aaf6b705d42d47f72e7310

include/container/sycl_iterator.h
```diff
@@ -194,51 +194,27 @@ template <cl::sycl::access::mode acc_md_t>
...
-  if constexpr (acc_md_t == cl::sycl::access::mode::read) {
-    return typename BufferIterator<element_t>::template accessor_t<acc_md_t>(
-        buffer_, cgh, cl::sycl::range<1>(size),
-        cl::sycl::id<1>(BufferIterator<element_t>::get_offset()));
-  } else {
-    // Skip data initialization if not accessing in read mode only
-    return typename BufferIterator<element_t>::template accessor_t<acc_md_t>(
-        buffer_, cgh, cl::sycl::range<1>(size),
-        cl::sycl::id<1>(BufferIterator<element_t>::get_offset()),
-        cl::sycl::property::no_init{});
```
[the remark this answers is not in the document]
OuadiElfarouki: @hjabird This was actually added to get rid of some warnings regarding buffers initialization when using AdaptiveCpp, but introducing it caused some tests to fail through the portBLAS backend API in oneMKL, that's why we're reverting. No performance aspects have been tested/verified but this change was tested with AdaptiveCpp (warnings are back but no errors/tests failures).

samples/CMakeLists.txt
```diff
@@ -1,3 +1,31 @@
...
+# *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# *  See the License for the specific language governing permissions and
+# *  limitations under the License.
+# *
+# *  portBLAS: BLAS implementation using SYCL
+# *
+# *  @filename CMakeLists.txt
+# *
+# **************************************************************************/
+cmake_minimum_required(VERSION 3.4.3)
```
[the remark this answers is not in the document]
OuadiElfarouki: Yes I just used the same version required by the main CMakeLists.txt, as I got some warnings/errors if I still remember when building the samples alone (header only with portBLAS)

Rbiessy (approved):

pgorlani (approved):

PR #506, opened 2024-03-22
Document difference in rotmg and rog references

Rbiessy:
Document the difference in rotmg and rotg operators as some reference libraries can give different output for these 2 operations. This has been discussed here before for rotmg: https://github.com/codeplaysoftware/portBLAS/pull/376. I remember rotg having similar issues.

Solves https://github.com/codeplaysoftware/portBLAS/issues/502


Rbiessy:
Closing as I was mistaken. We will re-visit this differently.

PR #512, opened 2024-04-11
Update compiler to support icpx and latest nightly

s-Nick:
This patch address compilation issues with `icpx` for AMD and NVIDIA target.  

It add 3 flags if and only if the project is compiled using `icpx` or `clang++` from any oneAPI release and the target is not and Intel device. The flags added force `icpx` to behave as the open source *intel/llvm* compiler, making the project works correctly.

This patch also update shuffle operation API to be compatible with the latest *intel/llvm* commit to allow the project to be compiled also with it.


test/unittest/CMakeLists.txt
```diff
@@ -143,7 +143,8 @@ foreach(blas_test ${SYCL_UNITTEST_SRCS})
   target_include_directories(${test_exec} PRIVATE ${CBLAS_INCLUDE} ${PORTBLAS_COMMON_INCLUDE_DIR})
 
   list (FIND SYCL_UNITTEST_NOFASTMATH ${blas_test} _index)
```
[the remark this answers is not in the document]
s-Nick: Thank you for you review @Rbiessy.
Linked to following comment. 
If the compiler used is not from Intel, the `CMAKE_CXX_COMPILER_ID` would be different hence the flag is not added. That's why I kept this part of the configuration.

cmake/Modules/FindDPCPP.cmake
```diff
@@ -96,6 +97,11 @@ function(add_sycl_to_target)
     "${multi_value_args}"
     ${ARGN}
   )
+  if((${CMAKE_CXX_COMPILER_ID} STREQUAL "IntelLLVM") AND
+      (TUNING_TARGET IN_LIST non_intel_target) )
+    target_compile_options(${SB_ADD_SYCL_TARGET} PRIVATE -fno-fast-math)
```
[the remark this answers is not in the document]
s-Nick: I don't think that having a different flag for nightly and release would be an issue, since it is transparent to users and, if they want, they can always add it themself. Moreover, I am not a fan of adding an unnecessary flag that may limit optimization to a compiler and this particular flag is easier to add that to remove while configuring the project build.
If in future nightly this become an issue we can add it later, what do you think?

I am happy to add a message when we add this flag.

cmake/Modules/FindDPCPP.cmake
```diff
@@ -96,6 +97,11 @@ function(add_sycl_to_target)
...
+    target_compile_options(${SB_ADD_SYCL_TARGET} PRIVATE -mllvm -loopopt=0 )
```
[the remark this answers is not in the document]
s-Nick: I agree, it is more future/feature proof to invert this check and add a message. With `DEFAULT_CPU` the optimization seems to work, should I stick to `INTEL_GPU` only or keep them enable also for CPU?

CMakeLists.txt
```diff
@@ -144,6 +144,12 @@ if (INSTALL_HEADER_ONLY)
   set_target_properties(portblas PROPERTIES
     INTERFACE_INCLUDE_DIRECTORIES "${PORTBLAS_INCLUDE};$<INSTALL_INTERFACE:src>"
   )
+  set(non_intel_target "AMD_GPU" "NVIDIA_GPU")
+  if((${CMAKE_CXX_COMPILER_ID} STREQUAL "IntelLLVM") AND
```
hjabird, line 148: Stupid question: `IntelLLVM` mean the closed source or the open-source `intel/llvm`?

hjabird (approved):
Looks good to me.

cmake/Modules/FindDPCPP.cmake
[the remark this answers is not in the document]
Rbiessy: We agreed it makes sense to add `-fno-fast-math` only when we also add `-mllvm -loopopt=0` i.e. for every target except `INTEL_GPU`.

CMakeLists.txt
s-Nick: Thank you for your review @hjabird 
`IntelLLVM` is the closed source. Cmake uses this name when referring to `icpx` or `clang++` from the oneAPI release, while the open source compiler is identified simply as `Clang`

cmake/Modules/FindDPCPP.cmake
```diff
@@ -96,6 +96,12 @@ function(add_sycl_to_target)
     "${multi_value_args}"
     ${ARGN}
   )
+  if((${CMAKE_CXX_COMPILER_ID} STREQUAL "IntelLLVM") AND NOT
+    (${TUNING_TARGET} STREQUAL "INTEL_GPU") )
+    target_compile_options(${SB_ADD_SYCL_TARGET} PRIVATE -fno-fast-math)
+    target_compile_options(${SB_ADD_SYCL_TARGET} PRIVATE -mllvm -loopopt=0 )
+    message(STATUS "Adding -fno-fast-math -mllvm -loopopt=0 to target ${SB_ADD_SYCL_TARGET}")
```
Rbiessy, line 103: Can you also add this message when portBLAS is used in header-only mode?

CMakeLists.txt
```diff
@@ -99,7 +99,11 @@ set(PORTBLAS_GENERATED_SRC ${CMAKE_CURRENT_BINARY_DIR}/generated_src)
 set(PORTBLAS_INCLUDE $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
   $<INSTALL_INTERFACE:include>)
 set(PORTBLAS_COMMON_INCLUDE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/common/include)
-set(PORTBLAS_SRC ${CMAKE_CURRENT_SOURCE_DIR}/src)
+if(INSTALL_HEADER_ONLY)
+  set(PORTBLAS_SRC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src> $<INSTALL_INTERFACE:src>)
+else()
+  set(PORTBLAS_SRC ${CMAKE_CURRENT_SOURCE_DIR}/src)
+endif()
```
Rbiessy, line 106: This looks a bit odd. I think it would work if you use `$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src> $<INSTALL_INTERFACE:src>` in this file below instead of `${PORTBLAS_SRC}`. I think only one of these 2 expressions is needed even.

CMakeLists.txt
```diff
@@ -142,8 +142,17 @@ include(CmakeFunctionHelper)
 if (INSTALL_HEADER_ONLY)
   add_library(portblas INTERFACE)
   set_target_properties(portblas PROPERTIES
-    INTERFACE_INCLUDE_DIRECTORIES "${PORTBLAS_INCLUDE};$<INSTALL_INTERFACE:src>"
+    INTERFACE_INCLUDE_DIRECTORIES "${PORTBLAS_INCLUDE};${PORTBLAS_SRC}"
   )
+  target_compile_definitions(portblas INTERFACE "SB_ENABLE_USM")
```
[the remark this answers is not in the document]
s-Nick: I don't know exactly why it was added there, I think that adaptiveCpp backend was broken when USM support was initially introduced. I can run some tests and understand the reason for those lines.

CMakeLists.txt
```diff
@@ -142,8 +142,17 @@ include(CmakeFunctionHelper)
...
+  target_compile_definitions(portblas INTERFACE "BLAS_ENABLE_COMPLEX=${BLAS_ENABLE_COMPLEX}")
```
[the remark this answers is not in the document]
s-Nick: If its definition always enable the feature I think we should a check before adding in to header only compilation. Complex support is quite a niche feature, still incomplete in my opinion, and adding it by default to header only compilation will increase compile time. What do you think?
Thank you for the function suggested I will look into it.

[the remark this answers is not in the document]
Rbiessy: Yes I was suggesting to add a check before defining `BLAS_ENABLE_COMPLEX`, like `set_target_compile_def` is doing.

CMakeLists.txt
s-Nick: The current situation is a little bit more tricky.
The reason why in oneMKL portblas configuration there is a function to reset include path [here](https://github.com/oneapi-src/oneMKL/blob/develop/src/blas/backends/portblas/CMakeLists.txt#L156-L159) is that using only `$<INSTALL_INTERFACE:src>`, as it was before, doesn't work. Moreover, there is a little mismatch also in installing the library in header only mode and using with `FetchContent`. The problem of using only `line 103` introduced in this PR is that `${PORTBLAS_SRC}` variable is used in `generate_blas_object()` and it fails if it uses `line 103` solution. I am not sure why but it seems it doesn't expand it correctly.

Rbiessy: But are you sure that `PORTBLAS_SRC` is used in multiple places when we set `INSTALL_HEADER_ONLY`? It looks to me that this [line](https://github.com/codeplaysoftware/portBLAS/pull/512/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR218) is the only place.

Functions like `generate_blas_object` expect this variable to be only a single path while the `install` function below can accept multiple paths. In other words it looks to me that we should have 2 different variables and we're trying to merge them in one. Since we need to set `$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src> $<INSTALL_INTERFACE:src>` only once for the `install` function, we might as well don't set a variable for it. Am I missing something?

CMakeLists.txt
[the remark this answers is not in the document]
s-Nick: For minor reasons, USM is not currently supported for AdaptiveCpp so we cannot enable it by default. I will add a check to add `SB_ENABLE_USM` only for dpcpp for now also in header only mode.

CMakeLists.txt
s-Nick: It was used only there, but I tried to use it also in this [line](https://github.com/codeplaysoftware/portBLAS/pull/512/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR149).

Trying to fix it, I ended up with the same conclusion as you. We need two different variables because they serve two different needs. 
I would change above lines in this way:
```suggestion
if(INSTALL_HEADER_ONLY)
  set(PORTBLAS_INSTALL_SRC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src> $<INSTALL_INTERFACE:src>)
endif()
```
Use this new variable for header only purpose and let the other for everything else.

CMakeLists.txt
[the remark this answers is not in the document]
Rbiessy: I can see this is done now

test/unittest/CMakeLists.txt
```diff
@@ -142,8 +135,7 @@ foreach(blas_test ${SYCL_UNITTEST_SRCS})
   target_link_libraries(${test_exec} PRIVATE gtest_main Clara::Clara blas::blas portblas)
   target_include_directories(${test_exec} PRIVATE ${CBLAS_INCLUDE} ${PORTBLAS_COMMON_INCLUDE_DIR})
 
-  list (FIND SYCL_UNITTEST_NOFASTMATH ${blas_test} _index)
-  if (${_index} GREATER -1)
+  if (${CMAKE_CXX_COMPILER_ID} STREQUAL "IntelLLVM")
```
pgorlani, line 138: Thanks for your PR, @s-Nick.

Please, could you add a comment stating that this condition is true just in case of the `icpx` release and not for `intel/llvm`?

CMakeLists.txt
```diff
@@ -142,8 +145,22 @@ include(CmakeFunctionHelper)
...
+    target_compile_definitions(portblas INTERFACE "SB_ENABLE_USM")
+  endif()
+  if(${BLAS_ENABLE_COMPLEX})
+    target_compile_definitions(portblas INTERFACE "BLAS_ENABLE_COMPLEX")
+  endif()
+  target_compile_definitions(portblas INTERFACE ${TUNING_TARGET})
+  if((${CMAKE_CXX_COMPILER_ID} STREQUAL "IntelLLVM") AND NOT
+    (${TUNING_TARGET} STREQUAL "INTEL_GPU") )
+      target_compile_options(portblas INTERFACE -fno-fast-math)
+      target_compile_options(portblas INTERFACE -mllvm -loopopt=0 )
```
pgorlani, line 161: Maybe, it is better use `-mllvm=-loopopt=0` in order to avoid possible issues due to CMake option deduplication. 

*The same apply to `cmake/Modules/FindDPCPP.cmake`.*

test/unittest/CMakeLists.txt
s-Nick: Thank you for your review @pgorlani 
I addressed your comments in this commit 7b2c73a2875ac96e0d649230fd0b68b8f266992e

pgorlani (approved):

Rbiessy (approved):

PR #516, opened 2024-05-07
Set -enable-global-offset=false

pgorlani:
This patch disables the compilation of unused/deprecated global-offset kernels.


s-Nick (approved):
LGTM

hjabird (commented):
Why does this all have to duplicated?

Rbiessy:
> Why does this all have to duplicated?

I think we don't have much choice because when building the library we want to use the `PRIVATE` visibility to only use the flag for portBLAS, in header-only mode this needs to the `INTERFACE` visibility to add this flag to anything using portBLAS.
We could consider having a helper function but I'm not sure it would help that much.

Rbiessy (approved):

PR #517, opened 2024-05-08
fix for DEFAULT TUNING_TARGET on AMD and NVIDIA GPUs

s-Nick:
This PR fixes most of the tests that fails on AMD and NVIDIA GPUs using DEFAULT configuration.
It fixes all of them for AMD and let only `trsm` operator to be fixed for NVIDIA.

In particular it fixes:
- iamax
- iamin
- trsv
- tbsv
- tpsv

**iamax/iamin**:
The `sycl:shift_group_left` api requires all group(sub_group) takes part to the operation, removing the _if-condition_ solves the problem.

**txsv operators**:
broadcast operations inside the kernel require a specific size of group and subgroup, so calling the kernel implementation from `default` is not enough due to hardware differences. This solution uses runtime checks to select the correct template parameters. This leads to compile more kernels than before but from my tests it doesn't affect significantly compilation time.


src/interface/blas2/backend/default.hpp
```diff
@@ -137,8 +137,21 @@ typename sb_handle_t::event_t _trsv(
...
-  return blas::internal::_trsv_impl<4, 2, uplo, trn, diag>(
-      sb_handle, _N, _mA, _lda, _vx, _incx, _dependencies);
+  const auto device = sb_handle.get_queue().get_device();
+  const std::string vendor =
+      device.template get_info<sycl::info::device::vendor>();
+  if (device.is_gpu()) {
+    if (vendor.find("Intel") == vendor.npos) {
+      return blas::internal::_trsv_impl<32, 4, uplo, trn, diag>(
+          sb_handle, _N, _mA, _lda, _vx, _incx, _dependencies);
+    } else {
```
hjabird, line 147:
```suggestion
  if (device.is_gpu()) {
    const std::string vendor =
        device.template get_info<sycl::info::device::vendor>();
    if (vendor.find("Intel") == vendor.npos) {
      return blas::internal::_trsv_impl<32, 4, uplo, trn, diag>(
          sb_handle, _N, _mA, _lda, _vx, _incx, _dependencies);
    } else {
```

src/interface/blas2/backend/default.hpp
```diff
@@ -137,8 +137,21 @@ typename sb_handle_t::event_t _trsv(
...
     container_t1 _vx, increment_t _incx,
     typename sb_handle_t::event_t _dependencies) {
-  return blas::internal::_trsv_impl<4, 2, uplo, trn, diag>(
-      sb_handle, _N, _mA, _lda, _vx, _incx, _dependencies);
+  const auto device = sb_handle.get_queue().get_device();
+  const std::string vendor =
+      device.template get_info<sycl::info::device::vendor>();
+  if (device.is_gpu()) {
+    if (vendor.find("Intel") == vendor.npos) {
+      return blas::internal::_trsv_impl<32, 4, uplo, trn, diag>(
```
hjabird, line 145: Why does the tuning of `trsv_impl` change here?

src/interface/blas2/backend/default.hpp
```diff
@@ -137,8 +137,21 @@ typename sb_handle_t::event_t _trsv(
...
+          sb_handle, _N, _mA, _lda, _vx, _incx, _dependencies);
+    } else {
+      std::cout << "Trsv operator currently not supported on Intel GPUs\n";
```
hjabird, line 148: I'm not a fan of printing issues to `std::cout` but perhaps that's the convention in portBLAS?

src/interface/blas2/backend/default.hpp
```diff
@@ -152,8 +165,21 @@ typename sb_handle_t::event_t _tbsv(
...
+      return blas::internal::_tbsv_impl<32, 4, uplo, trn, diag>(
+          sb_handle, _N, _K, _mA, _lda, _vx, _incx, _dependencies);
+    } else {
+      std::cout << "Tbsv operator currently not supported on Intel GPUs\n";
+      return {};
+    }
+  } else {
+    return blas::internal::_tbsv_impl<4, 2, uplo, trn, diag>(
+        sb_handle, _N, _K, _mA, _lda, _vx, _incx, _dependencies);
+  }
```
hjabird, line 182: Same comments here as already made for `trsv`.

src/interface/blas2/backend/default.hpp
```diff
@@ -163,12 +189,24 @@ namespace backend {
...
           typename container_t1, typename increment_t>
-typename sb_handle_t::event_t _tpsv(sb_handle_t& sb_handle, index_t _N,
-                                    container_t0 _mA, container_t1 _vx,
-                                    increment_t _incx,
-                                    const typename sb_handle_t::event_t& _dependencies) {
-  return blas::internal::_tpsv_impl<4, 2, uplo, trn, diag>(sb_handle, _N, _mA,
-                                                           _vx, _incx, _dependencies);
+typename sb_handle_t::event_t _tpsv(
+    sb_handle_t& sb_handle, index_t _N, container_t0 _mA, container_t1 _vx,
+    increment_t _incx, const typename sb_handle_t::event_t& _dependencies) {
```
hjabird, line 194: This looks like just a formatting change. Was the old version incorrectly formatted?

src/operations/blas1/IndexMaxMin.hpp
```diff
@@ -94,22 +94,14 @@ PORTBLAS_INLINE void IndexMaxMin<is_max, is_step0, lhs_t, rhs_t>::eval(
...
-    if (sg_local_id < i) {
-      element_t shfl_val = sycl::shift_group_left(sg, val.get_value(), i);
-      index_t shfl_idx = sycl::shift_group_left(sg, val.get_index(), i);
-      value_t shfl{shfl_idx, shfl_val};
-      val = op::eval(val, shfl);
-    }
+    element_t shfl_val = cl::sycl::shift_group_left(sg, val.get_value(), i);
+    index_t shfl_idx = cl::sycl::shift_group_left(sg, val.get_index(), i);
+    value_t shfl{shfl_idx, shfl_val};
+    val = op::eval(val, shfl);
```
hjabird, line 102: A classic sub-group mistake! Nice work identifying it.

* This is a reduction. It might be out of scope for this PR, but using SYCL group reduction operations would seem more natural than implementing it by hand.
* Can `op::eval` have side affects when called with garbage? Does it need to be in an `if`?
* I don't see the point in using `cl` here when `shift_group_left` is SYCL-2020.

src/interface/blas2/backend/default.hpp
s-Nick: All the if-statement you see are there to change those templates parameters. The first one set the `subgroup` size used inside the kernel and it used also to set localSize and globalSize  to launch the kernel. Without being certain that this parameter are a multiple of the actual subgroup size on the hardware `txsv` kernels hang on NVIDIA due to group operations like broadcast.
The second template parameter changed is the number of subgroups of the given subgroup size. It can be different but at this point I used the same value of the proper tuned kernel, because it gave the best performance when benchmarked while adding these operators to portBLAS.
Same reason applies to `tpsv` and `tbsv`.

src/interface/blas2/backend/default.hpp
s-Nick: I am not a huge fan too, but I wanted to warn eventual user of it and also I did not like the idea of an exception here. There is not a convention in portBLAS. I think this is the first time we encounter a similar situation, so if you have any suggestion feel free to post it.

src/operations/blas1/IndexMaxMin.hpp
s-Nick: Thank you for your comments @hjabird.

- I think that implementing `iamax/iamin` kernels using SYCL group reduction would require further investigation. In my opinion, the reason for this implementation is that for these operators the kernel must keep together value and initial index in the vector, you cannot call two different reductions and combine the result. A possible solution would be create a vector of tuple (value,position) and call a group reduction but, as I said, I have no idea about performance impact.
- For how the kernel is implemented and launched I don't think that garbage values `op::eval` read and used by it would ever ends up in the final result. Every thread read at least one _good_ value from the starting vector and only the first thread of each subgroup write out the result.
- We have this technical debt in other places and we plan to eventually update everywhere it in the future.

src/interface/blas2/backend/default.hpp
s-Nick: I used `git-clang-format` without checking. I double checked and yes previously it went over the length went over 80 chars.

src/interface/blas2/backend/default.hpp
hjabird: Can you explain why you don't like the idea of an exception here?

src/operations/blas1/IndexMaxMin.hpp
hjabird: That all sounds reasonable to me. Thanks!

src/interface/blas2/backend/default.hpp
s-Nick: As far as I know we don't throw any exception(a part from SYCL one) anywhere else that's a first concern. Secondly, do we throw and catch it here? At this point a message and a solution to not let everything crash is enough imho. Do we throw it here and catch it elsewhere? where? Do we want to catch it?
This kind of questions/ideas are the reason why I did not put an exception here.

src/operations/blas1/IndexMaxMin.hpp
Rbiessy, line 102: Have you checked that this is not going to read out of bounds and potentially fail the iamax or iamin?

src/interface/blas2/backend/default.hpp
s-Nick: You are absolutely right @pgorlani , totally my fault. I will update them all to throw a `std::runtime_error` exceptions.

src/operations/blas1/IndexMaxMin.hpp
s-Nick: I am not sure how it could read out of bounds and in case how to properly check it.
The `val` variable is set for every thread to a neutral value for the operator at the beginning of the kernel. Then for `global_linear_id` smaller than the input array size is iteratively set using the `op::eval`. AFAIK the shift function is inside the subgroup so no random value should be read here. 
The only misleading point I see is that `val` is updated for all thread even if, at the end, only the thread with `sg_local_id` equal to zero has the correct final result. We don't care about it since this is the only thread that will write its result to memory.
If I am missing something on how to check it please let me know.

Rbiessy: You're right. I had flashbacks of out of bound issues with other group algorithms that were taking pointers so removing a similar `if` condition was not trivial.
As you said this looks fine here, thanks!

Rbiessy (approved):

src/interface/blas2/backend/default.hpp
s-Nick: Thank you @hjabird , pushed changes in 6a91a46eb187668a82aaf2f542fcabc6052c5eeb

src/interface/blas2/backend/default.hpp
s-Nick: Thank you both, addressed in a626b95be70f2506014dc335b4f1950d8db4b119

src/interface/blas2/backend/default.hpp
```diff
@@ -152,8 +165,21 @@ typename sb_handle_t::event_t _tbsv(
     sb_handle_t& sb_handle, index_t _N, index_t _K, container_t0 _mA,
     index_t _lda, container_t1 _vx, increment_t _incx,
     const typename sb_handle_t::event_t& _dependencies) {
-  return blas::internal::_tbsv_impl<4, 2, uplo, trn, diag>(
-      sb_handle, _N, _K, _mA, _lda, _vx, _incx, _dependencies);
+  const auto device = sb_handle.get_queue().get_device();
+  const std::string vendor =
+      device.template get_info<cl::sycl::info::device::vendor>();
```
hjabird, line 170: Can this go instide the if statement so it is only evaluated if it is needed?

src/interface/blas2/backend/default.hpp
```diff
@@ -163,12 +189,24 @@ namespace backend {
...
-                                    increment_t _incx,
-                                    const typename sb_handle_t::event_t& _dependencies) {
-  return blas::internal::_tpsv_impl<4, 2, uplo, trn, diag>(sb_handle, _N, _mA,
-                                                           _vx, _incx, _dependencies);
+typename sb_handle_t::event_t _tpsv(
+    sb_handle_t& sb_handle, index_t _N, container_t0 _mA, container_t1 _vx,
+    increment_t _incx, const typename sb_handle_t::event_t& _dependencies) {
+  const auto device = sb_handle.get_queue().get_device();
+  const std::string vendor =
+      device.template get_info<cl::sycl::info::device::vendor>();
```
hjabird, line 197: Same comment

hjabird: > Thank you @hjabird for clarifying it for me, I addressed both issues in [98f5314](https://github.com/codeplaysoftware/portBLAS/commit/98f5314788375e8afd2db42e73b64caa5da8577a)

hjabird (approved):

PR #518, opened 2024-05-23
Enable trsm test on NVIDIA GPUs when using DEFAULT TUNING_TARGET

s-Nick:
This PR fixes the error in `trsm` that led to disable its tests. Adding special icpx flag `-mllvm=loopopt=0` to the linker solves the issue. A minor change was necessary to compile `gemm` with mixed precision for default `TUNING_TARGET`.


Rbiessy (approved):

src/operations/blas3/gemm_no_local_full_vec.hpp
```diff
@@ -381,7 +381,7 @@ class Gemm<input_t, output_t, DoubleBuffer, NbcA, NbcB, ClSize, tile_type,
       auto C = orig_C;
 
       /* register array used to store the result*/
-      value_t reg_res[item_rows * item_cols];
+      element_t reg_res[item_rows * item_cols];
```
hjabird, line 384: Can you explain whats happening here?

s-Nick: Thank you @hjabird for the question, I'll do my best.
`value_t` is set to be the same type of the input, see `input_t`, removing the `const` qualifier. Then this type is used also for the output, it works as long as the input and the output are of the *same* type. With mixed `gemm`, it doesn't work anymore and it breaks at compile time.
I did not personally come up with this change/work around, but I am mimicking how it is implemented in other gemm implementations. The idea is that the scalar that multiplies the matrices, it is of the correct output type and, as far as I can see, it was selected and used also to store intermediate results for the output.

hjabird (approved):

PR #519, opened 2024-05-28
Apply enable-global-offset just to NVIDIA and AMD targets

pgorlani:
This patch restricts the application of the `enable-global-offset` flag just to NVIDIA and AMD targets since it could be not supported for other compilations.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -150,14 +150,17 @@ if (INSTALL_HEADER_ONLY)
   check_cxx_compiler_flag("-fsycl" is_dpcpp)
   if(is_dpcpp)
     target_compile_definitions(portblas INTERFACE "SB_ENABLE_USM")
-    if ((${CMAKE_CXX_COMPILER_ID} STREQUAL "IntelLLVM"
-           AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 2024.1)
-         OR NOT ${CMAKE_CXX_COMPILER_ID} STREQUAL "IntelLLVM")
-       # Apply only for oneAPI releases >= 2024.1 OR for intel/llvm.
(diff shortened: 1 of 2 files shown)


Rbiessy (approved):

CMakeLists.txt
```diff
@@ -150,14 +150,17 @@ if (INSTALL_HEADER_ONLY)
...
+    if (${support_disable_global_offset})
+      if ((${CMAKE_CXX_COMPILER_ID} STREQUAL "IntelLLVM"
+             AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 2024.1)
+           OR NOT ${CMAKE_CXX_COMPILER_ID} STREQUAL "IntelLLVM")
+         # Apply only for oneAPI releases >= 2024.1 OR for intel/llvm.
+         target_link_options(portblas INTERFACE "-mllvm=-enable-global-offset=false")
+         target_compile_options(portblas INTERFACE "-mllvm=-enable-global-offset=false")
+         message(STATUS "Adding -mllvm=-enable-global-offset=false to portblas")
+       endif()
+    endif()
```
hjabird, line 163: Why don't we need to check cuda / amd here?

pgorlani: Thanks for your review, @hjabird.

The first [commit](https://github.com/codeplaysoftware/portBLAS/pull/519/commits/61aa2498e17d7f6858fe45bdc73aa591b2750839) implemented a check based upon the `TUNING_TARGET`. Unfortunately, it is not possible to implement a check over the SYCL triplet in case of header only compilation as suggested.

After an internal discussion, we decided to check the required flag was supported in order to apply it regardless of the `TUNING_TARGET` for allowing it also in case of non-tuned configurations.

hjabird (approved):

PR #520, opened 2024-05-30
Removing all ComputeCpp occurences

s-Nick:
This PR remove all reference to ComputeCpp compiler and features related only to it, POWER_VR tuning target and IMGDNN library support.
In removing ComputeCpp support, this patch enables again the `exprtest` that was disabled for DPC++, now fully functional.


Rbiessy (commented):
There is a small occurrence of computecpp in `.gitignore`

s-Nick:
Thank you @Rbiessy , my grep config skipped it. Fixed in 59f37e6

Rbiessy (approved):

include/operations/blas_constants.h
```diff
@@ -263,18 +249,4 @@ struct constant_pair {
...
-
-template <typename ind_t, typename val_t>
-struct cl::sycl::is_device_copyable<const blas::IndexValueTuple<ind_t, val_t>>
-    : std::true_type {};
-
-template <typename ind_t, typename val_t>
-struct std::is_trivially_copyable<blas::IndexValueTuple<ind_t, val_t>>
-    : std::true_type {};
-#endif
-
```
hjabird, line 279: Why are we able to get rid of this?

src/operations/blas1_trees.hpp
```diff
@@ -445,10 +445,11 @@ PORTBLAS_INLINE bool TupleOp<rhs_t>::valid_thread(
     cl::sycl::nd_item<1> ndItem) const {
   return ((ndItem.get_global_id(0) < get_size()));
 }
+
 template <typename rhs_t>
 PORTBLAS_INLINE typename TupleOp<rhs_t>::value_t TupleOp<rhs_t>::eval(
     typename TupleOp<rhs_t>::index_t i) {
-  return TupleOp<rhs_t>::value_t(i, cl::sycl::fabs(rhs_.eval(i)));
```
hjabird, line 451: How does this relate to removing ComputeCpp?

include/operations/blas_constants.h
s-Nick: Thank you for your question @hjabird.

As you see I removed also both copy constructors from `IndexValueTuple` class above. In this way the compiler generates them for us, this make the class *trivially_copyable* (as I understood  from [this definition](https://en.cppreference.com/w/cpp/language/classes#Trivially_copyable_class)). This is key to remove the lines you highlighted. By SYCL2020 reference, if a class is *trivially_copyable* it is also [*device_copyable*](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec::device.copyable), which is needed for all buffers(see [this](https://github.com/intel/llvm/blob/sycl/sycl/include/sycl/buffer.hpp#L168)). These lines above force to consider `IndexValueTuple` class as *device_copyable* but removing the copy constructors needed for ComputeCpp it already has this property, so we don't need them anymore.

src/operations/blas1_trees.hpp
s-Nick: With the removal of ComputeCpp, I also checked that all exprtest work on DPC++. The `collapse_nested_tuple` test was disabled for it due to an llvm compiler error, but this is not true anymore. The test failed for DPC++ for this `sycl::fabs` which is not necessary anymore.
To give you more context, I think this function was here for the previous implementation of `iamax` and `iamin` that used to return the absolute value and the index. I thought the `abs` was still necessary for `iamax` and `iamin` but current operators implemented inside the kernel already compute the absolute value for each number, so there is no need to do it also here. So now all exprtest and unittest pass with DPC++.

hjabird (approved):

PR #524, opened 2024-06-28
Disable txsv intel gpu

s-Nick:
This PR disables `txsv` operators specifically for Intel ARC and Intel GPU Max GPUs by adding a runtime check on device type, vendor, and name.

The new exception is caught in our test suite, allowing us to flag the test as skipped. In cases where `TUNING_TARGET` is `INTEL_GPU` or `DPCPP_SYCL_ARCH` matches specific ARC or GPU Max targets, `txsv` tests are not added to ctest.

### DIFF
--- a/include/blas_meta.h
+++ b/include/blas_meta.h
@@ -228,7 +228,7 @@ class unsupported_exception : public std::runtime_error {
       : std::runtime_error(operator_name), _msg(operator_name) {
     _msg += " operator currently not supported on selected device";
   };
-  const char *what() const noexcept { return _msg.c_str(); }
+  const char *what() const noexcept override { return _msg.c_str(); }
 
  private:
   std::string _msg{};
--- a/include/blas_meta.h
+++ b/include/blas_meta.h
@@ -222,6 +222,16 @@ struct is_complex_std
 
 #endif
 
+class unimplemented_exception : public std::runtime_error {
+ public:
+  unimplemented_exception(const std::string &err)
+      : std::runtime_error(err), _msg(err) {};
+  const char *what() const throw() { return _msg.c_str(); }
+
+ private:
+  std::string _msg{};
+};
+
 }  // namespace blas
 
 #endif  // BLAS_META_H
--- a/include/blas_meta.h
+++ b/include/blas_meta.h
@@ -226,7 +226,7 @@ class unsupported_exception : public std::runtime_error {
  public:
   unsupported_exception(const std::string &operator_name)
       : std::runtime_error(operator_name), _msg(operator_name) {
-    _msg += " operator currently not supported on Arc or GPU Max GPUs";
+    _msg += " operator currently not supported on selected device";
   };
   const char *what() const throw() { return _msg.c_str(); }
 
--- a/include/blas_meta.h
+++ b/include/blas_meta.h
@@ -222,10 +222,12 @@ struct is_complex_std
 
 #endif
 
-class unimplemented_exception : public std::runtime_error {
+class unsupported_exception : public std::runtime_error {
  public:
-  unimplemented_exception(const std::string &err)
-      : std::runtime_error(err), _msg(err) {};
+  unsupported_exception(const std::string &operator_name)
+      : std::runtime_error(operator_name), _msg(operator_name) {
+    _msg += " operator currently not supported on Arc or GPU Max GPUs";
+  };
   const char *what() const throw() { return _msg.c_str(); }
 
  private:
--- a/include/blas_meta.h
+++ b/include/blas_meta.h
@@ -228,7 +228,7 @@ class unsupported_exception : public std::runtime_error {
       : std::runtime_error(operator_name), _msg(operator_name) {
     _msg += " operator currently not supported on selected device";
   };
-  const char *what() const throw() { return _msg.c_str(); }
+  const char *what() const noexcept { return _msg.c_str(); }
 
  private:
   std::string _msg{};
--- a/include/portblas_helper.h
+++ b/include/portblas_helper.h
@@ -239,9 +239,7 @@ inline void check_intel_gpu_support(const sb_handle_t &sb_handle,
       const std::string name =
           device.template get_info<sycl::info::device::name>();
       if (name.find("Arc") != name.npos || name.find("GPU Max") != name.npos) {
-        operator_name.append(
-            " operator currently not supported on Arc or GPU Max GPUs");
-        throw unimplemented_exception(operator_name);
+        throw unsupported_exception(operator_name);
       }
     }
   }
--- a/include/portblas_helper.h
+++ b/include/portblas_helper.h
@@ -241,7 +241,7 @@ inline void check_intel_gpu_support(const sb_handle_t &sb_handle,
       if (name.find("Arc") != name.npos || name.find("GPU Max") != name.npos) {
         operator_name.append(
             " operator currently not supported on Arc or GPU Max GPUs");
-        throw std::runtime_error(operator_name);
+        throw unimplemented_exception(operator_name);
       }
     }
   }
--- a/include/portblas_helper.h
+++ b/include/portblas_helper.h
@@ -228,9 +228,14 @@ inline bool is_malloc_shared(sb_handle_t &sb_handle, const containerT _rs) {
   }
 }
 
+/*
+ @brief Check device and throw unsupported exception if Intel discrete GPU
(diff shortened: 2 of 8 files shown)


include/blas_meta.h
```diff
@@ -222,6 +222,18 @@ struct is_complex_std
 
 #endif
 
+class unsupported_exception : public std::runtime_error {
+ public:
+  unsupported_exception(const std::string &operator_name)
+      : std::runtime_error(operator_name), _msg(operator_name) {
+    _msg += " operator currently not supported on Arc or GPU Max GPUs";
+  };
+  const char *what() const throw() { return _msg.c_str(); }
```
hjabird, line 231: Should this be marked as virtual?

include/blas_meta.h
```diff
@@ -222,6 +222,18 @@ struct is_complex_std
 
 #endif
 
+class unsupported_exception : public std::runtime_error {
+ public:
+  unsupported_exception(const std::string &operator_name)
+      : std::runtime_error(operator_name), _msg(operator_name) {
```
hjabird, line 228: I don't like that we're storing the operator name unused in the parent class, but all `runtime_error` constructors seem to require this. Is there anything we can do to avoid this?

include/portblas_helper.h
```diff
@@ -228,6 +228,23 @@ inline bool is_malloc_shared(sb_handle_t &sb_handle, const containerT _rs) {
   }
 }
 
+template <typename sb_handle_t>
+inline void check_intel_gpu_support(const sb_handle_t &sb_handle,
```
hjabird, line 232: The name of this function doesn't really match what it does. Perhaps `throw_unsupported_on_intelgpu`?

hjabird, line 232: Needs documentation.

hjabird: Or be mores specific actually because iGPU is supported.

test/unittest/CMakeLists.txt
```diff
@@ -103,6 +103,16 @@ if(is_dpcpp)
   )
 endif()
 
+if( (DEFINED TUNING_TARGET AND (${TUNING_TARGET} STREQUAL "INTEL_GPU")) OR
```
Rbiessy, line 106: This would fail to be parsed if `TUNING_TARGET` is not defined despite the first condition. The quotes are the best practice to avoid this issue, that way you still get an empty string if `TUNING_TARGET` is not defined. (similarly below)
```suggestion
if( ("${TUNING_TARGET}" STREQUAL "INTEL_GPU") OR
```

include/blas_meta.h
```diff
@@ -222,6 +222,18 @@ struct is_complex_std
...
+    _msg += " operator currently not supported on Arc or GPU Max GPUs";
```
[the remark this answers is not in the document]
s-Nick: Thank you for your review @hjabird 
I agree, in retrospective this is not reusable. I changed it in `[...] not supported on selected device` in [936fe3b](https://github.com/codeplaysoftware/portBLAS/pull/524/commits/936fe3bef681952b8daeb5a7b21b52852403707c)

include/blas_meta.h
```diff
@@ -222,6 +222,18 @@ struct is_complex_std
 
 #endif
 
+class unsupported_exception : public std::runtime_error {
```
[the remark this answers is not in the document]
s-Nick: I started with `unimplemented` as you can see in [18a7bf6](https://github.com/codeplaysoftware/portBLAS/pull/524/commits/18a7bf67de8916e2999c5e0686806dea159e5536)  , but then changed it since technically it is implemented in portBLAS but not supported for the cases I am adding it. 
Yes, `unimplemented` is closer to oneMKL spec, if you strongly feel I should change it for better consistency to oneMKL put a thumbs up here and I'll update id. (@Rbiessy )

include/blas_meta.h
s-Nick: I checked the `std::runtime_error` class from [here](https://en.cppreference.com/w/cpp/error/runtime_error) and all constructors require a string or another exception as argument, so I don't think we can avoid it.

include/portblas_helper.h
s-Nick: Updated to `throw_unsupported_intel_dGPU` in [1d7efb9](https://github.com/codeplaysoftware/portBLAS/pull/524/commits/1d7efb99089ac2a85f20fc43790698ac77651f4a)

s-Nick: Done in [1d7efb9](https://github.com/codeplaysoftware/portBLAS/pull/524/commits/1d7efb99089ac2a85f20fc43790698ac77651f4a)

include/blas_meta.h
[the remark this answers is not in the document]
hjabird: I don't feel that strongly! If Rbiessy doesn't add anything, I'm happy for you to resolve this comment with no changes.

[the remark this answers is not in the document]
Rbiessy: That's fair, I don't feel strongly either so happy to see this comment resolved.

include/blas_meta.h
s-Nick: I don't think we should/will ever derive from this exception class, so if it is `virtual` or not doesn't matter.

test/unittest/CMakeLists.txt
s-Nick: Thank you for your comment @Rbiessy 
I was not aware of it. Updated in  [8116dbf](https://github.com/codeplaysoftware/portBLAS/pull/524/commits/8116dbf4453f9eaa3a51eaf1035355a06a165bfd)

include/blas_meta.h
hjabird: I think it should be indicated that its virtual in anyway using `override` - [C++ core guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rh-override). You'll also get warnings in oneMKL without this - see my previous [PR](https://github.com/oneapi-src/oneMKL/pull/454).

test/unittest/CMakeLists.txt
```diff
@@ -103,6 +103,16 @@ if(is_dpcpp)
...
+if( ("${TUNING_TARGET}" STREQUAL "INTEL_GPU") OR
+    ("${DPCPP_SYCL_ARCH}" MATCHES "_pvc|_agm") )
+    message(WARNING "disabling trsv, tbsv and tpsv tests for Intel ARC and Intel GPU Max GPUs")
+    set(TESTS_TO_SKIP
+      ${PORTBLAS_UNITTEST}/blas2/blas2_trsv_test.cpp
+      ${PORTBLAS_UNITTEST}/blas2/blas2_tbsv_test.cpp
+      ${PORTBLAS_UNITTEST}/blas2/blas2_tpsv_test.cpp
+    )
+endif()
+
```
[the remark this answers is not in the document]
s-Nick: Thank you for the review @pgorlani 

I thought to use both mechanism to avoid issues with tests. I disabled entirely the test for `INTEL_GPU` and specific target to entirely avoid output about them, since we know that right now they are not supported for these specific devices. I catch the exception inside the tests in case portBLAS is built with `DEFAULT` `TUNING_TARGET` to avoid test failure on all possible device. Do you find it redundant?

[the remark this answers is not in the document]
pgorlani: So far, these operators are supported for the integrated Intel GPUs, so they should be tested in case of `INTEL_GPU` target. In case of Arc or Max GPUs, the runtime check skips them as no matter is the tuning target.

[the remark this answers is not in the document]
s-Nick: yes, we can use the run-time checks only. I will remove the cmake checks entirely, even though I would have kept the  check on the arch.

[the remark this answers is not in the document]
s-Nick: Addressed in [75960e5](https://github.com/codeplaysoftware/portBLAS/pull/524/commits/75960e5bd3b523c9850d21feb017f6b3523b5595)

include/blas_meta.h
s-Nick: Thank you for the reference links, fixed in [75960e5](https://github.com/codeplaysoftware/portBLAS/pull/524/commits/75960e5bd3b523c9850d21feb017f6b3523b5595)

hjabird (approved):

Rbiessy (approved):

pgorlani (approved):

PR #525, opened 2024-07-03
Remove github workflow

s-Nick:
Current workflow crash when building the docker image due to outdated configuration. This PR removes the workflow entirely


hjabird (approved):

Rbiessy:
We won't need dependabot anymore so we can remove the whole `.github` folder

s-Nick:
> We won't need dependabot anymore so we can remove the whole `.github` folder

I removed everything now

Rbiessy (approved):

PR #533, opened 2024-07-31
Refactor SPR/SPR2 index group_broadcast

pgorlani:
This patch substitutes the `group_broadcast` of two individual indexes with a single broadcast of a vector type.

### DIFF
--- a/src/operations/blas2/spr.hpp
+++ b/src/operations/blas2/spr.hpp
@@ -101,12 +101,12 @@ typename rhs_1_t::value_t Spr<Single, isUpper, lhs_t, rhs_1_t, rhs_2_t>::eval(
 
   index_t row = 0, col = 0;
 
-#if defined(INTEL_GPU) || defined(NVIDIA_GPU)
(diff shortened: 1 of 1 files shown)


src/operations/blas2/spr.hpp
```diff
@@ -101,20 +101,21 @@ typename rhs_1_t::value_t Spr<Single, isUpper, lhs_t, rhs_1_t, rhs_2_t>::eval(
 
   index_t row = 0, col = 0;
 
-#ifndef __ADAPTIVECPP__
```
hjabird, line 104: Why is it safe to remove the `__ADAPTIVECPP__` macros?

hjabird (approved):

s-Nick (approved):

src/operations/blas2/spr.hpp
pgorlani: Thanks for your reviews.

I rechecked and contrary to what I thought this could be still necessary. I reintroduced it in 381afab.

